PR #70, opened 2021-08-22
fix: enable generation of sourcemap files

web-padawan:
## Description

Enabled the config options to generate sourcemaps for better DX and to avoid console warnings in `web-test-runner`.

Fixes vaadin/flow#11424

## Type of change

- Bugfix

### DIFF
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -4,11 +4,14 @@
     "experimentalDecorators": true,
     "importHelpers": true,
     "declaration": true,
+    "declarationMap": true,
     "target": "es2019",
     "module": "esnext",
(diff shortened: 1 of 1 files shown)


haijian-vaadin (commented):
It looks good, I think this should also be cherry-picked to previous versions? Could you help to make a PR to https://github.com/vaadin/flow/blob/master/flow-client/tsconfig.json as well?
This new client package is not in use yet. Current it still uses the one in the flow-client folder in Flow repo.

web-padawan:
Sure, I can submit a PR to Flow repo as well 👍

Lodin (approved):

PR #72, opened 2021-08-22
fix: add .js extension to all imports

### LINKED ISSUE #71: Consider adding .js extension to external imports
## Problem

Currently the compiled JS files do not contain the `.js` extension, which makes them incompatible with using in browsers by importing directly, without any tools that perform module resolution, such as `webpack` of `web-dev-server`.

Lit follows this pattern, see for example https://github.com/lit/lit/issues/1200#issuecomment-672323641

## Expected outcome

TS files should contain `.js` extensions, to make sure they are present in the output produced by `tsc`.

## Proposed solution

Add `.js` extension to all internal imports.

web-padawan:
## Description

1. Updated all the import statements to use `.js` extension, including those for `valiidator` package.
2. Disabled `import/no-unresolved` ESLint rule as it does not play well with ES modules (see comment).

Fixes #71

## Type of change

- Internal change

### DIFF
--- a/client/.eslintrc
+++ b/client/.eslintrc
@@ -4,6 +4,8 @@
   "rules": {
     "@typescript-eslint/ban-ts-comment": "off",
     "@typescript-eslint/no-non-null-assertion": "off",
+    // https://github.com/import-js/eslint-plugin-import/issues/2170#issuecomment-889657579
+    "import/no-unresolved": "off",
     "max-classes-per-file": "off",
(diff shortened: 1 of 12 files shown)


Lodin (approved):

web-padawan:
There was a merge conflict, @Lodin please take another look if everything looks good after I resolved it.

Lodin (approved):

PR #77, opened 2021-08-25
Extract form to a separate package & use new export features

Lodin:
This PR extracts the form code to a separate `@vaadin/form` package.

It also applies the new export features (similar for the approach `lit` uses).


Lodin:
@web-padawan, could you please also take a look at the PR? I would like to hear your opinion on changes I did for package `exports` field. I tried to implement the same `package.json` approach as `lit` team did.

PR #90, opened 2021-09-06
chore(frontend): more fixes in order to make the release runnable

Lodin:
- Rename GITHUB_TOKEN -> RELEASE_TOKEN in case GitHub overwrites the variable by a different value
- Make bash script falling on any command error

### DIFF
--- a/.github/workflows/frontend-release.yml
+++ b/.github/workflows/frontend-release.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Bump New Version
         run: sh scripts/bump.sh
         env:
-          RELEASE_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+          GITHUB_RELEASE_TOKEN: ${{ secrets.GITHUB_RELEASE_TOKEN }}
           REPO: ${{ github.repository }}
           VERSION_TAG: ${{ github.event.release.tag_name }}
       - name: Publish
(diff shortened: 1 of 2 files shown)


frontend/scripts/bump.sh
```diff
@@ -1,11 +1,13 @@
 #!/usr/bin/env bash
 
+set -eu
```
haijian-vaadin, line 3: for maintenance reason, could have a comment on this to explain the purpose?

.github/workflows/frontend-release.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - name: Bump New Version
         run: sh scripts/bump.sh
         env:
-          GITHUB_TOKEN: ${{ secrets.RELEASE_TOKEN }}
+          RELEASE_TOKEN: ${{ secrets.RELEASE_TOKEN }}
```
haijian-vaadin, line 39: I think it's still better to have GITHUB_THERE in the name. e.g. `RELEASE_GITHUB_TOKEN` or sth similar, so that it's easier to understand it's related to GitHub, otherwise, people have to dive into the code to find out what is the `RELEASE_TOKEN`.

haijian-vaadin (changes requested):

.github/workflows/frontend-release.yml
Lodin: What about `GIT_RELEASE_TOKEN`? I wanted to also change secret to have `GITHUB_RELEASE_TOKEN` but GitHub complains that secrets cannot start with `GITHUB`

haijian-vaadin (approved):

PR #100, opened 2021-09-08
chore(frontend): fix dependencies & simplify release workflow

Lodin:


frontend/packages/fusion-frontend/package.json
```diff
@@ -45,7 +45,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@vaadin/common-frontend": "^22.0.0-alpha1",
+    "@vaadin/common-frontend": "^0.0.10",
```
haijian-vaadin, line 48: should this be updated also automatically in the version bump script?

haijian-vaadin (changes requested):

frontend/packages/fusion-frontend/package.json
Lodin: Yeah, that's why I replaced `scripts/bump/update-package-dependencies.js` with `npx lerna version` in `scripts/bump/exec.sh`

Lodin: Though for now this change is necessary because bot already changed our `package.json` files, and without this change the bootstrap will break

haijian-vaadin (approved):

PR #112, opened 2021-09-20
fix: NumberModel should read empty string as undefined

haijian-vaadin:
Related to https://github.com/vaadin/spring/issues/904.
Porting the fix in https://github.com/vaadin/flow/pull/11169.


Lodin (approved):
LGTM

PR #113, opened 2021-09-21
chore: upgrade lit version to 2.0.0

haijian-vaadin:


frontend/packages/form/package.json
```diff
@@ -57,7 +57,7 @@
     "access": "public"
   },
   "dependencies": {
-    "lit": "^2.0.0-rc.2",
+    "lit": "^2.0.0",
```
Lodin, line 60: Here `lit` should be moved to `peerDependencies`:
```
"dependencies": {
  "validator": "^13.6.0"
},
"peerDependencies": {
  "lit": "^2.0.0"
}
```

frontend/packages/fusion-frontend/package.json
```diff
@@ -46,7 +46,7 @@
   },
   "dependencies": {
     "@vaadin/common-frontend": "^0.0.12",
-    "lit": "^2.0.0-rc.2"
+    "lit": "^2.0.0"
```
Lodin, line 49: Here `lit` should also be moved to `peerDependencies`:
```
"dependencies": {
  "@vaadin/common-frontend": "^0.0.12"
},
"peerDependencies": {
  "lit": "^2.0.0"
}
```

Lodin (changes requested):

frontend/packages/fusion-frontend/package.json
Haprog: If we release a new `@vaadin/common-frontend` first we might as well update the version of that here (though it's not necessarily needed since we use the caret here).

frontend/packages/form/package.json
haijian-vaadin: thanks, done

frontend/packages/fusion-frontend/package.json
haijian-vaadin: done

Lodin (approved):

PR #118, opened 2021-09-30
fix(frontend): add missing tslib dependency

Lodin:
Follow-up for vaadin/flow-fusion-common#3


haijian-vaadin (changes requested):
Do we need to repeat the `tslib` dependency? Isn't it enough to add the dependency in flow-fusion-common?

platosha:
Closing, turns out `tslib` is not missing here, the packages don’t use it.

PR #128, opened 2021-11-15
chore: minor improvements, mainly linting fixes

Haprog:
### Summary

Enable linting and type checking for test sources. Get rid of all current lint issues (also warnings). Fix command in readme and few minor refactorings.

### Details:

chore(readme): fix command in readme

- The main `package.json` is under the `frontend` directory and not in repository root. Fix the related `cd` command in the readme.

chore: improve IDE linting config for tests

- Add tsconfigs to the test directories to prevent linting errors in VS Code related to global mocha methods like describe being undefined. The test sources need to be included in some tsconfig for the TS integration in the IDE to use e.g. the mocha type declarations in the context of the test files.

chore: fix syntax error in `tsconfig.json`

chore: enable linting and type checking for test sources

test: improve cookie handling utils

- The old version of `clearCookie()` that just sets an empty value to a cookie doesn't actually remove the cookie. It just makes it so the cookie has an empty value but it still exists. To actually remove a cookie it needs to be set with an `Expires` option in the past.

test: remove unused const

test: remove non-functional cookie restoration code

- This is not a working way to save and restore cookies using `document.cookie`. This is also not needed since the tests that are setting cookies are also taking care of clearing them using `try..finally` and `clearCookie()`.

chore: turn off lint rule `lit/quoted-expressions`

- This rule needs to be off for element expressions. Would be nice if there was a rule to enable this only for the other type of expressions like attribute, property and event handler bindings.

chore: turn off lint rule `@typescript-eslint/no-explicit-any`

- We are currently using explicit any in most source files so let's ignore this for now so we can better focus on not introducing new lint warnings/errors.

chore: turn off lint rule `@typescript-eslint/explicit-module-boundary-types`

test: fix linting issues in tests

### DIFF
--- a/frontend/packages/fusion-frontend/tsconfig.json
+++ b/frontend/packages/fusion-frontend/tsconfig.json
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "rootDir": "src",
-    "outDir": "."
+    "outDir": ".",
+    "skipLibCheck": true
   },
   "include": ["src"],
   "exclude": ["!src/**/*"]
--- a/frontend/.eslintrc
+++ b/frontend/.eslintrc
@@ -25,6 +25,7 @@
     ],
 
     "no-unused-expressions": "off",
-    "chai-friendly/no-unused-expressions": "error"
+    "chai-friendly/no-unused-expressions": "error",
+    "lit/quoted-expressions": "off" // Needs to be off for element expressions, but would be nice to have on for other kind of expressions
   }
 }
--- a/frontend/.eslintrc
+++ b/frontend/.eslintrc
@@ -4,6 +4,7 @@
   "rules": {
     "@typescript-eslint/ban-ts-comment": "off",
     "@typescript-eslint/no-non-null-assertion": "off",
+    "@typescript-eslint/no-explicit-any": "off",
     // https://github.com/import-js/eslint-plugin-import/issues/2170#issuecomment-889657579
     "import/no-unresolved": "off",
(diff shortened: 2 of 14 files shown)


frontend/packages/fusion-frontend/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "rootDir": "src",
-    "outDir": "."
+    "outDir": ".",
+    "skipLibCheck": true
```
Lodin, line 6: Is it necessary? Does the absence of this rule break something?

Lodin (approved):
LGTM

frontend/packages/fusion-frontend/tsconfig.json
Haprog: Without this in `fusion-frontend` package the type check (`npm run typecheck`) fails with:
```
node_modules/fetch-mock/esm/client.d.ts:694:16 - error TS2649: Cannot augment module 'fetchMock' with value exports because it resolves to a non-module entity.

694 declare module 'fetch-mock/esm/client' {
                   ~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```
And this looked like something that can't be fixed in our code so I added this to make it pass.

I added the same option to the form package for consistency even though it didn't have similar errors for now.

Lodin: Ah, I see, thanks. Looks like I was satisfied with the Intellij check and haven't tested the `typecheck` command properly.

PR #129, opened 2021-11-16
feat: clear JWT cookie on logout

### LINKED ISSUE #49: Client-side Spring Security logout() helper should clear out the cookies
For the split-cookie approach, the client-side logout() helper should delete the `jwt.headerAndPayload` cookie, so that a user could log out properly even offline.

Haprog:
## Description

Delete the `jwt.headerAndPayload` cookie when using the client-side `logout()` helper. This enables a proper logout even when the user is offline.

Closes #49

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


packages/ts/fusion-frontend/src/Authentication.ts
```diff
@@ -47,6 +50,23 @@ async function doLogout(logoutUrl: string, headers: Record<string, string>) {
   await updateCsrfTokensBasedOnResponse(response);
 }
 
+/**
+ * Remove trailing '/' if the path is not exactly '/'.
+ * Exported for testing purposes only.
+ * @internal
+ */
+export function _removeTrailingSlashFromPath(path: string) {
```
platosha, line 58: Would be nicer to move into a separate `.ts` module and have a separate test suite, that way we could avoid making it a public API.

packages/ts/fusion-frontend/src/Authentication.ts
```diff
@@ -146,7 +166,12 @@ export async function logout(options?: LogoutOptions) {
     } catch (error) {
       // clear the token if the call fails
       clearSpringCsrfMetaTags();
-      throw error;
+
+      if (cookieExists(jwtCookieName)) {
+        deleteJWTCookie();
```
platosha, line 171: So it only attempts to delete the cookie when some error happens, and then it discards the error if the cookie is detected.

I don’t think it’s a good idea to discard the error without even asserting its type. This would make troubleshooting harder if there is some problem (for ex., a server-side logout hander misconfiguration), but no error. Maybe it is fine to handle if we determine the error caused by a connection failure, but `doLogout` also throws if there was a non-OK response.

I’d rather suggest that we check and delete the cookie in the `finally` block, so that it is:
- removed no matter if there was an error or not
- does not discard errors

packages/ts/fusion-frontend/src/Authentication.ts
platosha: Oh is this only used for cookies, so could be merged with `CookieUtils.ts` perhaps?

platosha (changes requested):

packages/ts/fusion-frontend/src/Authentication.ts
Haprog: Yeah good idea, then it won't be in any of the exported entry points. I didn't think of that when adding this.

The logic here is not really specific to cookies in any way but it's currently only used related cookies so I guess it's ok to have it in the CookieUtils for now to avoid creating a new util file just for this. I'll move it.

Haprog: Updated.

packages/ts/fusion-frontend/src/Authentication.ts
Haprog: Updated.

platosha (approved):

PR #110, opened 2021-09-16
feat: add new Fusion generator

Lodin:
Address vaadin/flow#9010

This PR introduces a new Fusion generator that works as a Maven and Gradle plugin.

PR #139, opened 2021-12-01
refactor: add barrel plugin, client plugin, and code coverage + various improvements

### LINKED ISSUE #134: A generator plugin for default connect client generation
The new generator doesn't generate the `connect-client.default.ts` file yet, which is used for the endpoint calls.

This could be achieved with a plugin for the new generator.

Lodin:
Follow-on PR for #110.
Resolves #134.


packages/ts/generator-typescript-plugin-client/package.json
```diff
@@ -0,0 +1,73 @@
...
+      "default": "./index.js"
+    },
+    "./index.js": {
+      "default": "./index.js"
+    }
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/vaadin/fusion.git",
+    "directory": "packages/ts/generator-typescript-utils"
```
Haprog, line 32:
```suggestion
    "directory": "packages/ts/generator-typescript-plugin-client"
```

packages/ts/generator-typescript-cli/test/GeneratorIO.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { expect } from 'chai';
+
+// TODO: This is a stub. We need to add real tests
```
Haprog, line 3: Do we expect to add these missing tests in the scope of this PR or later? If later, there should be another issue about these.

Haprog (changes requested):
Looking good mostly

packages/ts/generator-typescript-cli/test/GeneratorIO.spec.ts
Lodin: They should go later. The stub is added only to satisfy the test runner (otherwise, it fails).

packages/ts/generator-typescript-plugin-client/package.json
Lodin: Oops, thanks

platosha (dismissed):

packages/java/parser-jvm-core/src/main/java/com/vaadin/fusion/parser/core/Parser.java
```diff
@@ -68,48 +71,56 @@ public SharedStorage getStorage() {
...
+        }
+
+        public Collection<RelativeClassInfo> collectEntities(
+                Collection<RelativeClassInfo> endpoints) {
+            var entities = endpoints.stream().flatMap(
                     cls -> cls.getInheritanceChain().getDependenciesStream())
-                    .collect(Collectors.toCollection(LinkedHashSet::new));
+                    .collect(Collectors.toCollection(ArrayList::new));
 
             // ATTENTION: This loop mutates the collection during processing!
```
haijian-vaadin, line 107: This doesn't look good. I think what you can do is to create a copy of the entities, iterate through the copy, and modify the entities. e.g.,

```
var originalEntities = new ArrayList<>(entities);
for (var entity : originalEntities) {
    entity.getDependenciesStream().filter(e -> !entities.contains(e)).forEach(entities::add);
}
```

packages/ts/generator-typescript-plugin-backbone/test/complex-hierarchy/fixtures/ComplexHierarchyEndpoint.snap.ts
```diff
@@ -1,4 +1,4 @@
-import type ComplexHierarchyModel_1 from "./com/vaadin/fusion/parser/plugins/backbone/complexhierarchy/models/ComplexHierarchyModel.js";
-import client_1 from "./connect-client.default.js";
+import type ComplexHierarchyModel_1 from "./com/vaadin/fusion/parser/plugins/backbone/complexhierarchy/models/ComplexHierarchyModel";
```
haijian-vaadin, line 1: not very important, but why do we have these `_1` suffixes?

haijian-vaadin (changes requested):

packages/java/parser-jvm-core/src/main/java/com/vaadin/fusion/parser/core/Parser.java
Lodin: That won't work, unfortunately. The `getDependenciesStream` gives me only the one layer of dependencies while I need to collect all of them. Your example would allow collecting only that one layer because for entities added to the `entities` collection the `getDependenciesStream` is not called. It is because the iteration doesn't affect that collection. That's the reason I used that non-trivial approach.

packages/ts/generator-typescript-plugin-backbone/test/complex-hierarchy/fixtures/ComplexHierarchyEndpoint.snap.ts
Lodin: It is the result of TypeScript's `createUniqueName` function. It creates an identifier that will be unique across the source file. I use it everywhere to avoid naming collision.

packages/java/parser-jvm-core/src/main/java/com/vaadin/fusion/parser/core/Parser.java
Lodin: I think it is better to leave as is. All other solutions I came to are more complex and clunky. I would probably need to make comment more detailed to be more understandable.

haijian-vaadin: ok

haijian-vaadin (approved):

platosha (dismissed):

packages/ts/generator-typescript-plugin-client/test/basic/fixtures/client.snap.ts
```diff
@@ -0,0 +1,3 @@
+import { ConnectClient as ConnectClient_1 } from "@vaadin/flow-frontend/Connect";
```
haijian-vaadin, line 1: should be `@vaadin/fusion-frontend`

packages/ts/generator-typescript-plugin-client/src/ClientProcessor.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.#filePath = new PathManager('ts').createRelativePath(fileName);
+    this.#logger = logger;
+  }
+
+  public process(): SourceFile {
+    this.#logger.debug(`Generating ${this.#filePath}`);
+
+    const { exports, imports, paths } = new DependencyManager(new PathManager());
+    const clientImportId = imports.named.add(
+      paths.createBareModulePath('@vaadin/flow-frontend/Connect', true),
```
haijian-vaadin, line 22: same here, should be `@vaadin/fusion-frontend`

haijian-vaadin (changes requested):

platosha (dismissed):

packages/ts/generator-typescript-plugin-client/test/basic/fixtures/client.snap.ts
Haprog: What's the conclusion here? This was marked as resolved but it's still saying `@vaadin/flow-frontend` there. Also same for the next resolved item below.

haijian-vaadin: it should be fusion-frontend, flow-frontend was a typo. There was another issue I found and Vlad is fixing them both at the same time.

Lodin: oops, fixed

haijian-vaadin (dismissed):

Haprog (approved):

.eslintrc
```diff
@@ -9,6 +10,7 @@
     // https://github.com/import-js/eslint-plugin-import/issues/2170#issuecomment-889657579
     "import/no-unresolved": "off",
     "max-classes-per-file": "off",
+    "no-bitwise": "off",
```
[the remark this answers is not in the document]
Lodin: Ok, I agree with these arguments. Fixed

haijian-vaadin (approved):

PR #142, opened 2021-12-03
fix: revert TS target to ES2019 to support Webpack 4

Haprog:
## Description

With TS target of ES2020 the `fusion-frontend` and `form` packages aren't usable with Webpack 4 due to optional chaining not being supported with Webpack 4. Trying to use it results in unexpected token error during Webpack build.

For more info about the problem when using ES2020 see https://github.com/webpack/webpack/issues/10227

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


Lodin:
Could we override the target for only `fusion-frontend` and `form` instead of setting it for all the packages?

Lodin (approved):

PR #153, opened 2021-12-13
feat(java/maven-plugin): add ability to set plugin configuration

Lodin:
With this change, plugins will be able to have a configuration using the following approach:
```xml
<plugin>
    <groupId>com.vaadin.fusion</groupId>
    <artifactId>fusion-maven-plugin</artifactId>
    <version>1.0-SNAPSHOT</version>
    <configuration>
        <generator>
            <outputDir>frontend/generated-new</outputDir>
        </generator>
        <parser>
            <plugins>
                <use>
                    <plugin>
                        <name>com.vaadin.fusion.parser.plugins.nonnull.NonnullPlugin</name>
                        <configuration implementation="com.vaadin.fusion.parser.plugins.nonnull.NonnullPluginConfig">
                            <annotations>
                                <annotation>javax.annotation.Nonnull</annotation>
                                <annotation>org.jetbrains.annotations.NotNull</annotation>
                            </annotations>
                        </configuration>
                    </plugin>
                </use>
            </plugins>
        </parser>
    </configuration>
</plugin>
```


packages/java/maven-plugin/src/main/java/com/vaadin/fusion/maven/ParserProcessor.java
```diff
@@ -85,6 +86,8 @@ public ParserProcessor plugins(
         this.plugins = pluginStream
                 .collect(Collectors.toCollection(LinkedHashSet::new));
 
+        this.plugins.forEach(p -> System.out.println(p.getConfiguration()));
```
haijian-vaadin, line 89: for debugging? could be removed?

packages/java/maven-plugin/src/main/java/com/vaadin/fusion/maven/ParserConfiguration.java
```diff
@@ -46,6 +50,10 @@ public boolean equals(Object other) {
             return Objects.equals(name, ((Plugin) other).name);
         }
 
+        public Optional<PluginConfiguration> getConfiguration() {
```
haijian-vaadin, line 52: why use Optional here?

haijian-vaadin (changes requested):

packages/java/maven-plugin/src/main/java/com/vaadin/fusion/maven/ParserProcessor.java
Lodin: Oops

packages/java/maven-plugin/src/main/java/com/vaadin/fusion/maven/ParserConfiguration.java
Lodin: Just to note that it could be `null`. Should I remove it?

haijian-vaadin: In the code, I see there is `orElse(null)`, so probably doesn't need this optional wrapper in the first place.

Lodin: Ok, replaced with `@Nonnull` annotation for all non-nullable values

packages/java/parser-jvm-utils/src/main/java/com/vaadin/fusion/parser/utils/ConfigList.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public interface ConfigList<Item> {
```
haijian-vaadin, line 11: I think this class needs some java doc or maybe better method names. 
why getDisable() returns a Collection of Item? 
the same for getUse()

haijian-vaadin (changes requested):

packages/java/maven-plugin/src/main/java/com/vaadin/fusion/maven/ParserConfiguration.java
```diff
@@ -22,12 +30,17 @@
...
-    public Optional<PluginList> getPlugins() {
+    public Optional<Plugins> getPlugins() {
         return Optional.ofNullable(plugins);
     }
 
     public static class Plugin {
-        private final String name;
+        private PluginConfiguration configuration;
+        private String name;
+        private Integer order;
```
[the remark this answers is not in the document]
Lodin: No, because it can be `null`.

packages/java/parser-jvm-core/src/main/java/com/vaadin/fusion/parser/core/ParserConfig.java
```diff
@@ -158,10 +159,11 @@ public Builder openAPISource(@Nonnull String src,
         }
 
         @Nonnull
-        public Builder plugins(@Nonnull Collection<String> plugins) {
+        public Builder plugins(@Nonnull Collection<Plugin> plugins) {
             Objects.requireNonNull(plugins);
-            actions.add(
-                    config -> config.plugins = new LinkedHashSet<>(plugins));
+            actions.add(config -> config.plugins = plugins instanceof TreeSet
```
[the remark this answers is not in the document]
Lodin: Nice catch, thanks!

packages/java/maven-plugin/src/main/java/com/vaadin/fusion/maven/ParserConfiguration.java
```diff
@@ -1,14 +1,22 @@
...
 import java.util.Optional;
+import java.util.Set;
+
+import javax.annotation.Nonnull;
+
+import com.vaadin.fusion.parser.core.PluginConfiguration;
+import com.vaadin.fusion.parser.plugins.backbone.BackbonePlugin;
+import com.vaadin.fusion.parser.utils.ConfigList;
 
 public final class ParserConfiguration {
```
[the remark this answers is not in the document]
Lodin: Sure

packages/java/maven-plugin/src/main/java/com/vaadin/fusion/maven/ParserConfiguration.java
```diff
@@ -22,12 +30,17 @@
         return Optional.ofNullable(openAPIPath);
     }
 
-    public Optional<PluginList> getPlugins() {
+    public Optional<Plugins> getPlugins() {
         return Optional.ofNullable(plugins);
     }
 
     public static class Plugin {
```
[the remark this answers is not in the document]
Lodin: I'm trying to maintain an approach where I keep only the top-level structure package-private since all nested classes/methods will become package-private even if they are declared as `public`.

packages/java/parser-jvm-utils/src/main/java/com/vaadin/fusion/parser/utils/ConfigList.java
Lodin: Hmm, well, `getDisable` is a getter that returns `disable` property which is a maven-level configuration part:
```xml
<plugin>
  <configuration>
    <plugins>
      <disable>
        <plugin>com.vaadin.fusion.parser.plugins.backbone.BackbonePlugin</plugin>
      </disable>
      <use>
        <plugin>com.vaadin.fusion.parser.plugins.nonnull.NonnullPlugin</plugin>
      </use>
  </configuration>
</plugin
```
I thought I cannot change the property name since it should be read by maven. But probably I could use something like `@JsonProperty` of Jackson? Or I should just use another name for the getter?

haijian-vaadin: No need to use `@JsonProperty`, could you help to add this explanation as JavaDoc to what the `disable` and `use` do? It would help to understand how things work in the future.

Lodin: Done!

haijian-vaadin (approved):

Issue #135, opened 2021-11-30
A plugin for Fusion form model generation


haijian-vaadin:
For the Fusion form binding feature to work, it needs the auto-generated model, which is not available yet in the new generator. 

This could be achieved with a plugin for the new generator.

PR #171, opened 2021-12-28
feat(generator): emit model validators from Java validation constraint annotations

platosha:
Related to #135 

Based on https://github.com/vaadin/fusion/pull/152


packages/java/maven-plugin/src/main/java/com/vaadin/fusion/maven/GeneratorProcessor.java
```diff
@@ -18,7 +18,7 @@
     private final MavenProject project;
     private String input;
     private String outputDir = "frontend/generated";
-    private Set<GeneratorConfiguration.Plugin> plugins = new LinkedHashSet<>(3);
+    private Set<GeneratorConfiguration.Plugin> plugins = new LinkedHashSet<>(4);
```
haijian-vaadin, line 21: Is it the number of default plugins? maybe better to extract this magic number to a constant?

haijian-vaadin (changes requested):

haijian-vaadin (approved):

PR #179, opened 2022-01-10
refactor(java): move ValidationConstraint to the new Model plugin

### LINKED ISSUE #180: move openapi from backbone plugin to core
This ticket is created as a result of the code review from PR https://github.com/vaadin/fusion/pull/156#discussion_r782069325
The AssociationMap which provides the OenAPI information should be in the core as well.

Lodin:
Moves `ValidationConstraint` class created in #171 to the new Model plugin to preserve Backbone plugin as clean as possible.

It uses `AssociationMap` introduced in #156, so this PR can be merged only after #156 is merged.

This PR also addresses changed done in #175 by refactoring the `parser-jvm-utils` package.

Resolves #180


packages/java/parser-jvm-plugin-model/src/test/java/com/vaadin/fusion/parser/model/utils/TestBase.java
```diff
@@ -0,0 +1,46 @@
...
+
+import io.swagger.v3.core.util.Json;
+import io.swagger.v3.oas.models.OpenAPI;
+
+public abstract class TestBase {
+    protected final ObjectMapper mapper = Json.mapper();
+    protected final ResourceLoader resourceLoader;
+    protected Path targetDir;
+
+    {
```
haijian-vaadin, line 23: could we move this into a @BeforeClass method? I think it's more clear than this instance initializer block.

haijian-vaadin (changes requested):

packages/java/parser-jvm-plugin-model/src/test/java/com/vaadin/fusion/parser/model/utils/TestBase.java
Lodin: Nope, unfortunately. I don't understand why but JUnit fails to find a method with `@BeforeEach` in the parent class. If you know why it happens and how to fix that, it would be great to implement

packages/java/parser-jvm-core/src/main/java/com/vaadin/fusion/parser/core/PluginsToolset.java
```diff
@@ -0,0 +1,98 @@
+package com.vaadin.fusion.parser.core;
+
+import java.util.Optional;
+import java.util.SortedSet;
+
+public class PluginsToolset {
```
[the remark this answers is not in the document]
Lodin: Hmm, interesting idea. Will try

packages/java/parser-jvm-plugin-model/src/test/java/com/vaadin/fusion/parser/model/utils/TestBase.java
haijian-vaadin: ok, I pushed a change here, at least locally the tests are fine.

packages/java/parser-jvm-core/src/main/java/com/vaadin/fusion/parser/core/PluginsToolset.java
```diff
@@ -0,0 +1,52 @@
...
+    private final SortedSet<Plugin> plugins;
+
+    public PluginsToolset(SortedSet<Plugin> plugins) {
+        this.plugins = plugins;
+    }
+
+    public Optional<Integer> comparePluginOrders(Plugin current,
```
haijian-vaadin, line 13: Do we really Optional wrapper? Could we just use `null` for `Optional.empty()`?

Lodin: `PluginsToolset` is supposed to be used by the developer who works on their own plugin. So I decided to make it clearer.

haijian-vaadin: ok

haijian-vaadin (approved):

PR #181, opened 2022-01-13
refactor(java/parser-jvm-plugin-backbone): use fields instead of getters for OpenAPI of data entities

### LINKED ISSUE #147: Use fields instead of getters for POJO code generation
The current implementation considers only the methods that start with `get`. While it has the benefit of e.g. getting the calculated properties, it also has the limitation of relying on a specific naming pattern. A better way would be to use fields for now so it aligns with the old generator. Then on top of that, we can improve it to also consider getters with special annotations as described in https://github.com/vaadin/fusion/issues/146.

Lodin:
Resolves #147


packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/fusion/parser/plugins/backbone/EntityProcessor.java
```diff
@@ -125,10 +121,10 @@ public String getKey() {
         private final String key;
         private final Schema<?> value;
 
-        public ComponentSchemaPropertyProcessor(RelativeMethodInfo method) {
-            this.key = decapitalize(method.get().getName().substring(3));
-            this.value = new SchemaProcessor(method.getResultType(),
-                    associationMap).process();
+        public ComponentSchemaPropertyProcessor(RelativeFieldInfo method) {
```
haijian-vaadin, line 124: here the param name is still `method`

packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/fusion/parser/plugins/backbone/EntityProcessor.java
```diff
@@ -125,10 +121,10 @@ public String getKey() {
...
+            this.key = method.get().getName();
+            this.value = new SchemaProcessor(method.getType(), associationMap)
```
haijian-vaadin, line 126: also here

haijian-vaadin (changes requested):

haijian-vaadin (approved):

PR #182, opened 2022-01-16
fix(form): Cross field validation returning the property field as a s…

vicb:
...tring

## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes # https://github.com/vaadin/flow/issues/9460

## Type of change

- [X] Bugfix - I think
- [ ] Feature

## Checklist

Note that I have not done everything in the list yet. I'd like a quick discussion before on the idea of this PR.

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.

All the examples and tests for cross field validation return a model in the `property` field.

For example see the existing test:

```ts
      const recordValidator = {
        validate(value: Order) {
          if (value.customer.fullName === value.customer.nickName) {
            // here we return a model   
            return { property: binder.model.customer.nickName };
          }

          return true;
        },
        message: 'cannot be the same',
      };
```

However to do this you need to capture the binder which is not always practical.

In my use case I have several entities in the form that need the same cross field validation. I have created a class for my validator:

```ts
export class MyValidator implements Validator<MyValue> {
  async validate(value: MyValue) {
    if (...) {
      this.message = `...`;
      // Where `account` is a property of MyValue      
      return { property: 'account' };
    }

    return true;
  }
}
```

I think returning `account` is the best thing to do as there is no way to know the parent name without the binder/model being captured.

I would like to know what you think about this.
If you agree with the idea could you please also take a quick look at the implementation ?

Tests missing from this PR (if we agree on the idea):
- returning a string property a top level,
- returning an array of `ValidationResult`.

Thanks !

### DIFF
--- a/packages/ts/form/test/Validation.test.ts
+++ b/packages/ts/form/test/Validation.test.ts
@@ -17,6 +17,7 @@ import {
   ValueError,
 } from '../src';
 import {
+  Customer,
   IdEntity,
   IdEntityModel,
   Order,
@@ -273,6 +274,57 @@ describe('form/Validation', () => {
         const crossFieldError = errors.find(byPropertyName('customer.nickName'));
         expect(crossFieldError).not.to.be.undefined;
         expect(crossFieldError?.message).to.equal('cannot be the same');
+      });
+    });
+
+    it('record level cross field validation when the property is a string', async () => {
+      const byPropertyName = (value: string) => (error: ValueError<any>) => {
+        const propertyName = typeof error.property === 'string' ? error.property : binder.for(error.property).name;
+        return propertyName === value;
+      };
+
+      const recordValidator = {
+        validate(value: Order) {
+          if (value.customer.fullName === value.customer.nickName) {
+            return { property: 'customer.nickName' };
+          }
+
+          return true;
+        },
+        message: 'cannot be the same',
+      };
+      binder.addValidator(recordValidator);
+
+      binder.for(binder.model.customer.fullName).value = 'foo';
+      await binder.validate().then((errors) => {
+        const crossFieldError = errors.find((error) => error.validator === recordValidator);
+        expect(crossFieldError, 'recordValidator should not cause an error').to.be.undefined;
+      });
+
+      binder.for(binder.model.customer.nickName).value = 'foo';
+      await binder.validate().then((errors) => {
+        const crossFieldError = errors.find(byPropertyName('customer.nickName'));
+        expect(crossFieldError).not.to.be.undefined;
+        expect(crossFieldError?.message).to.equal('cannot be the same');
+      });
+
+      const customerValidator = {
+        validate(value: Customer) {
+          if (Array.from(value.fullName).reverse().join('') === value.nickName) {
+            return { property: 'nickName' };
+          }
+
+          return true;
+        },
+        message: 'cannot be anagram of full name',
+      };
+      binder.for(binder.model.customer).addValidator(customerValidator);
+
+      binder.for(binder.model.customer.nickName).value = 'oof';
+      await binder.validate().then((errors) => {
+        const crossFieldError = errors.find(byPropertyName('customer.nickName'));
+        expect(crossFieldError).not.to.be.undefined;
+        expect(crossFieldError?.message).to.equal('cannot be anagram of full name');
       });
     });
   });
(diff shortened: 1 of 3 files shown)


haijian-vaadin:
Thanks a lot @vicb for the PR, we will take a look and get back to you.

platosha:
Hi @vicb, thanks for the PR!

I think, this idea is good primarily for remote (think server side) validators use case, where we the model JS reference isn’t available.

For the validators defined in JS, though, returning `{ property: model }` should still be preferred if possible.

We probably want to have good support for both cases. I think, in addition to this fix, maybe we could add a `binderNode: BinderNode<MyValue>` parameter to the `Validator`/`validate` method also later on. (No need to include that in this PR now.)

packages/ts/form/test/Validation.test.ts
```diff
@@ -275,6 +275,38 @@ describe('form/Validation', () => {
...
+    it('record level cross field validation when the property is a string', async () => {
+      const byPropertyName = (value: string) => (error: ValueError<any>) => {
+        const propertyName = typeof error.property === 'string' ? error.property : binder.for(error.property).name;
+        return propertyName === value;
+      };
+
+      const recordValidator = {
+        validate(value: Order) {
+          if (value.customer.fullName === value.customer.nickName) {
+            return { property: 'customer.nickName' };
```
platosha, line 288: Seems that this test is only checking with the absolute property path: it's a top-level validator that receives the entire binder’s value.

In addition to that, we need a test case with a relative property string. Something like that:

```typescript
const customerValidator = {
  validate(value: Customer) {
    if (Array.from(value.fullName).reverse().join('') === value.nickName) {
      return { property: 'nickName' };
    }

    return true;
  },
  message: 'cannot be anagram of full name',
};
binder.for(binder.model.customer).addValidator(customerValidator);

/* ... */

binder.for(binder.model.customer.nickName).value = 'oof';
await binder.validate().then((errors) => {
  const crossFieldError = errors.find(byPropertyName('customer.nickName'));
  expect(crossFieldError).not.to.be.undefined;
  expect(crossFieldError?.message).to.equal('cannot be anagram of full name');
});
```

platosha (changes requested):

vicb:
@platosha thanks a lot for the review and the great comments. I have updated the PR.

Off topic: one of the problem I have using the form lib server side is that the binder pulls a dependency on the  DOM via `getDefaultFieldStrategy()`. Are you interested in a PR breaking that dependency ? (At least to be used as a base for discussion). Thanks

platosha (approved):

haijian-vaadin:
> Off topic: one of the problem I have using the form lib server side is that the binder pulls a dependency on the DOM via getDefaultFieldStrategy(). Are you interested in a PR breaking that dependency ? (At least to be used as a base for discussion). Thanks

Yes, of course, please feel free to open a (draft) PR for that, we can discuss more based on the PR. 
btw, it also requires you to sign the cla for us to merge the PR, thanks again for the contribution.

platosha:
> Off topic: one of the problem I have using the form lib server side is that the binder pulls a dependency on the DOM via getDefaultFieldStrategy(). Are you interested in a PR breaking that dependency ? (At least to be used as a base for discussion). Thanks

Sounds like a good idea for me. Do you have some use case where it could help, like workers or SSR? Would be great to know about those, by the way. I think, we haven’t considered so far that the form binder could be somehow be used without DOM.

vicb:
@haijian-vaadin @platosha I use the form framework on the server side (I have a node.js server). It is the best form framework I have found out there - nice clean code, powerful, and easy to use. My server side augments the client side validation with [a few async validators](https://github.com/vicb/flyxc/blob/c2717842fd0e92b7a11437ea65f5b19d8f105af6/app/src/routes/live-track.ts#L121-L142).

The only 2 updates I had to do to make it work is the current PR and breaking the dependency on the DOM (I have a [`NoDomBinder`](https://github.com/vicb/flyxc/blob/master/common/src/vaadin/form/NoDomBinder.ts) and made [`Binder`](https://github.com/vicb/flyxc/blob/master/frontend/src/vaadin/form/Binder.ts) extends it).

haijian-vaadin:
Hi @vicb, can you help to sign the Contributor License Agreement (cla) so that we can merge this PR? You can sign it by clicking on the details link of the cla check.
image

vicb:
I just signed the cla

Issue #200, opened 2022-02-07
Support Pageable, Page and UUID in endpoints with the new generator


haijian-vaadin:
The old generator supports Pageable, Page, and UUID which makes it easier to integrate with Spring Data. Implemented in https://github.com/vaadin/flow/pull/11648.
The new generator should have the same feature.

PR #202, opened 2022-02-10
refactor: Redesign core model system to support both ClassGraph and Java Reflection

Lodin:
Prepares the core model system to address #200.


packages/java/parser-jvm-plugin-pageable/src/main/java/dev/hilla/parser/plugins/pageable/PageablePlugin.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.annotation.Nonnull;
+
+import dev.hilla.parser.core.Plugin;
+import dev.hilla.parser.core.PluginsToolset;
+import dev.hilla.parser.core.RelativeClassInfo;
+import dev.hilla.parser.core.SharedStorage;
+import dev.hilla.parser.plugins.backbone.BackbonePlugin;
+import dev.hilla.parser.utils.PluginException;
+
+public class PageablePlugin implements Plugin.Transformer {
```
platosha, line 14: `Plugin.Transformer` seems not available yet. Intentional?

If not, how about introducing the entire pageable module as a separate PR?

platosha (changes requested):

PR #203, opened 2022-02-15
add dispatch workflow

haijian-vaadin:
## Description

Add a manual workflow that triggers the release

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,9 +4,11 @@ on:
   workflow_dispath:
     input:
       releaseVersion:
-        description: Release version
+        description: Release Version
         required: true
-
+      developmentVersion:
+        description: Next Development Version
+        required: true
 job:
   releasing:
     name: Release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Checkout Project Code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Java for publishing to Maven Central Repository
+        uses: actions/setup-java@v2
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+          server-id: github
```
Lodin, line 28: What is that field for? In the documentation it is described as "Value of the `distributionManagement/repository/id` field of the `pom.xml`" but I don't see `distributionManagement` field in Hilla's `pom.xml`

Lodin (approved):

.github/workflows/release.yml
haijian-vaadin: right, removed it for now, the `distributionManagement` is configured in another branch. Will first get this manual workflow working, then connect the dots.

PR #371, opened 2022-04-06
feat(parser-jvm-plugin-transfertypes): add UUID -> String replacer

### LINKED ISSUE #200: Support Pageable, Page and UUID in endpoints with the new generator

Lodin:
Resolves #200.

Based on #362, so should be merged after it.


packages/java/parser-jvm-plugin-transfertypes/src/main/java/dev/hilla/parser/plugins/transfertypes/UUIDReplacer.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.UUID;
+
+import dev.hilla.parser.core.ClassMappers;
+
+final class UUIDReplacer implements Replacer {
+    private ClassMappers classMappers;
+
+    @Override
+    public void process() {
+        classMappers.add(createReplacer(UUID.class, String.class));
```
platosha, line 14: Now that the `Replacer` interface is introduced, it’s a bit confusing here that `createReplacer` returns something else. Should we rename `createReplacer` to `createMapper`?

platosha (approved):

Issue #208, opened 2022-02-16
Move the fusion-endpoint module from Flow to Hilla repo


haijian-vaadin:
The [fusion-endpoint](https://github.com/vaadin/flow/tree/master/fusion-endpoint) module contains the endpoint controller and the code generator. Now we have the new generator under development, once it's ready, we should move the fusion-endpoint module to hilla.

DoD:
- Flow no longer has the `fusion-endpoint` module
- The Fusion-specific security tests in the [flow-tests/vaadin-spring-tests](https://github.com/vaadin/flow/tree/master/flow-tests/vaadin-spring-tests) module are moved from the Flow repo to Hilla repo.

manolo:
Remember to find all occurrences of fusion in sources and pom files, and fix them as much as possible

PR #397, opened 2022-04-29
feat: add fusion-endpoint module

platosha:
Connected with: #208


platosha:
Closing in favour of #432 that also brings IT tests.

PR #405, opened 2022-05-10
feat: Emit state change events from FluxConnection

Artur-:
These can be used to implement reconnect logic for Fluxes but also to update the UI to indicate that server pushes are not being received


platosha (approved):

cromoteca (approved):

packages/ts/hilla-frontend/src/FluxConnection.ts
```diff
@@ -116,4 +153,33 @@ export class FluxConnection {
...
     return hillaSubscription;
   }
+
+  /**
+   * Adds a listener for the given event type.
+   *
+   * @param type  the type of event
+   * @param listener  the listener to call when the event occurs
+   */
+  addEventListener(type: string, listener: (event: CustomEvent) => void) {
```
platosha, line 163: Any reasons for not extending `EventTarget`? Supported both in browsers and Node.js even.

Artur-: Tried to do that, seems to work. Not sure if all the uglyness is needed though

platosha: `implements EventTarget` is probably not needed, but does `extends EventTarget` work and make code more compact?

Artur-: Done

platosha (approved):

PR #421, opened 2022-05-13
feat(form): support vaadin-combo-box string value / use model in field strategy

### LINKED ISSUE #151: Combo box binder value not correctly set when using objects as items
### Description

When the items property of a combo box is populated with objects, the value set by the binder is the entire object and not the value specified by the `item-value-path` property.

The name is shown correctly in the drop down (as specified with the `item-label-path`) and the value in the value-changed event is also correct.

But when inspecting the entity in `submitTo`, the value of the property in the entity is set to the entire object.

```
<vaadin-combo-box 
   label="Last name" ${field(this.binder.model.lastName)}
   .items="${[{name: 'Name1', value: 'Value1'}, {name: 'Name2', value: 'Value2'}]}" 
   item-label-path="name" 
   item-value-path="value" 
   @value-changed="${(e:CustomEvent) => console.log(e.detail.value)}">
</vaadin-combo-box>
```

https://stackoverflow.com/questions/70288099/issue-with-object-items-in-vaadin-fusion-combo-box

### Expected outcome

In the example, I expected lastName to have the value "Value1" when I select "Name1" in the dropdown.

### Actual outcome

The actual value are {name: 'Name1', value: 'Value1'} when I select Name1 in the dropdown.

https://github.com/rna2i/vaadin-combobox-binder-test

### Steps to reproduce

1. Compile and open the example app.
2. Select a value for the last name
3. Press the save button
4. Check the console for output

### Environment

Vaadin v22 and Java 8 on Mac 11.6.1-

### Browsers affected

- Chrome 96.0.4664.93

platosha:
Fixes #151

Enables using models in the field strategy. The `<vaadin-combo-box>` strategy
uses `value` property when binding with models of primitive types
(boolean, string, number) and `selectedItem` for object and array types.

### DIFF
--- a/packages/ts/form/src/Binder.ts
+++ b/packages/ts/form/src/Binder.ts
@@ -239,8 +239,8 @@ export class Binder<T, M extends AbstractModel<T>> extends BinderNode<T, M> {
    *
    * @param elm the bound element
    */
-  public getFieldStrategy(elm: any): FieldStrategy {
-    return getDefaultFieldStrategy(elm);
+  public getFieldStrategy<T>(elm: any, model?: AbstractModel<T>): FieldStrategy {
+    return getDefaultFieldStrategy(elm, model);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


cromoteca (approved):
I verified that by changing how the combo box is bound to the model (`${field(this.binder.model...)}`), the correct value is passed, which means the whole object or the chosen property.

PR #446, opened 2022-06-21
fix(parser): address JDK11 bug

### LINKED ISSUE #445: JDK11 bug breaks nested annotations
According to the [JDK11 issue](https://bugs.openjdk.org/browse/JDK-8198526) there is a bug in `java.lang.reflect` that breaks the correct access to annotations of nested classes when you access it with `getAnnotatedOwnerType`.

So, let's take that you have the following annotation:

```java
// Foo.java
class Foo {
    A.B.@Foo C bar;
}

// A.java
class A {
    static class B {
        static class C {}
      }
}
```
You want to access annotations of `B` part assuming that it is empty (`@Foo` applies to `C` part, right?). But when you try to access it with the `getAnnotatedOwnerType()`, you find that `@Foo` annotation is stored not only in `C` part, but also in `B` and even in `A`. 

That is the bug of the JDK11 which could affect the code produced by the generator because generator does not distinguish the code by how it is written.

Lodin:
Fixes #445.

The utility wrapper for `getAnnotatedOwnerUtils` allows collecting all the annotations for all owners at once and then distribute them correctly as modern JDK (e.g. JDK17) does. 

The wrapper implements the following algorithm:

```java
Owner.CurrentType.@Foo Child bar;
```

* if `Owner` exists AND (`CurrentType` is static OR `Owner` is static and parametrized) — then annotations for that type are stored in `Owner`.
* if `Child` exists AND ((it is dynamic AND `CurrentType` is dynamic) OR (it is dynamic AND `CurrentType` is static but not parametrized)) — then annotations for that type are `Child`'s annotations and shouldn't be used.
* if `Child` and `CurrentType` are static — getting the owner annotations should be skipped because for static parts all annotations shared between all the parts

### DIFF
--- a/packages/java/parser-jvm-utils/src/main/java/dev/hilla/parser/utils/AnnotatedOwnerUtils.java
+++ b/packages/java/parser-jvm-utils/src/main/java/dev/hilla/parser/utils/AnnotatedOwnerUtils.java
@@ -49,8 +49,8 @@ public final class AnnotatedOwnerUtils {
      *
      * <li>if {@code Child} exists AND ((it is dynamic AND {@code CurrentType}
      * is dynamic) OR (it is dynamic AND {@code CurrentType} is static but not
-     * parametrized)) — then annotations for that type are {@code Child}'s annotations
-     * and shouldn't be used.</li>
+     * parametrized)) — then annotations for that type are {@code Child}'s
+     * annotations and shouldn't be used.</li>
      *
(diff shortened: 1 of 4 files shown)


cromoteca (approved):
I still need to have better understanding of the generator, but for what I can see here, that's OK for me

platosha (approved):

PR #449, opened 2022-06-23
fix: rename the `init` parameter used to cancel requests to avoid name clashes

cromoteca:
Unfortunately renaming `endpointRequestInit` to `init` leads to collisions if a developer uses `init` as name for a parameter in and `Endpoint` method. This patch restores the long name to minimize the risks.


Lodin:
What do you think about using `_init` variable name instead? Special symbol looks more specific for me.

BTW, we should also put a name check to make the generator throw an error if one of the user-declared parameters have the same name as our `init` parameter

cromoteca:
> What do you think about using `_init` variable name instead? Special symbol looks more specific for me.

Why not? Personally, a long parameter name does not hurt me, especially given that the user won't have to write it in his code. It just shows up in code completion, so I don't care as much. We should choose the name that limits clashes better.

> BTW, we should also put a name check to make the generator throw an error if one of the user-declared parameters have the same name as our `init` parameter

I'd leave the old gen as it is. For the multi-module one, given that the code is built in its finest details without the need of a ugly template as before, I think we can modify the name of the additional parameter if it clashes.

Lodin:
> Why not? Personally, a long parameter name does not hurt me, especially given that the user won't have to write it in his code. It just shows up in code completion, so I don't care as much. We should choose the name that limits clashes better.

Well, it's not about long names, actually. I just try to calculate the possibility that user would name their parameter either `_init` or `endpointInitRequest`. The one with underscore looks less affected: as far as I know, it is against the Java styleguide.

I would even consider adding two underscores as Lit team does for their private properties, like `__init`.

cromoteca:
Waiting for #451 to be merged as the changed files will be moved along

PR #451, opened 2022-06-23
refactor!(fusion-endpoint): align module name, groupId, version

### LINKED ISSUE #450: Align groupId and version for the fusion-endpoint module
As [pointed out here](https://github.com/vaadin/hilla/pull/432#discussion_r904623423), having a Java module a different groupId `com.vaadin` complicates deployment to Sonatype Nexus.

Solution: use the `dev.hilla` groupId for the `fusion-endpoint` module.

In addition, let us change the version to `1.2-SNAPSHOT` to align with other modules in the repository.

platosha:
Fixes #450


manolo:
I don't see any need for the `backend` prefix in the package name, none these java packages are not supposed for running in the frontend, I think just `endpoint` is enough. Another option would be `hilla-backend` to align with the `ts` folder, but I'd rather `endpoint` 

The same with the `hilla-tests` module, I would just rename to `integration-tests` or just `tests` because `hilla`prefix is not used anywhere else.

My proposal:
- endpoint
- tests

Screenshot 2022-06-24 at 08 26 49

Regarding `tests` folder structure I would remove the `hilla` and `test` parts and leave as:
- spring
- csfr
- csfr-context

Screenshot 2022-06-24 at 08 27 08

manolo:
> I’m not sure about dropping the `test` part. In certain contexts, I’m presented with only the module name, and it is confusing to have IT module names mixed with the rest. Here is the module list in my IDEA, where `endpoint` and former `test-endpoints` could easily be mixed up, for example:

Well, that was a proposal. I don't use Idea, but if you think keeping the test- prefix facilitates DX, then go ahead,
The only thing is that all modules are aligned

cromoteca (approved):
Lots of renaming and formatting, good job!

PR #452, opened 2022-06-24
fix(parser): make annotation model consider its parameters

Lodin:
Before, `AnnotationInfoModel` didn't pay attention to its parameters. So, the following annotations were equal:
```java
@Foo(name = "Bar", effect = SomeEffect.class)
```
and
```java
@Foo
```
It was also impossible to get these parameters out of an `AnnotationInfoModel`.

This is clearly not correct. This PR fixes this bug and allows annotation to consider parameters.


packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/AnnotationInfoReflectionModel.java
```diff
@@ -1,35 +1,71 @@
...
+        if (parameters == null) {
+            var methods = origin.annotationType().getDeclaredMethods();
+
+            parameters = new HashSet<>(methods.length);
+
+            for (var method : methods) {
+                try {
+                    parameters.add(AnnotationParameterModel.of(method.getName(),
+                            method.invoke(origin)));
+                } catch (InvocationTargetException | IllegalAccessException e) {
```
cromoteca, line 58: Checked exceptions rethrown as runtime exceptions: is this by design?

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/AnnotationParameterEnumValueSourceModel.java
```diff
@@ -0,0 +1,60 @@
...
+    public ClassInfoModel getClassInfo() {
+        try {
+            if (classInfo == null) {
+                classInfo = ClassInfoModel
+                        .of(Class.forName(origin.getClassName()));
+            }
+
+            return classInfo;
+        } catch (ClassNotFoundException e) {
+            throw new ModelException(e);
```
cromoteca, line 42: Checked exceptions rethrown as runtime exceptions: is this by design?

cromoteca: Following discussion, it is OK since those exceptions shouldn't happen, so it is correct to report as RE if something happens here.

cromoteca (approved):

PR #453, opened 2022-06-25
refactor(parser): reduce repeating code in models

Lodin:
This PR applies DRY principles to models system by adding additional abstraction that contains all repeating code for underlying classes. 

Depends on #452 and should be merged after it.


packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/MethodParameterInfoReflectionModel.java
```diff
@@ -44,43 +29,32 @@ public String getName() {
...
 
     @Override
-    public SignatureModel getType() {
-        if (type == null) {
-            type = SignatureModel.of(origin.getAnnotatedType());
-        }
-
-        return type;
+    public boolean isMandated() {
+        return (origin.getModifiers() & '耀') != 0;
```
cromoteca, line 38:
```suggestion
        return (origin.getModifiers() & 0x8000) != 0;
```

cromoteca (changes requested):

cromoteca (approved):

cromoteca (approved):

PR #455, opened 2022-06-27
feat(generator): add `init` param to generated endpoints

### LINKED ISSUE #454: Multi-module generator does not add init parameter to generated endpoints
Issue #435 also applies to the new generator: the init parameter is missing at the moment

cromoteca:
## Description

The implemented optional `init` parameter, used to cancel endpoint request` is ported to the multi-module generator.

Fixes #454


packages/ts/generator-typescript-plugin-backbone/src/EndpointMethodRequestBodyProcessor.ts
```diff
@@ -18,54 +18,87 @@ export type EndpointMethodRequestBody = ReadonlyDeep<OpenAPIV3.RequestBodyObject
...
+            undefined,
+            undefined,
+            undefined,
+            this.#defaultInitParamName,
+            ts.factory.createToken(ts.SyntaxKind.QuestionToken),
+            ts.factory.createTypeReferenceNode(this.#endpointRequestInitIdentifier),
+          ),
+        ],
+        packedParameters: ts.factory.createObjectLiteralExpression(),
+        initParam: ts.factory.createIdentifier(this.#defaultInitParamName),
```
Lodin, line 58: I guess, you don't need to make it unique manually:
```suggestion
        initParam: ts.factory.createUniqueIdentifier(this.#defaultInitParamName, ts.GeneratedIdentifierFlags.Optimistic),
```

I'm not completely sure about `ts.GeneratedIdentifierFlags.Optimistic` flag because it is not documented anywhere, so it probably would require a couple of tests before it satisfy our needs.

packages/ts/generator-typescript-plugin-backbone/src/EndpointMethodRequestBodyProcessor.ts
```diff
@@ -18,54 +18,87 @@ export type EndpointMethodRequestBody = ReadonlyDeep<OpenAPIV3.RequestBodyObject
...
   readonly #dependencies: DependencyManager;
   readonly #owner: Plugin;
   readonly #requestBody?: EndpointMethodRequestBody;
+  readonly #endpointRequestInitIdentifier: ts.Identifier;
 
   public constructor(
     requestBody: ReadonlyDeep<OpenAPIV3.ReferenceObject | OpenAPIV3.RequestBodyObject> | undefined,
     dependencies: DependencyManager,
     owner: Plugin,
+    endpointRequestInitIdentifier: ts.Identifier,
```
Lodin, line 36: I suggest this name to have `Type` part somewhere. For the first time, I was confused it with the parameter name. 
```suggestion
    endpointRequestInitTypeIdentifier: ts.Identifier,
```
Or we could simplify here as well:
```suggestion
    initTypeIdentifier: ts.Identifier,
```

packages/ts/generator-typescript-plugin-backbone/src/EndpointMethodRequestBodyProcessor.ts
```diff
@@ -18,54 +18,87 @@ export type EndpointMethodRequestBody = ReadonlyDeep<OpenAPIV3.RequestBodyObject
...
+        ],
+        packedParameters: ts.factory.createObjectLiteralExpression(),
+        initParam: ts.factory.createIdentifier(this.#defaultInitParamName),
       };
     }
 
     const parameterData = this.#extractParameterData(this.#requestBody.content[defaultMediaType]?.schema);
+    let initParamName = this.#defaultInitParamName;
+
+    while (parameterData.map(([name]) => name).includes(initParamName)) {
```
Lodin, line 65: You should use `find` function here:
```suggestion
    while (parameterData.find(([name]) => name === initParamName))) {
```

**P.S.** Though, if you go with `createUniqueIdentifier`, this part will disappear.

packages/ts/generator-typescript-plugin-backbone/test/name-clash/NameClash.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+import { expect, use } from 'chai';
+import sinonChai from 'sinon-chai';
+import BackbonePlugin from '../../src/index.js';
+import { createGenerator, loadInput } from '../utils/common.js';
+
+use(sinonChai);
+use(snapshotMatcher);
+
+describe('BackbonePlugin', () => {
+  context('when primitive types are used', () => {
```
Lodin, line 12: Probably, this description should have something about name clash?

packages/ts/generator-typescript-plugin-backbone/src/EndpointMethodRequestBodyProcessor.ts
cromoteca: I'm a bit puzzled about this one because I don't even get `createUniqueIdentifier` in the code completion and it is marked as not existing if I write it.

cromoteca: Found: `createUniqueName`

packages/ts/generator-typescript-plugin-backbone/test/name-clash/NameClash.spec.ts
cromoteca: Yes, I forgot that

packages/ts/generator-typescript-plugin-backbone/src/EndpointMethodRequestBodyProcessor.ts
cromoteca: Problem is: it's not aware of other parameters' names, so I can make it fail quite easily

Lodin: Meh, that's what I was afraid of 😩 
BTW, that also means that my approach for using `createUniqueName` is insufficient. That's not good.

Lodin: Ok, I gonna address it in the different PR, let's have it this way for now.

packages/ts/generator-typescript-plugin-backbone/src/EndpointMethodRequestBodyProcessor.ts
```diff
@@ -18,54 +18,88 @@ export type EndpointMethodRequestBody = ReadonlyDeep<OpenAPIV3.RequestBodyObject
 export type EndpointMethodRequestBodyProcessingResult = Readonly<{
   parameters: readonly ParameterDeclaration[];
   packedParameters?: ObjectLiteralExpression;
+  initParam: ts.Identifier;
 }>;
 
 export default class EndpointMethodRequestBodyProcessor {
+  readonly #defaultInitParamName = 'init';
```
Lodin, line 25: I think we could make it `static` for consistency: I usually follow the logic that since the field cannot be changed during class instantiation, it is unnecessary for it to be a non-static member.

Lodin (changes requested):

packages/ts/generator-typescript-plugin-backbone/src/EndpointMethodRequestBodyProcessor.ts
```diff
@@ -18,54 +18,88 @@ export type EndpointMethodRequestBody = ReadonlyDeep<OpenAPIV3.RequestBodyObject
 export type EndpointMethodRequestBodyProcessingResult = Readonly<{
   parameters: readonly ParameterDeclaration[];
   packedParameters?: ObjectLiteralExpression;
+  initParam: ts.Identifier;
 }>;
 
 export default class EndpointMethodRequestBodyProcessor {
+  static readonly #defaultInitParamName = 'init';
```
platosha, line 25:
```suggestion
  static readonly #defaultInitParamName = '__init';
```

cromoteca: That works differently in the multi-module generator: the default name is `init` and is modified if there's another parameter with the same name, so there's no need to add the underscores by default.

Lodin (approved):

PR #459, opened 2022-06-29
feat: add binder support for vaadin-multi-select-combo-box

### LINKED ISSUE #456: Add Binder support for <vaadin-multi-select-combo-box>

Currently the Hilla Binder does not work with the newly added `<vaadin-multi-select-combo-box>` component. The component uses the `selectedItems` property for setting and getting the value, which is always an array of items (either primitives, or complex entities). However the default field strategy of the Hilla Binder assumes that the `value` property contains the value, which does not apply to this component.

The Binder should support the `<vaadin-multi-select-combo-box>` component. I think it should be enough to add a custom field strategy for this component that reads / writes the value to `selectedItems` instead of `value`.
I don't think any additional logic is required for mapping entities, it should be enough to dump an array of entities into the property, and send an array of entities to the server when submitting.

The component could maybe be refactored to use `value` instead. But given that there are already other field strategies the map the value property, I think this is an acceptable solution.

**Additional context**

N/A

sissbruecker:
## Description

Adds support for `<vaadin-multi-select-combo-box>` to the Binder by adding a `MultiSelectComboBoxFieldStrategy` that maps the value to the component's `selectedItems` property.

Closes #456 

## Type of change

- Feature


packages/ts/form/test/Field.test.ts
```diff
@@ -739,6 +740,52 @@ describe('form/Field', () => {
       });
     });
 
+    [
+      { model: binder.model.fieldArrayString as AbstractModel<any>, value: ['a', 'b'] },
+      { model: binder.model.fieldArrayModel as AbstractModel<any>, value: [{ idString: 'id' }] },
```
sissbruecker, line 745: Only the array types make sense here.

packages/ts/form/src/Field.ts
```diff
@@ -148,6 +158,8 @@ export function getDefaultFieldStrategy<T>(elm: any, model?: AbstractModel<T>):
       return new ComboBoxFieldStrategy(elm, model);
     case 'vaadin-list-box':
       return new SelectedFieldStrategy(elm, model);
+    case 'vaadin-multi-select-combo-box':
+      return new MultiSelectComboBoxFieldStrategy(elm, model);
```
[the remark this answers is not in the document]
cromoteca: Looks like we're not checking bound types at the moment. We'll have to update [the documentation](https://hilla.dev/docs/data-binding/vaadin-components/#multiple-selection) to explain how to use this component.

cromoteca (approved):

packages/ts/form/test/Field.test.ts
```diff
@@ -739,6 +740,52 @@ describe('form/Field', () => {
...
+    ].forEach(async ({ model, value }) => {
```
Lodin, line 746: This approach is unsupported by Idea (I cannot run test created this way using Idea's interface), so it would be better to use two tests that runs the generalized function.

cromoteca: Actually there are many other tests written like this one in the same file and maybe somewhere else. We should change them all if that's a problem.

Lodin (approved):

Issue #458, opened 2022-06-29
Multi-module generator fails on `reactor.core.publisher.Flux`


cromoteca:
While searching for the dependencies of `public org.reactivestreams.Publisher<V> apply(java.util.List<? extends org.reactivestreams.Publisher<?>>)`, the generator seems unable to resolve `V`:

```
...
Caused by: java.lang.IllegalArgumentException: Could not resolve V against parameters of the defining method or enclosing class
    at io.github.classgraph.TypeVariableSignature.resolve (TypeVariableSignature.java:118)
    at dev.hilla.parser.models.DependencyCollector$Source.collectTypeVariableSignature (DependencyCollector.java:195)
    at dev.hilla.parser.models.DependencyCollector$Source.collect (DependencyCollector.java:151)
    at dev.hilla.parser.models.DependencyCollector$Source.collectTypeArgument (DependencyCollector.java:187)
    at dev.hilla.parser.models.DependencyCollector$Source.collect (DependencyCollector.java:153)
    at dev.hilla.parser.models.DependencyCollector$Source.collectClassRefTypeSignature (DependencyCollector.java:182)
    at dev.hilla.parser.models.DependencyCollector$Source.collect (DependencyCollector.java:155)
    at dev.hilla.parser.models.DependencyCollector.collect (DependencyCollector.java:33)
    at dev.hilla.parser.models.SourceSignatureModel.getDependenciesStream (SourceSignatureModel.java:13)
    at dev.hilla.parser.models.MethodInfoModel.getResultDependenciesStream (MethodInfoModel.java:60)
    at dev.hilla.parser.models.MethodInfoModel.getDependenciesStream (MethodInfoModel.java:33)
    at java.util.stream.ReferencePipeline$7$1.accept (ReferencePipeline.java:271)
    at java.util.ArrayList$ArrayListSpliterator.forEachRemaining (ArrayList.java:1655)
    at java.util.stream.AbstractPipeline.copyInto (AbstractPipeline.java:484)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto (AbstractPipeline.java:474)
    at java.util.stream.ForEachOps$ForEachOp.evaluateSequential (ForEachOps.java:150)
    at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential (ForEachOps.java:173)
    at java.util.stream.AbstractPipeline.evaluate (AbstractPipeline.java:234)
[...]

PR #466, opened 2022-07-11
feat(build): use multi-module generator in integration tests

### LINKED ISSUE #414: Verify the multi-module generator with fusion-endpoint and end-to-end tests

cromoteca:
Needs vaadin/flow#14408 to run tests with both generator.

Fixes #414


cromoteca:
Fails due to #458

pom.xml
```diff
@@ -254,6 +254,41 @@
...
+            <configuration>
+              <executable>npm</executable>
+              <arguments>
+                <argument>run</argument>
+                <argument>build</argument>
+              </arguments>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
platosha, line 291: Is this still needed now that the Flow Maven plugin was fixed to run `TaskGenerateHilla` after npm install?

Should not be needed for CI at least, GitHub validation build job does those.

platosha (changes requested):

pom.xml
cromoteca: Integration tests fail locally without that. If it's not needed in CI and we're OK having to run those two commands by hand locally, we can remove those plugins.

platosha (approved):

PR #468, opened 2022-07-13
feat(generator): implement push support in the multi-module generator

### LINKED ISSUE #447: Support for Flux<T> and EndpointSubscription<T> in the multi-module engine
The multi-module generator does not support handling Push endpoints.

The old generator's parser supports those by mapping `Flux<T>` and `EndpointSubscription<T>` to a customized `ArraySchema` type.

The old TypeScript generator emits `client.subscribe()` instead of regular `client.call()` for those, and the return type is declared as `Subscription<T>`.

cromoteca:
As part of the process of aligning features between the two generators, this PR implements the code generation changes needed to support Push in endpoints.

Closes #447


cromoteca:
Blocked by vaadin/flow#14233

packages/java/parser-jvm-plugin-transfertypes/src/test/resources/dev/hilla/parser/plugins/transfertypes/push/openapi.json
```diff
@@ -0,0 +1,145 @@
...
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "array",
+                  "nullable": true,
+                  "items": {
+                    "type": "string",
+                    "nullable": true
+                  },
+                  "x-class-name": "reactor.core.publisher.Flux"
```
platosha, line 53: Apparently this refers to the original class name before the replacement. Is this intentional or just a natural consequence of things?

cromoteca: Yes, I need that to identify those methods. Maybe could be renamed to `x-original-class`? But in many cases it just re-emits the same class name as in the API (see the many changes to existing tests).

platosha: The name sounds good to me.

Lodin (changes requested):
* For Java: there are several things I would like to change.
* For TS: refactored code to make it look more like other plugins; also updated `DependencyManager` utility to avoid updating imports manually.

cromoteca:
> I guess we should have a transformed class name here. It looks more natural: the class translated is shown here. Otherwise, if it has API different to the original one, it could become a mess.

For some reason I cannot reply directly to that comment. What do you mean by "transformed class name"?

Lodin:
Ah, sorry, I meant, the replacer class name. So, if we have an original class as `reactor.core.publisher.Flux` and replaced it with a `dev.hilla.runtime.transfertypes.Flux`, I believe, the last one should be a `x-class-name`.

Lodin (approved):

PR #477, opened 2022-08-02
refactor(parser): fix & improve code according to final test results

Lodin:
This PR finalizes testing of parser models and introduces fixes and improvements according to test results.


packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/SchemaProcessor.java
```diff
@@ -132,11 +132,19 @@ private Schema<?> iterableSchema() {
...
+        var typeArguments = _type.getTypeArguments();
         var values = new SchemaProcessor(typeArguments.get(1), info, storage)
                 .process();
 
-        return nullify(new MapSchema(), true).additionalProperties(values);
+        var schema = nullify(new MapSchema(), true)
+                .additionalProperties(values);
+
+        if (type.isNonJDKClass()) {
+            schema.addExtension("x-classname", _type.getName());
```
cromoteca, line 144: Is `x-classname` used anywhere?

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/AnnotationInfoSourceModel.java
```diff
@@ -24,8 +25,11 @@ public String getName() {
     }
 
     @Override
-    protected ClassInfoModel prepareClassInfo() {
-        return ClassInfoModel.of(origin.getClassInfo());
+    protected Optional<ClassInfoModel> prepareClassInfo() {
+        var cls = origin.getClassInfo();
+
+        return cls != null ? Optional.of(ClassInfoModel.of(cls))
+                : Optional.empty();
```
platosha, line 32: Could use `Optional.map()`:

```suggestion
        return Optional.of(origin.getClassInfo()).map(ClassInfoModel::of);
```

Just nitpicking here, feel free to ignore.

cromoteca (approved):

platosha (approved):

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/AnnotationInfoSourceModel.java
Lodin: Yeah, I thought about that and chose the ternary just because it has one time less object creation than the `Optional.map`. Guess it is kind of a micro optimization though.

PR #500, opened 2022-08-11
fix(endpoint): set Flow resources directory for NodeTasks test

platosha:


platosha:
duplicate of #499 :-)

PR #504, opened 2022-08-15
fix(engine): add support for custom connect client

### LINKED ISSUE #485: The multi-module generator does not support using a custom client
The `endpoints-custom-client` test module fails when using the multi-module generator as it doesn't pick up the provided custom client.

It should mimic the existing feature as implemented at https://github.com/vaadin/hilla/blob/42ea5619a0d670b39ef330be68caaf0423c37ad0/packages/java/endpoint/src/main/java/dev/hilla/frontend/TaskGenerateEndpointImpl.java#L50-L54

cromoteca:
Adds a check to verify if a file named `frontend/connect-client.ts` exists, just like in the single-module generator.

Fixes #485

### DIFF
--- a/packages/ts/generator-typescript-cli/src/index.ts
+++ b/packages/ts/generator-typescript-cli/src/index.ts
@@ -46,7 +46,7 @@ const logger = new LoggerFactory({ verbose });
 const io = new GeneratorIO(outputDir, logger);
 
 const resolvedPlugins = await Promise.all(Array.from(new Set(plugins), (pluginPath) => io.loadPlugin(pluginPath)));
-const generator = new Generator(resolvedPlugins, logger, outputDir);
+const generator = new Generator(resolvedPlugins, { logger, outputDir });
 
 const files = await generator.process(await processInput(input, io));
 await io.cleanOutputDir();
--- a/packages/ts/generator-typescript-cli/src/index.ts
+++ b/packages/ts/generator-typescript-cli/src/index.ts
@@ -46,7 +46,7 @@ const logger = new LoggerFactory({ verbose });
 const io = new GeneratorIO(outputDir, logger);
 
 const resolvedPlugins = await Promise.all(Array.from(new Set(plugins), (pluginPath) => io.loadPlugin(pluginPath)));
-const generator = new Generator(resolvedPlugins, logger);
+const generator = new Generator(resolvedPlugins, logger, outputDir);
 
 const files = await generator.process(await processInput(input, io));
 await io.cleanOutputDir();
--- a/packages/ts/generator-typescript-core/src/Generator.ts
+++ b/packages/ts/generator-typescript-core/src/Generator.ts
@@ -15,11 +15,11 @@ export default class Generator {
   readonly #parser: SwaggerParser;
   readonly #outputDir: string | undefined;
 
-  public constructor(plugins: readonly PluginConstructor[], logger: LoggerFactory, outputDir?: string) {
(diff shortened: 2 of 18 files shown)


cromoteca:
Looks like the upgrade to Lit 2.3.0 (coming from Flow) broke a test which is not related to this PR.

packages/ts/generator-typescript-cli/src/index.ts
```diff
@@ -46,7 +46,7 @@ const logger = new LoggerFactory({ verbose });
 const io = new GeneratorIO(outputDir, logger);
 
 const resolvedPlugins = await Promise.all(Array.from(new Set(plugins), (pluginPath) => io.loadPlugin(pluginPath)));
-const generator = new Generator(resolvedPlugins, logger);
+const generator = new Generator(resolvedPlugins, logger, outputDir);
```
Lodin, line 49: Let's create a `Context` data structure that contains `outputDir` and `logger` to make it easier to change the public API if it is necessary

Lodin (changes requested):

packages/ts/generator-typescript-core/src/Generator.ts
```diff
@@ -13,11 +13,13 @@ export default class Generator {
   readonly #logger: LoggerFactory;
   readonly #manager: PluginManager;
   readonly #parser: SwaggerParser;
+  readonly #outputDir: string | undefined;
 
-  public constructor(plugins: readonly PluginConstructor[], logger: LoggerFactory) {
+  public constructor(plugins: readonly PluginConstructor[], context: { logger: LoggerFactory; outputDir?: string }) {
```
Lodin, line 18: It's better to have named types instead of object literals. Since it is a Generator context, let's delcare it just above the `Generator` class:
```ts
export type GeneratorContext = Readonly<{
  logger: LoggerFactory;
  outputDir?: string;
}>;
```

packages/ts/generator-typescript-plugin-backbone/src/index.ts
```diff
@@ -29,31 +31,34 @@ export default class BackbonePlugin extends Plugin {
...
-    const endpoints = new Map<string, EndpointProcessor>();
+    const endpoints = new Map<string, Map<string, ReadonlyDeep<OpenAPIV3.PathItemObject>>>();
 
-    for (const [path, pathItem] of Object.entries(storage.api.paths)) {
-      if (!pathItem) {
-        // eslint-disable-next-line no-continue
-        continue;
-      }
+    Object.entries(storage.api.paths)
+      .filter(([, pathItem]) => Boolean(pathItem))
```
Lodin, line 39:
```suggestion
      .filter(([, pathItem]) => !!pathItem)
```

packages/ts/generator-typescript-plugin-client/src/index.ts
```diff
@@ -1,18 +1,35 @@
...
+      sources.push(clientFile);
+    }
+  }
+
+  private static async checkForCustomClientFile(path?: string): Promise<boolean> {
+    const dir = path && path.startsWith('file:') ? fileURLToPath(path) : path;
+    return Boolean(dir && (await open(`${dir}/${ClientPlugin.CUSTOM_CLIENT_FILE_NAME}.ts`, 'r')));
+  }
+
+  public static async clientFileName(path?: string): Promise<string> {
```
Lodin, line 30:
```suggestion
  public static async getClientFileName(path?: string): Promise<string> {
```

Lodin (changes requested):

Lodin (approved):

PR #511, opened 2022-08-17
fix(parser-jvm-core): only parse entities of endpoint methods or entity fields

### LINKED ISSUE #487: Multi-module generator: leak of classes in the generated OpenAPI
When using the multi-module generator in the `tests/spring/security` module, many classes are added to the `components` sections of the OpenAPI definition. Those classes are not used by the endpoints to exchange data, but they are added anyway. They come from Spring, Reactor and so on.

Furthermore, the generator is not able to cope with some of those classes, thus making impossible to run the tests.

Let's take for example the [Balance](https://github.com/vaadin/hilla/blob/main/packages/java/tests/spring/security/src/main/java/com/vaadin/flow/spring/fusionsecurity/endpoints/BalanceEndpoint.java) endpoint: if `BankService` is removed and methods are kept by returning dumb values, that endpoint is generated correctly and does not add all those classes to the list, which otherwise are not generated correctly. Same applies to the other endpoints.

Here's the generated OpenAPI:
[security-openapi.zip](https://github.com/vaadin/hilla/files/9269749/security-openapi.zip)

platosha:
Fixes #487

### DIFF
--- a/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/ScanElementsCollector.java
+++ b/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/ScanElementsCollector.java
@@ -9,6 +9,7 @@ import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 
 import dev.hilla.parser.models.ClassInfoModel;
+import dev.hilla.parser.utils.Streams;
 
 import io.github.classgraph.ScanResult;
 
@@ -41,7 +42,7 @@ public final class ScanElementsCollector {
 
         entities = endpoints.stream()
                 .flatMap(cls -> cls.getInheritanceChainStream()
-                        .flatMap(ClassInfoModel::getDependenciesStream))
+                        .flatMap(ClassInfoModel::getMethodDependenciesStream))
                 .map(classMappers::map).filter(ClassInfoModel::isNonJDKClass)
                 .distinct().collect(Collectors.toList());
 
@@ -90,9 +91,14 @@ public final class ScanElementsCollector {
         for (var i = 0; i < entities.size(); i++) {
             var entity = entities.get(i);
 
-            entity.getDependenciesStream().map(classMappers::map)
-                    .filter(ClassInfoModel::isNonJDKClass)
-                    .filter(e -> !entities.contains(e)).forEach(entities::add);
+            Streams
+                .combine(entity.getFieldDependenciesStream(),
+                    entity.getSuperClassStream(),
+                    entity.getTypeParameterDependenciesStream())
+                .filter(ClassInfoModel::isNonJDKClass).distinct()
+                .map(classMappers::map)
+                .filter(ClassInfoModel::isNonJDKClass)
+                .filter(e -> !entities.contains(e)).forEach(entities::add);
         }
 
         return this;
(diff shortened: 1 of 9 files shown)


packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/ScanElementsCollector.java
```diff
@@ -41,7 +42,7 @@ public ScanElementsCollector collect() {
 
         entities = endpoints.stream()
                 .flatMap(cls -> cls.getInheritanceChainStream()
```
cromoteca, line 44: Should we also parse interfaces implemented by endpoint to find default methods?

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/ScanElementsCollector.java
```diff
@@ -90,7 +91,11 @@ public ScanElementsCollector collect() {
         for (var i = 0; i < entities.size(); i++) {
             var entity = entities.get(i);
 
-            entity.getDependenciesStream().map(classMappers::map)
+            Streams.combine(entity.getFieldDependenciesStream(),
+                    entity.getSuperClassStream(),
+                    entity.getTypeParameterDependenciesStream())
```
Lodin, line 96: Do we consider type parameters?

Lodin (approved):

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/ScanElementsCollector.java
cromoteca: I'd definitely say so: look at https://github.com/vaadin/hilla/blob/e2f4dc35147235e0ab6ec93ffb25dfbea334d6ee/packages/java/endpoint/src/test/java/dev/hilla/generator/endpoints/superclassmethods/PagedData.java#L27-L28

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/ScanElementsCollector.java
platosha: We don’t, removed

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/ScanElementsCollector.java
platosha: Added

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/ScanElementsCollector.java
```diff
@@ -40,8 +42,9 @@ public ScanElementsCollector collect() {
                 .collect(Collectors.toList());
 
         entities = endpoints.stream()
-                .flatMap(cls -> cls.getInheritanceChainStream()
-                        .flatMap(ClassInfoModel::getDependenciesStream))
+                .flatMap(cls -> Streams.combine(cls.getInheritanceChainStream(),
+                        cls.getInterfacesStream()))
```
cromoteca, line 46: Looks like this is not covered by tests. It is somewhat related to the upcoming test for `@EndpointExposed`: https://github.com/vaadin/hilla/pull/520/commits/995d788445b43ce844deebacae207c7c68165d74

platosha: True, but let us address it in https://github.com/vaadin/hilla/pull/520 maybe?

cromoteca (approved):

PR #522, opened 2022-08-18
feat: avoid running Maven at each file generation

### LINKED ISSUE #519: Multi-module engine: avoid running Maven at each file generation

Currently, the multi-module engine runs Maven every time source files are modified to update generated files. This is needed since the generation process needs the Maven configuration.

That configuration could be retrieved once and for all at the first Maven run (e.g. when the developer starts the dev server), and then we could run the generation code directly by using that configuration, without the need to run an external process each time.

**Additional context**

Running that process also opens the door to errors like #507.

cromoteca:
This PR adds a new Maven goal to the plugin. This goal stores the Maven configuration in a file.

Then, the Hilla task is modified to run Java code directly if that file is found.

Unfortunately, this feature requires adding the new goal to the POM file, so if the goal is not available, the configuration will not be written and the generation will still have to run Maven as before.

I think that this PR is a good idea, but it adds some complexity: let's discuss about it.

Closes #519


cromoteca:
At least I can confirm that using the `gather-configuration` goal allows to run the multi-module generator on Windows without issues.

packages/java/maven-plugin/src/main/java/dev/hilla/maven/PluginConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.PropertyAccessor;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class PluginConfiguration {
+    public static final String RESOURCE_NAME = "hilla-generator-configuration.json";
```
platosha, line 13:
```suggestion
    public static final String RESOURCE_NAME = "hilla-engine-configuration.json";
```

BTW should we name it with “engine”, as it combines parser and generator configuration?

platosha (dismissed):

PR #525, opened 2022-08-19
Fix non-null check

### LINKED ISSUE #523: @Nonnull is ignored in Hilla 1.2.0.beta2
### Description of the bug 
Given
```
    @Nonnull
    public Page<@Nonnull SamplePerson> list(Pageable page) {
        return service.list(page);
    }
```

the TS return type should be
```
Promise<Array<SamplePerson>>
```

It is
```
Promise<Array<SamplePerson | undefined> | undefined>
```

start.vaadin.com -> 23.2 (pre) -> Master-Detail using Hilla

### Versions:
Hilla: 23.2.0.beta2
Flow: 23.2.0.beta2
Java: Homebrew 18.0.2
OS: aarch64 Mac OS X 12.4
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36

cromoteca:
Fix how the check is performed to restore the old behavior and always honors `additionalAnnotations`.

Fixes #523

### DIFF
--- a/packages/java/endpoint/src/test/resources/dev/hilla/generator/endpoints/nonnullable/expected-NonNullableEndpoint.ts
+++ b/packages/java/endpoint/src/test/resources/dev/hilla/generator/endpoints/nonnullable/expected-NonNullableEndpoint.ts
@@ -56,7 +56,7 @@ function _getNotNullReturnType(__init?: EndpointRequestInit): Promise<ReturnType
   return client.call('NonNullableEndpoint', 'getNotNullReturnType', {}, __init);
 }
 
-function _returnNonnullMappedType(__init?: EndpointRequestInit): Promise<Array<string | undefined>> {
+function _returnNonnullMappedType(__init?: EndpointRequestInit): Promise<Array<string>> {
   return client.call('NonNullableEndpoint', 'returnNonnullMappedType', {}, __init);
 }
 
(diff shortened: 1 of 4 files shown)


packages/java/endpoint/src/test/resources/dev/hilla/generator/endpoints/nonnullable/expected-NonNullableEndpoint.ts
```diff
@@ -56,6 +56,10 @@ function _getNotNullReturnType(__init?: EndpointRequestInit): Promise<ReturnType
   return client.call('NonNullableEndpoint', 'getNotNullReturnType', {}, __init);
 }
 
+function _returnNonnullMappedType(__init?: EndpointRequestInit): Promise<Array<string | undefined>> {
```
platosha, line 59: It seems that the `@Nonnull` is still ignored for the type argument. Or am I missing something?

The endpoint returns `@Nonnull Page<@Nonnull String>`, so I’d expect the `Array<string>`, but not `Array<string | undefined>` in the return type for the generated method.

```suggestion
function _returnNonnullMappedType(__init?: EndpointRequestInit): Promise<Array<string>> {
```

platosha (changes requested):

packages/java/endpoint/src/test/resources/dev/hilla/generator/endpoints/nonnullable/expected-NonNullableEndpoint.ts
cromoteca: You're right and unfortunately this is a different issue: while the one I fixed was a regression in the method logic when adding support for `@NonNullApi`, here I see that the code is completely unaware of that annotation as the type has been mapped. I'm searching for a fix.

platosha (approved):

PR #534, opened 2022-08-29
fix(engine): pass OpenAPI JSON to generator using stdin to avoid size limitations

### LINKED ISSUE #533: Multi-module project support for TS generation exits with: The command line is too long.
### Description of the bug 
mvn -X hilla:generate gives: The command line is too long.
cmd.exe /c C:\my-proj\node_modules\.bin\tsgen.cmd -o ... is very long. It has enum values etc and will not scale 
Create some objects with enum etc 

### Versions:
Hilla: 23.2.0.beta2
Flow: 23.2.0.beta2
Java: JetBrains s.r.o. 17.0.2
OS: amd64 Windows 10 10.0
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/104.0.0.0 Safari/537.36
Live reload: Java unavailable (HotswapAgent): Front end unavailable

cromoteca:
This PR refactors existing code to pass OpenAPI JSON though standard input.

Fixes #533

### DIFF
--- a/packages/ts/generator-typescript-cli/src/index.ts
+++ b/packages/ts/generator-typescript-cli/src/index.ts
@@ -10,6 +10,8 @@ const {
 } = meow(
   `
 Usage:
+  tsgen
+    (will read JSON from stdin)
   tsgen <OpenAPI JSON string>
   tsgen <OpenAPI file path>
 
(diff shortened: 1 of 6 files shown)


packages/ts/generator-typescript-cli/src/index.ts
```diff
@@ -10,6 +10,8 @@ const {
 } = meow(
   `
 Usage:
+  tsgen
+    (will read JSON from stdin)
   tsgen <OpenAPI JSON string>
```
Lodin, line 15: I would suggest to remove this behavior since we start using `stdin`.

Lodin (approved):

packages/ts/generator-typescript-cli/src/index.ts
Lodin: I mean, passing OpenAPI JSON string as an argument

PR #535, opened 2022-08-31
refactor!(parser): revisit node walking for the parser

Lodin:

PR #553, opened 2022-09-19
refactor!(parser): replace numeric plugin order with runBefore / runAfter specification

platosha:
Depends on #535


packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/AbstractCompositePlugin.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public void setStorage(SharedStorage storage) {
+        super.setStorage(storage);
+        plugins.iterator()
+                .forEachRemaining(plugin -> plugin.setStorage(storage));
+    }
+
+    private void verifyPluginsOrder() {
```
cromoteca, line 49: I think that, given that the order is imposed by the new strategy, verifying it is useless

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/Plugin.java
```diff
@@ -1,32 +1,46 @@
...
+        if (runBefore().contains(o.getClass())) {
+            return -1;
+        }
+        if (o.runAfter().contains(getClass())) {
+            return -1;
+        }
+        if (o.runBefore().contains(getClass())) {
+            return 1;
+        }
+        return getClass().getName().compareTo(o.getClass().getName());
```
cromoteca, line 44: This should allow to test for invalid configurations (would need to add a test for that):

```suggestion
        int result = runBefore().contains(other.getClass())
                || other.runAfter().contains(getClass()) ? -1 : 0;
        if (runAfter().contains(other.getClass())
                || other.runBefore().contains(getClass())) {
            result ^= 1;
        }
        if (result == -2) {
            throw new IllegalStateException(String.format(
                    "Plugins %s and %s have conflicting order needs",
                    getClass().getName(), other.getClass().getName()));
        }
        return result;
```

cromoteca (changes requested):

PR #572, opened 2022-09-29
fix: generate default parameter names when they are not available in compiled classes

### LINKED ISSUE #571: Cannot run a project containing endpoints without parameter names in class files
When the project containing the endpoints is compiled without the `parameters` flag, ClassGraph always returns `null` as parameter name, whereas Java reflection generates `argN` placeholder names.

While it is recommended to add the compiler flag for readability, its absence should not break the application: we should generate the missing parameter names.

cromoteca:
Fixes #571 by adding a `getIndex` method in `MethodParameterInfoModel` and its subclasses, and using that method to generate missing parameter names in the form `argN`.

### DIFF
--- a/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/MethodParameterInfoModel.java
+++ b/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/MethodParameterInfoModel.java
@@ -49,6 +49,13 @@ public abstract class MethodParameterInfoModel extends AnnotatedAbstractModel
 
     public abstract int getModifiers();
 
+    /**
+     * Returns a zero-based index of the parameter in the method parameter list.
+     * 
+     * @return the index, from 0 to n-1
+     */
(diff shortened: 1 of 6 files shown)


packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/MethodParameterInfoModel.java
```diff
@@ -49,6 +49,13 @@ public Class<MethodParameterInfoModel> getCommonModelClass() {
 
     public abstract int getModifiers();
 
+    /**
+     * Returns a zero-based index of the parameter in the method parameter list.
+     * 
+     * @return the index, from 0 to n-1
+     */
+    public abstract int getIndex();
```
Lodin, line 57: I would suggest using the same memoization approach as used for other properties. It would also allow constructing the schema if necessary.

Lodin (approved):

PR #577, opened 2022-10-05
fix(form): use more reliable method to collect model children

### LINKED ISSUE #559: Form-Binder: Unreliable validation of nullable properties
### Description of the bug 
Having a form with a custom validator, the validator is only called the first time you enter a value for a nullable property

[my-hilla-app.zip](https://github.com/vaadin/hilla/files/9635330/my-hilla-app.zip)

This example will show you the bug.
It will only occur when you call _this.binder.read_ with the optional property missing.

### Expected behavior
Entered value is validated every time you change that value

### Actual behavior
The value is only validated the first time you enter a value to the field

### Versions:
    - Vaadin / Hilla version: 1.2.1

Lodin:
Fixes #559


packages/ts/form/src/BinderNode.ts
```diff
@@ -338,8 +338,8 @@ export class BinderNode<T, M extends AbstractModel<T>> {
       // from initial `binder.read()` or `binder.clear()` or by using a
       // binder node (e. g., form biding) for a nested field.
```
cromoteca, line 339: A good chance to fix the comment ;)

```suggestion
      // binder node (e. g., form binding) for a nested field.
```

packages/ts/form/test/Binder.test.ts
```diff
@@ -186,7 +186,7 @@ describe('form/Binder', () => {
       expect(requestUpdateStub).to.be.calledOnce;
     });
 
-    it('should update when clearing validation', async () => {
+    it.only('should update when clearing validation', async () => {
```
cromoteca, line 189: Is this intentional?

Lodin: Nope, just forgot to remove, thanks!

cromoteca (approved):

PR #597, opened 2022-10-13
fix: Use Atmosphere (esm) instead of socket.io for servlet 5 compatibility

Artur-:


Artur-:
based on #596

platosha (approved):

rob-64:
@Artur- I know this just changed from socketIo to Atmosphere... but, I found an issue with the latest hilla Hello World and creating a webSocket endpoint... it never makes it to the WebSocketHttpRequestHandler as the path that comes in in the request is /HILLA/push/ (with a trailing slash) but registration only does : `registry.addHandler(engineIoHandler, "/HILLA/push")` I assume this is a VAADIN flow thing changing the path.

As a test i added another WebSocketConfigurer and added the path with the trailing slash and it worked...

can we update it to registry.addHandler(engineIoHandler, "/HILLA/push", "/HILLA/push/")

PR #599, opened 2022-10-14
Servlet5

Artur-:


Artur-:
Depends on https://github.com/vaadin/flow/pull/14853

vursen:
> I wonder if the reason is Hilla still using VaadinWebSecurityConfigurerAdapter in ITs despite that it has been deprecated and VaadinWebSecurity should be used instead. I'm investigating that.

Yeah, using `VaadinWebSecurity` instead of the deprecated class helped get rid of this exception.

vursen:
The `spring/security-jwt` test failures will be resolved with the next Flow release.

Artur-:
We should sync the versions like annotation api etc with https://github.com/vaadin/flow/pull/14911

vursen (approved):

PR #606, opened 2022-10-19
fix(endpoint): websocket path matching is not matching because of trailing slash

rob-64:
old fusion-endpoint had trailing slash

## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes # (issue)

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

- [ ] I have added a description following the guideline.

### DIFF
--- a/packages/java/endpoint/src/main/java/dev/hilla/push/SocketIoWebsocketConfigurer.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/push/SocketIoWebsocketConfigurer.java
@@ -29,7 +29,7 @@ public class SocketIoWebsocketConfigurer implements WebSocketConfigurer {
 
     @Override
     public void registerWebSocketHandlers(WebSocketHandlerRegistry registry) {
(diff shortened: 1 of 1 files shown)


Artur-:
What is it not matching?

rob-64:
The websocket handler doesn't get activated on the http request because the path the browser sends ends with a/but the registration doesn't so it always falls back to pulling instead of web socket

Artur-:
At least the path here https://github.com/vaadin/hilla/blob/1.2/packages/ts/hilla-frontend/src/FluxConnection.ts#L51 does not use a trailing slash either

Artur-:
Seems like socket.io ALWAYS appends a `/` to the path...

rob-64:
Yea. So I couldn't get a web socket with reactive end points until I double registered it with the trailing slash

platosha (approved):

platosha (approved):

PR #570, opened 2022-09-28
(WIP) resolve circular dependency 

ZheSun88:
## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #623, opened 2022-10-28
chore: resolve circular dependency  (#570) (CP: 1.3)

ZheSun88:
* chore: resolve circular dependency

* chore: add lumo dependency

* chore: component view dependency

* chore: move component dependency to test root

* chore:add custom-field dependency


ZheSun88:
this will work after new WC 23.3.0.alpha has been released.

platosha (approved):

Issue #146, opened 2021-12-07
Consider annotated getters in the generation of TS models


simasch:
JPA and Jackson lets you set annotations on getters instead of fields.

**JPA**
For JPA that means if the getter has an annotation the field should not be used.

This can also be specified setting the `AccessType` with the `Access `annotation: https://javaee.github.io/javaee-spec/javadocs/javax/persistence/Access.html 
In that case the defined AccessType should be used to access the state of the entity.

**Jackson**
Jackson has the JsonProperty, JsonGetter and JsonSetter annotations. 
If these are used then also getter or setter should be used instead of the field.

Why do we want to use that?

There could be the case that we want to serialize a calculated field that is calculated in the getter of the DTO or Entity.
For example if we have a Order we want to calculate the total price of the order and serialize that to the client.

platosha:
Added in PR #720

Issue #140, opened 2021-12-02
Allow POJO Getters to be non-nullable


tarekoraby:
**Is your feature request related to a problem? Please describe.**
If my endpoint returns a POJO whose getters are marked to return a non-nullable type (e.g. using `@Nonnull`), Fusion disregard this annotation on the getters. I find it confusing that Fusion only considers the non-nullable annotations on my private fields, while it disregards the same annotation if it's placed on my public getters.

**Describe the solution you'd like**
Marking my POJO getters as non-nullable should have the same effect of doing the same thing for my POJO fields (i.e. Fusion should generate frontend types in which the corresponding fields are non-nullable)

PR #720, opened 2023-01-17
feat(parser-jvm): implement Jackson Properties

### LINKED ISSUE #322: @JsonProperty does not work as documented
The [current documentation for the `@JsonProperty` annotation](https://vaadin.com/docs/v15/flow/typescript/custom-serialization.html#the-jsonproperty-annotation) is misleading: in reality it seem to have no effect on the TypeScript type generator (as of Vaadin 15.0.4).

1. After reading the docs I would expect that `@JsonProperty("bookId") private String id;` in Java would lead to `bookId: string;` in TypeScript. This is not the case; the generated TS type has `id: string`.
2. After reading the docs I would expect that
   ```java
    public class Student {
      // NOTE: no field, only getter and setter

      @JsonProperty("name")
      public void setFirstName(String name) {
         this.name = name;
      }
  
      @JsonProperty("name")
      public String getFirstName() {
         return name;
      }
    }
    ```
    would lead to `name: string;` in TypeScript. This is not the case; the generated TS type has neither `name` nor `firstName` property at all.

> ### The `@JsonProperty` annotation
>
> It is used to define a method as a "setter" or "getter" for a logical property, or to define a field for being serialized and deserialized as a specific logical property.
> 
> The annotation value indicates the name of the property in the JSON object, by default it takes the java name of the method or field.
> ```java
>public class Student {
>   @JsonProperty("bookId")
>   private String id;
>   private String name;
>
>   @JsonProperty("name")
>   public void setFirstName(String name) {
>      this.name = name;
>   }
>
>   @JsonProperty("name")
>   public String getFirstName() {
>      return name;
>   }
>
>   @JsonProperty
>   public int getRating() {
>      return StudentRating.getRatingFor(name);
>   }
>}
>```

Lodin:
Fixes #322


packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PropertyInfoModel.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public int hashCode() {
+        return (getName().hashCode() + getOwner().hashCode()) ^ 0x73448be4;
+    }
+
+    public abstract boolean isTransient();
+
+    @Override
+    protected List<AnnotationInfoModel> prepareAnnotations() {
+        return Stream
```
cromoteca, line 100: I think you're doing too much here: in Jackson you should get the `BeanDescription` of your class and get all property details from there.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
```diff
@@ -328,6 +329,14 @@ public PackageInfoModel getPackage() {
         return pkg;
     }
 
+    public List<PropertyInfoModel> getProperties() {
```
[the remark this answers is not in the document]
platosha: Still `PropertyInfoModel` is a 👍, just not sure if `ClassInfoModel` should host the properties.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PropertyInfoReflectionModel.java
```diff
@@ -0,0 +1,55 @@
...
+        return origin;
+    }
+
+    @Override
+    public boolean hasGetter() {
+        return origin.hasGetter();
+    }
+
+    @Override
+    public boolean isTransient() {
```
cromoteca, line 27: We should no longer care about transient stuff.

platosha (commented):
Nice, looks really straightforward.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PropertyInfoReflectionModel.java
```diff
@@ -0,0 +1,55 @@
...
+        return !origin.hasGetter() && origin.getField().isTransient();
+    }
+
+    @Override
+    protected FieldInfoModel prepareField() {
+        return FieldInfoModel.of(origin.getField().getAnnotated());
+    }
+
+    @Override
+    protected Optional<MethodInfoModel> prepareGetter() {
```
cromoteca, line 37: Do we really need to know about getters? Jackson properties should already deal with them.

cromoteca (changes requested):

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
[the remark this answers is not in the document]
cromoteca: In Jackson, properties are in a [JavaType](https://github.com/FasterXML/jackson-databind/blob/2.15/src/main/java/com/fasterxml/jackson/databind/JavaType.java), which is basically a class, so it looks fine to me.

cromoteca:
When I prototyped a Jackson-based parser, I wrote [this class](https://github.com/cromoteca/jackson-parser/blob/main/src/main/java/dev/hilla/parser/Parser.java) which basically does most of the tasks. Please have a look at how it works and how it is tested, because it contains some hints about how to get the needed information without going the hard way of manually checking getters and other internal stuff.

Also, and considering the good practices of the Acceleration Kit project, I'd start thinking twice before approving PRs that do not have enough:

- tests
- Javadoc or TSDoc
- code comments

But of course that should be discussed in the team.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PropertyInfoModel.java
Lodin: Ok, I gonna replace it with `origin.getPrimaryMember()` then.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PropertyInfoReflectionModel.java
Lodin: Unfortunately, it's not true. By default, Jackson doesn't handle transient stuff; it just parses transient fields as they are regular fields which is not suitable for us.

cromoteca: That's true, but as we discussed some time ago, `transient` exists for on-disk object serialization and does not concern data transfer. Jackson correctly ships those fields. In my opinion we should do the same and don't hide transient fields. I know this is a breaking change, but that's the good moment to make those.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PropertyInfoReflectionModel.java
Lodin: Well, they're just helper methods in case anyone wants to use field or getter directly. It has the same goal as `get()` method in any other model but works with models.

Lodin:
@cromoteca I have taken a close look at your class but found that it serves a bit different purpose: it searches for endpoints and entities and makes them `BeanDescription` but have almost no interaction with `BeanPropertyDefinition`. Unfortunately, Jackson gives us not full experience with properties: e.g., it does not filter it according to `@JsonIgnoreProperties` annotation. So I need to do some manual stuff.

I gonna update project according to the comments, please, take another look.

cromoteca:
> I have taken a close look at your class but found that it serves a bit different purpose: it searches for endpoints and entities and makes them `BeanDescription` but have almost no interaction with `BeanPropertyDefinition`. Unfortunately, Jackson gives us not full experience with properties: e.g., it does not filter it according to `@JsonIgnoreProperties` annotation. So I need to do some manual stuff.

You're right, my bad, you're already doing what I was trying to suggest. I don't know why yesterday I had the impression that you were getting the property list differently.

packages/java/tests/spring/endpoints/src/main/java/com/vaadin/flow/connect/ObjectWithNullValues.java
```diff
@@ -2,4 +2,8 @@
 
 public class ObjectWithNullValues {
     private String propWithNullValue;
+
+    public String getPropWithNullValue() {
+        return propWithNullValue;
+    }
```
cromoteca, line 8: I'd try to avoid modifying existing tests and use a configured `ObjectMapper` instead. That would also allow to offer backwards compatibility as an option.

packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/BackbonePlugin.java
```diff
@@ -1,15 +1,23 @@
...
     public BackbonePlugin() {
         super(new EndpointPlugin(), new EndpointExposedPlugin(),
                 new MethodPlugin(), new MethodParameterPlugin(),
-                new EntityPlugin(), new FieldPlugin(),
-                new TypeSignaturePlugin(), new JSONPlugin());
+                new EntityPlugin(),
+                new PropertyPlugin(new ObjectMapper()
+                        .registerModule(new ParameterNamesModule())
+                        .registerModule(new Jdk8Module())
+                        .registerModule(new JavaTimeModule())),
```
cromoteca, line 20: I'd prefer use the same (configurable) `ObjectMapper` on both the parser and the controller

packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/PropertyPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+            this.description = description;
+        }
+
+        public Stream<BeanPropertyDefinition> stream() {
+            var properties = description.findProperties().stream();
+            properties = filterSuperClassProperties(properties);
+            properties = filterPropertiesWithIgnoredTypes(properties);
+
+            // Remove ones without matching mutator (if it is configured).
+            if (config.isEnabled(MapperFeature.REQUIRE_SETTERS_FOR_GETTERS)) {
```
cromoteca, line 100: What's the use case for `REQUIRE_SETTERS_FOR_GETTERS`?

packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/PropertyPlugin.java
```diff
@@ -0,0 +1,194 @@
...
+                    ignores.put(type, result);
+                }
+
+                return !result;
+            });
+        }
+
+        /**
+         * @see <a href=
+         *      "https://github.com/FasterXML/jackson-databind/blob/2.15/src/main/java/com/fasterxml/jackson/databind/ser/BeanSerializerFactory.java#L696">BeanSerializerFactory#removeSetterlessGetters</a>
```
cromoteca, line 164: Link line has changed 🙂

cromoteca (changes requested):

packages/java/endpoint/src/main/java/dev/hilla/EngineObjectMapperConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+    private static final Logger logger = LoggerFactory
+            .getLogger(EngineObjectMapperConfiguration.class);
+    static final String HILLA_ENGINE_FEATURE_FLAG = "hillaEngine";
+
+    @Bean
+    @Qualifier(EndpointController.VAADIN_ENDPOINT_MAPPER_BEAN_QUALIFIER)
+    public ObjectMapper createObjectMapper(ApplicationContext context) {
+        var builder = context.getBean(Jackson2ObjectMapperBuilder.class);
+        var objectMapper = builder.createXmlMapper(false).build();
+        objectMapper.registerModule(new ByteArrayModule());
```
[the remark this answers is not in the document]
Lodin: Oops

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/jackson/JacksonPropertyModel.java
```diff
@@ -0,0 +1,135 @@
...
+    public Optional<MethodInfoModel> getGetter() {
+        if (getter == null) {
+            getter = origin.hasGetter()
+                    ? Optional.of(MethodInfoModel
+                            .of(origin.getGetter().getAnnotated()))
+                    : Optional.empty();
+        }
+
+        return getter;
+    }
```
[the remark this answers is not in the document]
Lodin: Hmm, but how we could represent anything in Typescript if we don't have getter or field?

packages/java/parser-jvm-core/src/test/java/dev/hilla/parser/models/jackson/JacksonPropertyModelTests.java
```diff
@@ -0,0 +1,214 @@
...
+        assertEquals(ctx.getField(name), model.getField());
+    }
+
+    @DisplayName("It should provide property getter (if there is one)")
+    @ParameterizedTest(name = ModelProvider.testNamePattern)
+    @ArgumentsSource(ModelProvider.class)
+    public void should_ProvidePropertyGetter(JacksonPropertyModel model,
+            String name) {
+        assertEquals(ctx.getGetter(name).isPresent(), model.hasGetter());
+        assertEquals(ctx.getGetter(name), model.getGetter());
```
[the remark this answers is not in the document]
Lodin: Well, I don't really want to test internals of Jackson, so it's just about covering the implementation of the model. Just in case anything is changed in the model code.

packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/BackbonePlugin.java
Lodin: Yes, I'm working on it but it's gonna be the next PR to not bloat this one.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/jackson/JacksonPropertyModel.java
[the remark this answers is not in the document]
cromoteca: I'd still output the property, but readonly, if we can pass the info through OpenAPI:
```typescript
export default interface Params {
  readonly limit: number;
}

const params: Params = {
  limit: 5,
}
```
Wouldn't that work?

packages/java/parser-jvm-core/src/test/java/dev/hilla/parser/models/jackson/JacksonPropertyModelTests.java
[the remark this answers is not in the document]
cromoteca: Can't you just compare the getter name to a string?

packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/PropertyPlugin.java
Lodin: Jackson [has](https://github.com/FasterXML/jackson-databind/blob/2.15/src/main/java/com/fasterxml/jackson/databind/ser/BeanSerializerFactory.java#L606) it in the source code. The whole class is just a adapted code from the Jackson BeanSerializationFactory. I guess all of it is necessary for the same behavior in generator and controller.

packages/java/parser-jvm-core/src/test/java/dev/hilla/parser/models/jackson/JacksonPropertyModelTests.java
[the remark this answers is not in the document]
Lodin: Yep, you're right. Let's do it this way.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/jackson/JacksonPropertyModel.java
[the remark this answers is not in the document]
cromoteca: Well, actually I mixed things here, sorry. It should be `writeonly`, but there's probably nothing like that at interface level.

Still, I'd emit that property, as this is a real use case. I'd just keep it always nullable as a value for that property would never be sent to the client.

```typescript
export default interface Params {
  limit: number | undefined;
}
```

WDYT?

Lodin:
@cromoteca, @platosha, what do you think, what annotations should be returned for the property in the following case:
```java
class Foo {
  @Field
  private final String bar = "bar";

  @Getter
  public String getBar() {
    return bar;
  }

  public void setBar(@Setter String value) {
    this.bar = value;
  } 
}
```

Should it be all of them, or it should have some priority, like:

1. If you have a getter, it will be `@Getter`;
2. If you have a setter, it will be `@Setter`;
3. If there is only a field (e.g., a public field), it will be `@Field`?

cromoteca:
> what do you think, what annotations should be returned for the property in the following case:
> ... 
> Should it be all of them, or it should have some priority, like:
> 
> 1. If you have a getter, it will be `@Getter`;
> 2. If you have a setter, it will be `@Setter`;
> 3. If there is only a field (e.g., a public field), it will be `@Field`?

This is the Jackson [implementation](https://github.com/FasterXML/jackson-databind/blob/4ca3952dfb058c7b523162e6b2d14cfff70b1be5/src/main/java/com/fasterxml/jackson/databind/introspect/POJOPropertyBuilder.java#L677-L688) of `getPrimaryMember` that's used:

```java
    @Override
    public AnnotatedMember getPrimaryMember() {
        if (_forSerialization) {
            return getAccessor();
        }
        AnnotatedMember m = getMutator();
        // for setterless properties, however...
        if (m == null) {
            m = getAccessor();
        }
        return m;
    }
```

That's not good for us since `_forSerialization` is `true`, so that's why the setter is not taken into account. @Lodin, I'd check `getAccessor` and if it's `null`, then check `getMutator` to still have a chance to find the setter. Then I'd trust related annotations. WDYT?

platosha:
Related issues:
- https://github.com/vaadin/hilla/issues/146 about considering getters in general, even without a corresponding field
- https://github.com/vaadin/hilla/issues/140 is about considering annotations on getters specifically

On top of it, I think setters are naturally fit for validation annotations that we use for forms.

Considering the above and also the fact that Hilla does not draw the line between sending and receiving data structures, I would suggest to collect annotations form both getters, setters, and fields combined, giving priority to getters and setters whenever there are conflicts.

cromoteca:
> Considering the above and also the fact that Hilla does not draw the line between sending and receiving data structures, I would suggest to collect annotations form both getters, setters, and fields combined, giving priority to getters and setters whenever there are conflicts.

I agree. I'd just drop priorities and leave the result as indeterminate if the developer messes with annotations (e.g. setter non null and getter nullable). Anyway, in the nonnull plugin we already have annotation priorities in place.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/jackson/JacksonPropertyTypeModel.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public String toString() {
+        return "JacksonPropertyTypeModel[" + origin + "]";
+    }
+
+    @Override
+    protected List<AnnotationInfoModel> prepareAnnotations() {
+        return Stream.of(getField(), getGetter(), getSetter())
+                .filter(Optional::isPresent).map(Optional::get)
```
cromoteca, line 82:
```suggestion
                .flatMap(Optional::stream)
```
if you prefer that syntax

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/jackson/JacksonPropertyTypeModel.java
```diff
@@ -0,0 +1,86 @@
...
+                .flatMap(AnnotatedModel::getAnnotationsStream)
```
cromoteca, line 83:
```suggestion
                .flatMap(Optional::stream)
```
if you prefer that syntax

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/jackson/JacksonPropertyTypeModel.java
cromoteca, line 82:
```suggestion
                .flatMap(Optional::stream)
```
if you prefer that syntax

Lodin: Oh, that's the nice one! 😮

cromoteca (approved):

PR #736, opened 2023-01-31
feat!(parser-jvm): add support for ObjectMapper shared custom factory

### LINKED ISSUE #729: Use the same ObjectMapper in engine parser and in runtime
After #720 gets merged, the multi-module engine will start using an ObjectMapper instance for configuring bean introspection, applying visibility rules and such. Another ObjectMapper instance is used by the runtime endpoint controller for the actual JSON (de-)serialization.

It is important to make sure that both instances have the same configuration to avoid any bugs from having differences. Let us make both instances created in the same place to guarantee it.

Lodin:
Resolves #729


packages/java/endpoint/src/main/java/dev/hilla/EndpointControllerConfiguration.java
```diff
@@ -55,6 +55,119 @@ public EndpointControllerConfiguration(
...
+    /**
+     * Registers a {@link ExplicitNullableTypeChecker} bean instance.
+     *
+     * @return the explicit nullable type checker
+     */
+    @Bean
+    public ExplicitNullableTypeChecker typeChecker() {
+        return new ExplicitNullableTypeChecker();
+    }
+
```
cromoteca, line 170: Those big code blocks moving in the same class make it hard to spot changes

cromoteca (approved):

packages/java/endpoint/src/main/java/dev/hilla/EndpointControllerConfiguration.java
Lodin: Oops, sorry, I tried to maintain the original order. Should I probably put it back?

Issue #733, opened 2023-01-30
Make multi-module engine default


platosha:
The multi-module engine has been an experimental feature during 1.2 and 1.3 feature releases. Now we have some confidence that feature-wise it is ready to replace the old source-based generator.

Make the multi-module engine run by default, without the `hilaEngine` experimental feature flag.

The source based generator could be removed afterwards.

PR #738, opened 2023-02-03
feat: enable multi-module engine without feature flag

platosha:
Fixes: #733


packages/java/maven-plugin/src/main/java/dev/hilla/maven/EngineConfigureMojo.java
```diff
@@ -49,16 +51,17 @@ public final class EndpointCodeGeneratorMojo extends AbstractMojo {
     private boolean failOnMissingGenerator;
 
     @Override
-    public void execute() throws EndpointCodeGeneratorMojoException {
-        PluginConfiguration conf = new PluginConfiguration();
+    public void execute() throws EngineConfigureMojoException {
+        EngineConfiguration conf = new EngineConfiguration();
 
         try {
```
Lodin, line 57: I guess, we could combine two try-catch blocks here as well.

packages/java/engine-runtime/src/main/java/dev/hilla/internal/TaskGenerateHillaImpl.java
```diff
@@ -55,13 +53,13 @@ public void execute() throws ExecutionFailedException {
...
 
-            PluginConfiguration config = null;
+            EngineConfiguration config = null;
 
             if (buildDirectoryName != null) {
                 var buildDir = new File(projectDirectory, buildDirectoryName);
 
                 try {
-                    config = PluginConfiguration.load(buildDir);
+                    config = EngineConfiguration.load(buildDir);
```
Lodin, line 62: Probably, it would also be a good idea to improve my code here to not have nested try-catches 😅

Lodin (commented):
LGTM so far

packages/java/engine-runtime/src/main/java/dev/hilla/internal/ParserClassPathConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        ParserClassPathConfiguration that = (ParserClassPathConfiguration) o;
```
Lodin, line 33:
```suggestion
        var that = (ParserClassPathConfiguration) o;
```

packages/java/engine-runtime/src/main/java/dev/hilla/internal/ParserConfiguration.java
```diff
@@ -42,6 +42,49 @@ public Optional<Plugins> getPlugins() {
...
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        ParserConfiguration that = (ParserConfiguration) o;
```
Lodin, line 53:
```suggestion
        var that = (ParserConfiguration) o;
```

packages/java/engine-runtime/src/test/java/dev/hilla/internal/EngineConfigurationTest.java
```diff
@@ -0,0 +1,100 @@
...
+    @AfterEach
+    public void tearDown() {
+        if (this.configFile.exists()) {
+            this.configFile.delete();
+        }
+        this.temporaryDirectory.toFile().delete();
+    }
+
+    @Test
+    public void should_store() throws IOException {
```
Lodin, line 74: Probably, a bit more details in the test name? 🙂
Also, it could be better to follow that format: `should_DoSomething[_When_SomeCondition]`.

packages/java/engine-runtime/src/test/java/dev/hilla/internal/EngineConfigurationTest.java
```diff
@@ -0,0 +1,100 @@
...
+                .relativize(storedBaseDir);
+        storedConfig.set("baseDir", new TextNode(relativeBaseDir.toString()));
+        var expectedConfig = objectMapper.readTree(TEST_CONFIG);
+
+        assertEquals(expectedConfig.toPrettyString(),
+                storedConfig.toPrettyString());
+    }
+
+    @Test
+    public void should_load() throws IOException, URISyntaxException {
```
Lodin, line 91: Same here

packages/java/parser-jvm-plugin-backbone/src/test/java/dev/hilla/parser/plugins/backbone/customname/CustomNameTest.java
```diff
@@ -0,0 +1,25 @@
...
+import dev.hilla.parser.core.Parser;
+import dev.hilla.parser.plugins.backbone.BackbonePlugin;
+import dev.hilla.parser.plugins.backbone.test.helpers.TestHelper;
+import org.junit.jupiter.api.Test;
+
+public class CustomNameTest {
+    private final TestHelper helper = new TestHelper(getClass());
+
+    @Test
+    public void should_UseAppropriateSchema_When_SimpleTypesAreUsed()
```
Lodin, line 16: I guess you forgot to change the test name

Lodin (changes requested):
I have marked a lot of small improvements along with a couple of some important places. Also there is an empty `AbstractParserConfig` file in the `parser-jvm-core` package sneaked back.

packages/java/engine-runtime/src/main/java/dev/hilla/internal/ParserProcessor.java
```diff
@@ -64,8 +68,18 @@ public ParserProcessor endpointExposedAnnotation(
         return this;
     }
 
-    public ParserProcessor openAPIBase(@Nonnull String openAPIPath) {
-        this.openAPIPath = Objects.requireNonNull(openAPIPath);
+    ParserProcessor endpointPrefix(@Nonnull String endpointPrefix) {
```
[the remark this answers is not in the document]
platosha: Removed for now. This can be specified in the user’s OpenAPI base file, if necessary.

cromoteca (changes requested):
I think it's crucial to extract the common code between `TaskGenerateOpenAPIImpl` and `EngineRunner`. Also, the `openapi.json` file path comes from Flow: I'd prefer to have it under our control.

packages/java/engine-runtime/src/main/java/dev/hilla/internal/ParserProcessor.java
```diff
@@ -29,6 +31,7 @@ public final class ParserProcessor {
     private String endpointAnnotationName = "dev.hilla.Endpoint";
     private String endpointExposedAnnotationName = "dev.hilla.EndpointExposed";
     private String openAPIPath;
+    private ClassLoader classLoader = ClassLoader.getSystemClassLoader();
```
Lodin, line 34: Have just checked the Jackson annotation issue: using the system class loader gets different annotation class that continues to break the Jackson parser. However, if we use the following variant:

```java
var classLoader = getClass().getClassLoader();
```

it works. 

So, I would suggest replacing this line with the following one:
```suggestion
    private ClassLoader classLoader = getClass().getClassLoader();
```

packages/java/engine-runtime/src/main/java/dev/hilla/internal/GeneratorConfiguration.java
```diff
@@ -1,19 +1,35 @@
...
-        return Optional.ofNullable(plugins);
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        GeneratorConfiguration that = (GeneratorConfiguration) o;
```
Lodin, line 22:
```suggestion
        var that = (GeneratorConfiguration) o;
```

packages/java/engine-runtime/src/main/java/dev/hilla/internal/EngineConfiguration.java
```diff
@@ -79,4 +80,25 @@ public String getBuildDir() {
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        EngineConfiguration that = (EngineConfiguration) o;
```
Lodin, line 92:
```suggestion
        var that = (EngineConfiguration) o;
```

packages/java/engine-runtime/src/main/java/dev/hilla/internal/AbstractTaskEndpointGenerator.java
```diff
@@ -0,0 +1,158 @@
...
+        if (engineConfiguration == null) {
+            prepareEngineConfiguration();
+        }
+
+        return engineConfiguration;
+    }
+
+    Logger log() {
+        return LoggerFactory.getLogger(AbstractTaskEndpointGenerator.class);
+    }
```
Lodin, line 79:
```suggestion
    private static final logger = LoggerFactory.getLogger(AbstractTaskEndpointGenerator.class);
```

packages/java/maven-plugin/src/main/java/dev/hilla/maven/EngineGenerateMojo.java
```diff
@@ -0,0 +1,76 @@
...
+    public void execute() throws EngineGenerateMojoException {
+        try {
+            var conf = Objects.requireNonNull(
+                EngineConfiguration.load(buildDirectory));
+            var classPath = conf.getClassPath();
+            var urls = new ArrayList<URL>(classPath.size());
+            for (var classPathItem : classPath) {
+                urls.add(new File(classPathItem).toURI().toURL());
+            }
+            var classLoader = new URLClassLoader(urls.toArray(URL[]::new));
```
Lodin, line 61: What if we provide an existing class loader here? like `getClass().getClassLoader()`? Wouldn't it work better and provide an uniform access for all classes loaded by Maven?

packages/java/parser-jvm-core/src/test/java/dev/hilla/parser/core/ParserConfigTests.java
```diff
@@ -41,12 +43,13 @@ public void setup() throws URISyntaxException {
...
         defaultOpenAPI = new OpenAPI()
-                .info(new Info().title("Vaadin Application").version("1.0.0"))
+                .info(new Info().title("Hilla Application").version("1.0.0"))
                 .servers(List
                         .of(new Server().url("http://localhost:8080/connect")
-                                .description("Vaadin Backend")))
+                                .description("Hilla Backend")))
                 .paths(new Paths());
-        parser = new Parser().classPath(defaultClassPathElements)
+        parser = new Parser().classLoader(ClassLoader.getSystemClassLoader())
```
Lodin, line 51: Not sure, but probably we would also need to change class loader in tests as well.

Lodin (changes requested):
I have some thoughts about the correct class loader due to my attempts to make Jackson parser working correctly with the real project.

packages/java/maven-plugin/src/main/java/dev/hilla/maven/EngineConfigureMojo.java
```diff
@@ -0,0 +1,82 @@
...
+    private MavenProject project;
+
+    @Parameter(defaultValue = "${project.build.directory}", readonly = true)
+    private File buildDirectory;
+
+    // If set to false, the plugin will not fail if the generator is not
+    // available. This allows to run this goal just to save the configuration,
+    // without having to run 'npm install' to avoid an error
+    @Parameter(property = "hilla.failOnMissingGenerator", defaultValue = "true")
+    private boolean failOnMissingGenerator;
```
cromoteca, line 50: I added this option to be able to run `hilla:generate` without failing to just store the configuration, but since you brought back `hilla:configure`, I guess this option is useless, and so probably the `GeneratorUnavailableException`.

packages/java/maven-plugin/src/main/java/dev/hilla/maven/EngineGenerateMojo.java
platosha: To my knowledge, `getClass().getClassLoader()` would not work. According to the [Maven Class Loading article](https://cwiki.apache.org/confluence/display/MAVEN/Maven+3.x+Class+Loading):

> The class path of a plugin class loader is derived from the dependencies of the plugin as defined in the POM of the plugin artifact and optionally by user-defined plugin dependencies listed in the <plugin> declaration of the project that uses the plugin. The plugin class loader does neither contain the dependencies of the current project nor the compiled main/test classes of the project. Plugins that need to load classes from the compile/runtime/test class path of a project need to create a custom URLClassLoader in combination with the mojo annotation @requiresDependencyResolution.

Lodin: Hmm, ok, I see. In my case, using `getClass().getClassLoader()` in the `Parser` worked perfectly but I don't know if it is going to work with your changes.

cromoteca (changes requested):

packages/java/engine-core/src/main/java/dev/hilla/engine/ParserConfiguration.java
```diff
@@ -88,16 +118,19 @@ public int hashCode() {
 
     public static class Plugins implements ConfigList<Plugin> {
         private final Set<Plugin> disable = new HashSet<>();
-        private final boolean disableAllDefaults = false;
+        private boolean disableAllDefaults = false;
         private final Set<Plugin> use = new HashSet<>();
 
+        // Jackson requires a constructor for deserialization
         public Plugins() {
         }
```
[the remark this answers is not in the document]
cromoteca: As discussed, this is not applicable due to Maven

cromoteca (approved):

packages/java/engine-core/src/main/java/dev/hilla/engine/ParserConfiguration.java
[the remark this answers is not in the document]
platosha: Making constructor private does not work, fails deserialization in the Maven plugin tests.

Corrected the comment: actually Maven has even more strict requirement here than Jackson, it requires a public default constructor.

packages/java/engine-core/src/main/java/dev/hilla/engine/GeneratorConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            Plugins plugins = (Plugins) o;
```
Lodin, line 112:
```suggestion
            var plugins = (Plugins) o;
```

packages/java/engine-core/src/main/java/dev/hilla/engine/ParserConfiguration.java
```diff
@@ -0,0 +1,184 @@
...
+
+    void setOpenAPIBasePath(String openAPIBasePath) {
+        this.openAPIBasePath = openAPIBasePath;
+    }
+
+    void setPlugins(Plugins plugins) {
+        this.plugins = plugins;
+    }
+
+    public static class Plugin {
```
Lodin, line 80: Could we re-use `Plugin` and `Plugins` classes for both `ParserConfiguration` and `GeneratorConfiguration`?

Lodin (approved):

PR #739, opened 2023-02-03
fix: make `Optional`s always nullable

### LINKED ISSUE #718: Under NonNullApi, Optionals aren't nullable
### Description of the bug 
When I use `@NonNullApi` on the package containing my endpoint, I still expect `Optional<T>` return types to be nullable.

Consider this method signature:
```java
public Optional<String> sayHello(String name) { ... }
```
The generated return type will be `Promise<string | undefined>`, but if I add the `@NonNullApi` package-level annotation, it becomes `Promise<string>`.

This happens with both generators. I'm using version `2.0.0.alpha8`.

cromoteca:
Fixes #718

### DIFF
--- a/packages/java/parser-jvm-plugin-nonnull/src/main/java/dev/hilla/parser/plugins/nonnull/NonnullPlugin.java
+++ b/packages/java/parser-jvm-plugin-nonnull/src/main/java/dev/hilla/parser/plugins/nonnull/NonnullPlugin.java
@@ -53,8 +53,13 @@ public final class NonnullPlugin extends AbstractPlugin<NonnullPluginConfig> {
     @Override
     public void exit(NodePath<?> nodePath) {
         var node = nodePath.getNode();
-        if (node.getTarget()instanceof Schema schema) {
-            if (((SpecializedModel) node.getSource()).isOptional()) {
+
+        if (node.getTarget() instanceof Schema) {
+            var schema = (Schema<?>) node.getTarget();
+            var nodeSource = node.getSource();
+
+            if ((nodeSource instanceof SpecializedModel)
+                    && ((SpecializedModel) nodeSource).isOptional()) {
                 // Optional is always nullable, regardless of annotations
                 schema.setNullable(true);
             } else {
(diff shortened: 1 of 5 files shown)


packages/java/parser-jvm-plugin-nonnull/src/main/java/dev/hilla/parser/plugins/nonnull/NonnullPlugin.java
```diff
@@ -51,29 +52,35 @@ public void enter(NodePath<?> nodePath) {
...
-        // Apply annotations from package (NonNullApi)
-        var annotations = getPackageAnnotationsStream(nodePath);
-
-        // Apply from current node, if source is annotated
-        if (nodePath.getNode().getSource() instanceof AnnotatedModel) {
-            annotations = Stream.concat(annotations,
-                    ((AnnotatedModel) nodePath.getNode().getSource())
-                            .getAnnotationsStream());
+        var node = nodePath.getNode();
+        if (node.getTarget()instanceof Schema schema) {
```
cromoteca, line 56: There is a formatting rule concerning `instanceof` that is not working correctly or is just wrong.

Lodin, line 56: Hmm, as far as I know, we cannot use that syntax in the production code yet since we need to support Java 11.
```suggestion
        if (node.getTarget() instanceof Schema) {
          var schema = node.getTarget();
```

packages/java/parser-jvm-plugin-nonnull/src/main/java/dev/hilla/parser/plugins/nonnull/NonnullPlugin.java
```diff
@@ -51,29 +52,35 @@ public void enter(NodePath<?> nodePath) {
...
+            if (((SpecializedModel) node.getSource()).isOptional()) {
```
Lodin, line 57: Are we sure we have a `SpecializedModel` here?

Lodin (changes requested):

packages/java/parser-jvm-plugin-nonnull/src/main/java/dev/hilla/parser/plugins/nonnull/NonnullPlugin.java
cromoteca: Are we going to cherry-pick to old releases?

Lodin: Hmm, I'm confused 😅 @platosha, what would you say?

Lodin (approved):

platosha (approved):

PR #755, opened 2023-02-08
fix: Make push compatible with Spring native

Artur-:
Uses manual registration of the atmosphere handler instead of the annotation scanning in Atmosphere. This also allows us to create the handler using Spring so that autowiring works out of the box

PR #756, opened 2023-02-09
fix: Automatically register endpoint types when building for native

Artur-:
Based on #755


packages/java/endpoint/src/main/java/dev/hilla/springnative/HillaHintsRegistrar.java
```diff
@@ -19,8 +30,13 @@
  */
 public class HillaHintsRegistrar implements RuntimeHintsRegistrar {
```
platosha, line 31: Needs some tests.

I guess it would be hard to setup a native image integration test. How about a unit test for now though?

packages/java/endpoint/src/main/java/dev/hilla/springnative/HillaHintsRegistrar.java
```diff
@@ -19,8 +30,13 @@
...
 
+    private static final String GENERATED_OPENAPI_JSON = "target/generated-resources/openapi.json";
```
platosha, line 33: Since we’re in runtime, I suppose you could get this file from resources. Would this work?

platosha (changes requested):

packages/java/endpoint/src/main/java/dev/hilla/springnative/HillaHintsRegistrar.java
Artur-: No, and if you compile a project then `openapi.json` is not available in the jar either

packages/java/endpoint/src/main/java/dev/hilla/springnative/HillaHintsRegistrar.java
marcushellberg: Feel free to copy anything of mine you think is useful. I'm not sure what's needed here.

packages/java/endpoint/src/main/java/dev/hilla/springnative/HillaHintsRegistrar.java
```diff
@@ -32,6 +48,31 @@ public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
...
+        try {
+            String openApiAsText = FileUtils.readFileToString(openApiFile,
+                    StandardCharsets.UTF_8);
+            JsonNode openApi = new ObjectMapper().readTree(openApiAsText);
+            ObjectNode schemas = (ObjectNode) openApi.get("components")
+                    .get("schemas");
+            schemas.fieldNames().forEachRemaining(type -> {
+                hints.reflection().registerType(TypeReference.of(type),
+                        MemberCategory.values());
+            });
```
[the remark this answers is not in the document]
Artur-: Good point. I wonder why the chat example works when it does not register the endpoint class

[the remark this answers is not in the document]
Artur-: But of course, `@Endpoint` is marked as `@Component` so it is a Spring bean. Those are handled automatically by Spring

packages/java/endpoint/src/main/java/dev/hilla/springnative/HillaHintsRegistrar.java
Artur-: Added a unit test for now

platosha (approved):

Artur-:
I tried testing the native app but it fails now with
```
Error: Push support in Hilla is not enabled. Enable it in the debug window or by adding com.vaadin.experimental.hillaPush=true to vaadin-featureflags.properties
```

platosha:
> Push support in Hilla is not enabled

Old snapshot maybe? I can't find message in the repo.

Artur-:
Might be something with my local build

Artur-:
I will test again once there is a release. Seems like a project using snapshot still uses `@hilla/frontend": "2.0.0-beta1"` which does not have the latest fixes

PR #759, opened 2023-02-09
chore: enable github action for 2.0 branch

ZheSun88:


.github/workflows/validation.yml
```diff
@@ -2,7 +2,7 @@ name: Validation
 
 on:
   push:
-    branches: [ main, '1.1', '1.0' ]
+    branches: [ main, '2.0', '1.1', '1.0' ]
```
cromoteca, line 5: Maybe we could set a [pattern](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) in all branches to avoid differences and maintenance.

PR #761, opened 2023-02-16
fix: make Jackson annotations work when using hilla:generate

platosha:


packages/java/maven-plugin/src/main/java/dev/hilla/maven/EngineGenerateMojo.java
```diff
@@ -44,7 +44,8 @@ public void execute() throws EngineGenerateMojoException {
             for (var classPathItem : classPath) {
                 urls.add(new File(classPathItem).toURI().toURL());
             }
```
Lodin, line 46: I guess you could use `stream()` here.

packages/java/maven-plugin/src/main/java/dev/hilla/maven/EngineConfigureMojo.java
```diff
@@ -50,11 +50,8 @@ public final class EngineConfigureMojo extends AbstractMojo {
     public void execute() throws EngineConfigureMojoException {
         try {
             EngineConfiguration conf = new EngineConfiguration();
-            conf.setClassPath(Stream
-                    .of(project.getCompileClasspathElements(),
-                            project.getRuntimeClasspathElements())
-                    .flatMap(Collection::stream)
-                    .collect(Collectors.toCollection(LinkedHashSet::new)));
+            conf.setClassPath(
+                    new LinkedHashSet<>(project.getRuntimeClasspathElements()));
```
Lodin, line 54: Do we ignore compile-time classpath elements?

platosha: Not really, according to [the source](https://github.com/apache/maven/blob/b82479bf111e4d0dc806d9be6d87f1fec4c2ba40/maven-core/src/main/java/org/apache/maven/project/MavenProject.java#L365), Maven does include compile-time artifacts there.

Lodin: That's... weird 🤨

packages/java/maven-plugin/src/main/java/dev/hilla/maven/EngineGenerateMojo.java
platosha: The `toURL()` method throws `MalformedURLException`, and exception handling for Stream items is not nice. Luckily, with good'ol plain code this exception is handled with the surrounding `try / catch`.

Lodin: ah, I see. Meh, that broken Stream...

Lodin (approved):

PR #776, opened 2023-02-22
fix: align to Flow changes

cromoteca:
Fixes code according to https://github.com/vaadin/flow/pull/15975. **This PR will fail until the Flow one is merged and the snapshots updated.**

Also, changes how the `EndpointGeneratorTaskFactory` is declared and removes obsolete Java17 Maven profile in ITs.


platosha (approved):

platosha (approved):

platosha (approved):

packages/java/tests/spring/endpoints-discovery/pom.xml
```diff
@@ -70,6 +70,18 @@
...
+                        <goals>
+                            <goal>prepare-frontend</goal>
+                            <!--
+                             Needs explicit frontend build, because the tests
+                             dont wait for DevServer's startup frontend tasks
+                              -->
+                            <goal>build-frontend</goal>
+                        </goals>
+                    </execution>
+                </executions>
```
cromoteca, line 84: If so, it is probably better to remove this here and restore it at line 149 in `packages/java/tests/pom.xml`. I removed it in this PR, but it is maybe needed everywhere.

platosha: I’ll keep both, as they are enabled for different reasons. In this case its test assertions, whereas everywhere else it is for CI performance.

platosha: Ah, you cleaned it up. Thanks!

platosha (approved):

PR #794, opened 2023-02-27
fix(engine): use Node executable provided from Flow to run the generator

### LINKED ISSUE #778: Unable to find auto installed node
### Description of the bug 
Running the build without a global node installed (tested inside Docker) results in

```
#13 51.27 [INFO] Running `npm install` to resolve and optionally download frontend dependencies. This may take a moment, please stand by...
#13 51.58 [INFO] using '/home/myuser/.vaadin/node/node /home/myuser/.vaadin/node/lib/node_modules/npm/bin/npm-cli.js --no-update-notifier --no-audit --scripts-prepend-node-path=true --ignore-scripts install' for frontend package installation
#13 82.47 [INFO] Frontend dependencies resolved successfully.
#13 83.06 [INFO] Hilla engine configuration not found: configure using build system plugin
#13 85.23 [INFO] JVM Parser started
#13 86.79 [INFO] JVM Parser finished successfully
#13 86.80 /usr/bin/env: 'node': No such file or directory
```

`mvn install -Pproduction` without a global Node installation

### Expected behavior
It builds using the existing logic for finding the node installation

### Actual behavior
It thinks `node` is always in PATH and fails

### Versions:
    - Vaadin / Hilla version: 2 beta 3

Lodin:
Fixes #778.


cromoteca:
I think that some common code about running external processes could be shared in the https://github.com/vaadin/hilla/tree/main/packages/java/engine-runtime/src/main/java/dev/hilla/internal/runner package.

packages/java/engine-runtime/src/main/java/dev/hilla/internal/FlowOptions.java
```diff
@@ -0,0 +1,85 @@
+package dev.hilla.internal;
+
+import java.io.File;
+import java.net.URI;
+
+// FIXME: remove after https://github.com/vaadin/flow/issues/16005 is done
```
platosha, line 6: The issue was fixed, let’s clean this up.

platosha (changes requested):

Lodin:
@cromoteca, if I understand correctly, you mean that `GeneratorShellRunner` could be moved to `dev.hilla.internal.runner`, right? That sounds like a good idea but there is a problem: `engine-runtime` depends on `engine-core` but not vice versa. Probably, we could move some code to `utils` package or do something similar but I believe it goes outside of the scope of this PR.

cromoteca:
Yes, I agree to do that as a follow-up

platosha (approved):

PR #797, opened 2023-03-02
fix(engine): change how model checks for nullable fields

### LINKED ISSUE #790: createEmptyValue creates an object with only undefined values
### Description of the bug 
Given
```
public class SampleBook extends AbstractEntity {

    @Lob
    @Column(length = 1000000)
    private byte @Nonnull [] image;
    @Nonnull
    private String name;
    @Nonnull
    private String author;
    private LocalDate publicationDate;
    @Nonnull
    private Integer pages;
    @Nonnull
    private String isbn;
```
then `SampleBookModel.createEmptyValue()` returns an invalid value: `
{author: undefined, image: undefined, isbn: undefined, name: undefined, pages: undefined, …}`

In Hilla 1, the empty value is
`{image: Array(0), name: '', author: '', publicationDate: undefined, pages: 0, …}` which is valid according to `@Nonnull` rules.

As above

### Expected behavior
The empty value should be valid

### Versions:
    - Vaadin / Hilla version: 2 beta 5

cromoteca:
Parser no longer emits the list of required fields for an entity, but model generator was relying on it. The implementation changes to use `nullable` as in the entity generation.

Fixes #790.

### DIFF
--- a/packages/ts/generator-typescript-plugin-model/src/EntityModelProcessor.ts
+++ b/packages/ts/generator-typescript-plugin-model/src/EntityModelProcessor.ts
@@ -6,6 +6,7 @@ import {
   isComposedSchema,
   isEmptyObject,
   isEnumSchema,
+  isNullableSchema,
   isObjectSchema,
   isReferenceSchema,
   ObjectSchema,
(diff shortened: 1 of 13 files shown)


platosha (dismissed):

packages/ts/generator-typescript-plugin-model/src/EntityModelProcessor.ts
```diff
@@ -139,19 +139,14 @@ export class EntityClassModelProcessor extends EntityModelProcessor {
...
 
-    const requiredSet = new Set(required);
     return Object.entries(properties).map(([name, schema]) => {
       const type = new ModelSchemaTypeProcessor(schema, this[$dependencies]).process();
-      const args = new ModelSchemaExpressionProcessor(
-        schema,
-        this[$dependencies],
-        (_) => !requiredSet.has(name),
-      ).process();
+      const args = new ModelSchemaExpressionProcessor(schema, this[$dependencies]).process();
```
Lodin, line 149: This usage is the only one that sets `checkOptional` parameter for the [`ModelSchemaExpressionProcessor` constructor](https://github.com/vaadin/hilla/blob/main/packages/ts/generator-typescript-plugin-model/src/ModelSchemaProcessor.ts#L226). Since we don't need it anymore, we could drop the constructor parameter as well to make code cleaner.

Lodin (changes requested):

Lodin (approved):

PR #798, opened 2023-03-02
fix(engine): let NonNullPlugin find `NonNullApi` annotation on parent packages

### LINKED ISSUE #788: `@NonNullApi` does not work on sub-packages
According to the [Spring reference](https://docs.spring.io/spring-data/commons/docs/current/reference/html/#repositories.nullability.annotations), `@NonNullApi` should work on the whole package, including sub-packages. This does not happen in Hilla 2.x, where only classes directly in the package are affected.

cromoteca:
Modifies the NonnullPlugin to apply `@NonNullApi` annotation from ancestor packages.

Fixes #788

### DIFF
--- a/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PackageInfoReflectionModel.java
+++ b/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PackageInfoReflectionModel.java
@@ -2,6 +2,7 @@ package dev.hilla.parser.models;
 
 import java.util.List;
 import java.util.Objects;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 class PackageInfoReflectionModel extends PackageInfoModel {
@@ -31,7 +32,7 @@ class PackageInfoReflectionModel extends PackageInfoModel {
         var classLoader = getClass().getClassLoader();
         return getAllAncestorPackageNames(origin.getName())
                 .map(classLoader::getDefinedPackage).filter(Objects::nonNull)
-                .map(PackageInfoModel::of).collect(Collectors::toList);
+                .map(PackageInfoModel::of).collect(Collectors.toList());
     }
 
     /**
--- a/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PackageInfoReflectionModel.java
+++ b/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PackageInfoReflectionModel.java
@@ -1,6 +1,8 @@
 package dev.hilla.parser.models;
 
 import java.util.List;
+import java.util.Objects;
+import java.util.stream.Stream;
 
 class PackageInfoReflectionModel extends PackageInfoModel {
     private final Package origin;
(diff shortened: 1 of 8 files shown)


packages/java/parser-jvm-plugin-nonnull/src/main/java/dev/hilla/parser/plugins/nonnull/NonnullPlugin.java
```diff
@@ -115,9 +115,43 @@ private Optional<PackageInfoModel> findClosestPackage(
...
 
+    /**
+     * Returns a stream of all ancestor packages, starting with the immediate
+     * parent package.
+     *
+     * @param nodePath
+     *            the node path
+     * @return the stream of all ancestor packages
+     */
+    private Stream<PackageInfoModel> findAscendantPackages(
```
Lodin, line 126: I believe this method should be a part of `PackageInfoModel` to encapsulate the interaction with the Java Reflection. I guess, it's ok to return a parent package only, so this method could be called in a loop to reach the most top one.

So the method could have the following signature:
```java
public PackageInfoModel getParent();
```

Lodin (changes requested):

packages/java/parser-jvm-plugin-nonnull/src/main/java/dev/hilla/parser/plugins/nonnull/NonnullPlugin.java
Lodin: Hmm, ok, then let's have all ancestors at once. Something like that:
```java
private List<PackageInfoModel> getAncestorPackages();
```

cromoteca: You mean `public`

Lodin: Oops, sure 😅

Or probably we could use `Stream` here as well 🤔

cromoteca: What about implementing in `PackageInfoSourceModel`?

Lodin: Guess it's ok to have just a stub there. We need to remove them eventually because we definitely won't use them anymore

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/PackageInfoReflectionModel.java
```diff
@@ -23,4 +25,26 @@ public String getName() {
     protected List<AnnotationInfoModel> prepareAnnotations() {
         return processAnnotations(origin.getDeclaredAnnotations());
     }
+
+    @Override
+    public List<PackageInfoModel> getAncestors() {
+        var classLoader = getClass().getClassLoader();
+        return getAllAncestorPackageNames(origin.getName())
+                .map(classLoader::getDefinedPackage).filter(Objects::nonNull)
+                .map(PackageInfoModel::of).toList();
```
Lodin, line 34:
```suggestion
                .map(PackageInfoModel::of).collect(Collectors::toList);
```
Since it is gonna be cherry-picked to 1.x, we should support Java 11. We could use it in tests but not in modules.

Lodin (changes requested):

Lodin (approved):

Lodin (approved):

PR #799, opened 2023-03-02
fix(engine-core): rewrite EngineConfiguration to have all path properties OS-agnostic

Lodin:
Tests for `engine-core`, `engine-runtime` and `maven-plugin` packages do not run on Windows machine. It happens because the `EngineConfiguration` path properties comparison happens OS-dependently.

This PR fixes the issue by using `java.nio.file.Path` objects instead of strings. The configuration file produced also contains absolute paths instead of relative ones. Also, for tests to be OS-agnostic, a specific test resolver is written.


platosha (changes requested):
Let us clean up the resources related with the old generator.

cromoteca (approved):

packages/java/engine-core/src/main/java/dev/hilla/engine/EngineConfiguration.java
```diff
@@ -3,47 +3,67 @@
...
+     * @param configFile
+     *            a path to a configuration file.
+     * @return the configuration, or <code>null</code> if the configuration file
+     *         does not exist
+     * @throws IOException
+     *             if thrown while reading the configuration file
+     * @throws ConfigurationException
+     *             if the configuration file is invalid
+     */
+    public static EngineConfiguration load(File configFile) throws IOException {
```
platosha, line 66: Since we don't intent to make the file name configurable, should this be a private method?

packages/java/engine-core/src/main/java/dev/hilla/engine/EngineConfiguration.java
```diff
@@ -63,104 +83,140 @@ public static EngineConfiguration load(File targetDir) throws IOException {
...
-        this.parser = parser;
+    @Override
+    public int hashCode() {
+        return Objects.hash(baseDir, classPath, generator, parser, buildDir,
+                classesDir, outputDir);
     }
 
-    public String getBuildDir() {
-        return buildDir;
+    public void store(File file) throws IOException {
```
platosha, line 138: How about `storeInDirectory`, similar with `loadDirectory`?

platosha:
Feel free to ignore my questions here. I'm mostly just wondering if we can make it even cleaner...

platosha (approved):

packages/java/engine-core/src/main/java/dev/hilla/engine/EngineConfiguration.java
Lodin: Good catch, thanks!

PR #811, opened 2023-03-08
fix: Make openapi.json available runtime for native apps

Artur-:
Fixes 
```
2023-03-08T11:46:01.758+02:00 ERROR 39368 --- [           main] dev.hilla.EndpointController             : Resource '/dev/hilla/openapi.json' is not available
```
on native app startup


platosha (changes requested):
Could we have a unit test for this, just to make sure we don't break it by accident?

Artur-:
Will an integration test for native compiling a small application do?

platosha:
> Will an integration test for native compiling a small application do?

Yes, even better so.

platosha (approved):

PR #812, opened 2023-03-08
test: Upgrade to TestBench 9.0.0

Artur-:
´


cromoteca:
I see that this change is also coming in https://github.com/vaadin/hilla/pull/799 which should be merged soon.

PR #818, opened 2023-03-08
fix: use classloader from current class to find packages

### LINKED ISSUE #816: @NonNullApi in package-info does not work in 2.0.0
### Description of the bug 
I have an app that has a `package-info.java` file with a `@NonNullApi` annotation. In 2.0.0.beta4, the generated endpoint TS was correct. With 2.0.0 all types have `| undefined` .

1. Create a new Hilla 2.0.0 project. 
2. Add a `package-info.java` file with a `@NonNullApi` annotation
3. Create an endpoint in the package that returns an object. The TS type is nullable.

### Expected behavior
The type is not nullable. 

### Actual behavior
The type is nullable. 

### Versions:
    - Vaadin / Hilla version: 2.0.0
    - Node version: 19.5.0
    - Java version: OpenJDK Runtime Environment GraalVM CE 22.3.0 (build 19.0.1+10-jvmci-22.3-b08)
    - OS version: MacOS 13.1

cromoteca:
Move the API to the ClassInfoModel where we have the good class loader. Finding ancestors is still a "best effort" since packages start existing when a class is loaded from them (see test for an example).

Fixes #816

### DIFF
--- a/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
+++ b/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
@@ -32,7 +32,6 @@ public abstract class ClassInfoModel extends AnnotatedAbstractModel
     private List<ClassInfoModel> innerClasses;
     private List<ClassRefSignatureModel> interfaces;
     private List<MethodInfoModel> methods;
-    private PackageInfoModel pkg;
     private List<PackageInfoModel> ancestors;
     private Optional<ClassRefSignatureModel> superClass;
     private List<TypeParameterModel> typeParameters;
@@ -300,14 +299,6 @@ public abstract class ClassInfoModel extends AnnotatedAbstractModel
         return methods;
     }
 
-    public PackageInfoModel getPackage() {
-        if (pkg == null) {
-            pkg = preparePackage();
-        }
-
-        return pkg;
-    }
-
     /**
      * Tries to find all packages that are ancestors of this class. From the
      * list of all possible ancestor packages, only those whose existence is
@@ -449,8 +440,6 @@ public abstract class ClassInfoModel extends AnnotatedAbstractModel
     protected abstract List<ClassRefSignatureModel> prepareInterfaces();
 
     protected abstract List<MethodInfoModel> prepareMethods();
-
-    protected abstract PackageInfoModel preparePackage();
 
     protected abstract List<PackageInfoModel> prepareAncestors();
 
--- a/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
+++ b/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
@@ -33,6 +33,7 @@ public abstract class ClassInfoModel extends AnnotatedAbstractModel
     private List<ClassRefSignatureModel> interfaces;
     private List<MethodInfoModel> methods;
     private PackageInfoModel pkg;
+    private List<PackageInfoModel> ancestors;
     private Optional<ClassRefSignatureModel> superClass;
     private List<TypeParameterModel> typeParameters;
 
@@ -307,6 +308,22 @@ public abstract class ClassInfoModel extends AnnotatedAbstractModel
         return pkg;
     }
 
+    /**
+     * Tries to find all packages that are ancestors of this class. From the
+     * list of all possible ancestor packages, only those whose existence is
+     * confirmed are returned. This is a "best effort" to find them as, even if
+     * they exist, there is no guarantee that they have already been loaded.
+     *
+     * @return a list of packages that are ancestors of this class
+     */
+    public List<PackageInfoModel> findAllAvailableAncestors() {
+        if (ancestors == null) {
+            ancestors = prepareAncestors();
+        }
+
+        return ancestors;
+    }
+
     public abstract String getSimpleName();
 
     @Override
@@ -318,6 +335,7 @@ public abstract class ClassInfoModel extends AnnotatedAbstractModel
         return typeParameters;
     }
 
+    @Override
     public int hashCode() {
         return 3 + getName().hashCode();
     }
@@ -394,6 +412,7 @@ public abstract class ClassInfoModel extends AnnotatedAbstractModel
         return isAssignableFrom(cls.getClass());
     }
 
+    @Override
     public abstract boolean isEnum();
 
     public abstract boolean isFinal();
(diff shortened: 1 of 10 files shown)


packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
```diff
@@ -33,6 +33,7 @@ public abstract class ClassInfoModel extends AnnotatedAbstractModel
     private List<ClassRefSignatureModel> interfaces;
     private List<MethodInfoModel> methods;
     private PackageInfoModel pkg;
+    private List<PackageInfoModel> ancestors;
```
Lodin, line 36: I guess, if we store a list of ancestors, we no longer need a separate `pkg` field. We could just return a first list item. Or we could even remove the `getPackage` method.

Lodin (changes requested):

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
cromoteca: It depends. In a general API I'd keep both, as the new one is more expensive to compute, so keeping the old one makes sense. But here, if we know that we're going to always use both anyway, your suggestion could be valid. WDYT?

Lodin: I believe that the cost for computation is relatively negligible. Also, we are not using the single ancestor, only all of them in `@NonNullApi`, and this is the only place where we need that computation for now. So I would say that using `List` is preferable here.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
```diff
@@ -299,12 +299,20 @@ public List<MethodInfoModel> getMethods() {
...
-            pkg = preparePackage();
+    /**
+     * Tries to find all packages that are ancestors of this class. From the
+     * list of all possible ancestor packages, only those whose existence is
+     * confirmed are returned. This is a "best effort" to find them as, even if
+     * they exist, there is no guarantee that they have already been loaded.
+     *
+     * @return a list of packages that are ancestors of this class
+     */
+    public List<PackageInfoModel> findAllAvailableAncestors() {
```
Lodin, line 310: Oh, almost missed. I guess, there should be not `ancestors` but `packages` because it is `ClassInfoModel`, not `PackageInfoModel`. Also, I guess, it would be better to have it as a getter. Not sure, but for me `get` is more clear than `find`.

Like:
```suggestion
    public List<PackageInfoModel> getAllAvailablePackages() {
```

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoReflectionModel.java
```diff
@@ -220,8 +222,25 @@ protected List<MethodInfoModel> prepareMethods() {
...
+
+    /**
+     * Returns a stream of all ancestor package names, starting with the package
+     * itself.
+     *
+     * @param name
+     *            the package name
+     * @return the stream of all ancestor package names
+     */
+    private static Stream<String> getAllAncestorPackageNames(String name) {
```
Lodin, line 241: Maybe, drop the `Ancestor` part here?
```suggestion
    private static Stream<String> getAllPackageNames(String name) {
```

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
```diff
@@ -431,7 +441,7 @@ public String toString() {
 
     protected abstract List<MethodInfoModel> prepareMethods();
 
-    protected abstract PackageInfoModel preparePackage();
+    protected abstract List<PackageInfoModel> prepareAncestors();
```
Lodin, line 444: Same here. Also, I usually try to maintain same naming for `get` and `prepare` methods.
```suggestion
    protected abstract List<PackageInfoModel> prepareAllAvailablePackages();
```

Lodin (changes requested):
Sorry, I missed that part in my previous review.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
cromoteca: There's no obligation to use the `get` prefix to access something. This is not a property. The reason to use "find" was to make more explicit that here we're searching for something that could be even not found. I'd keep the "find" prefix here.

Concerning the "ancestors", there's no notion of packages (plural) for a class, so while `getPackage` makes sense to me, `getPackages` doesn't.

I'd rather simplify to `findAncestors` (no need to mention `packages` as it can be inferred by the returned type).

Lodin: Wouldn't users be confused if it will return a package but not the parent class (in case of a nested class)?

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoReflectionModel.java
cromoteca: As above, the purpose is to explain how multiple packages are related to a class name. I'd rather change the `name` parameter to `packageName`.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
cromoteca: I don't think so, given that the return type is `List<PackageInfoModel>`.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
cromoteca: Ok I'll sync them once we choose the right one for the public method.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoReflectionModel.java
cromoteca: And the "all" is because here we get them all, as names, even the ones that will be filtered out after.

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoModel.java
Lodin: Hmm, ok, sounds reasonable. I agree with having it shorter then:
```suggestion
    public List<PackageInfoModel> findAncestors() {
```

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoReflectionModel.java
Lodin: yeah, since we've agreed on the general naming, this one is no more necessary

Lodin (approved):

PR #835, opened 2023-03-13
feat(form): consider errors from HTML and Vaadin components built-in validation

### LINKED ISSUE #702: Date picker allow invalid input // Form is sent with 'null' value for date fields

platosha:
Fixes #702

In some form element binding cases (notably, for user-typed value in
`<vaadin-date-picker>`), the binder validation additionally verify
the value from the element using the element's own internal validation
defined by the `checkValidity()` result and the `validity` property.

This change implements such verification, so that a value invalidated
by the element's internal validation is considered invalid by the
binder.

### DIFF
--- a/packages/ts/form/src/BinderNode.ts
+++ b/packages/ts/form/src/BinderNode.ts
@@ -31,6 +31,7 @@ import {
   ObjectModel,
 } from './Models.js';
 import type { Validator, ValueError } from './Validation.js';
+import { _validity } from './Validity.js';
 
 const _ownErrors = Symbol('ownErrorsSymbol');
 const _visited = Symbol('visited');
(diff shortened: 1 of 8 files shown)


packages/ts/form/src/BinderNode.ts
```diff
@@ -39,6 +40,30 @@ function getErrorPropertyName(valueError: ValueError<any>): string {
...
+  private binderNode: BinderNode<T, M>;
+
+  // eslint-disable-next-line no-useless-constructor
+  constructor(binderNode: BinderNode<T, M>) {
+    this.binderNode = binderNode;
+  }
+
+  validate(): boolean {
+    if (!this.binderNode[_validity]) {
+      return true;
```
Lodin, line 60: Codecov says that this line is not covered by tests. Are you going to add a test for this case?

platosha: I ended up removing it. There's no point in this double checking of the state.

Lodin (approved):

PR #857, opened 2023-03-23
chore: revert lerna update

platosha:
Need to revert because of issues in the release build.


platosha:
Closed in favor of https://github.com/vaadin/hilla/pull/858

PR #862, opened 2023-03-28
fix(parser): consider annotations from both fields and accessors

### LINKED ISSUE #806: List<@Nonnull String> should always map to Array<string>
### Description of the bug 
Given
```
public class HelloWorldEndpoint {

    public static class Obj {
        @Nonnull List<@Nonnull String> hellos;

        public List<String> getHellos() {
            return hellos;
        }

        public void setHellos(List<String> hellos) {
            this.hellos = hellos;
        }
    }

    public @Nonnull Obj sayHello(@Nonnull List<@Nonnull String> something) {
        Obj o = new Obj();
        o.setHellos(List.of("Hello stranger"));
        return o;
    }
}
```

You get
```ts
async function sayHello_1(something: Array<string>, init?: EndpointRequestInit_1): Promise<Obj_1> { return client_1.call("HelloWorldEndpoint", "sayHello", { something }, init); }
```

as expected but you also get

```
interface Obj {
    hellos: Array<string | undefined>;
}
```
which is incorrect.

### Expected behavior
```ts
interface Obj {
    hellos: Array<string>;
}
```

### Versions:
    - Vaadin / Hilla version: 2 rc2

### LINKED ISSUE #830: Client side validators not generated in model for hilla 2.0
### Description of the bug 
Client side validators not generated in model for hilla 2.0

* Check out the basics tutorial project
  `git clone https://github.com/vaadin/hilla-basics-tutorial.git`
* Reset the branch to hilla 1.2.7
  `git reset --hard 1804c05a91fd7e37131df59edeec64e395d45844`
* Set the execute permission for the mvnw script
  `chmod +x mvnw`
* Run the project
  `./mvnw`

The `TodoModel` contains the expected validator

* Reset the branch to hilla 2.0
  `git reset --hard HEAD` followed by `git pull`
* Run the project
  `./mvnw`

The `TodoModel` does not contain the expected validator

### Expected behavior
The model should have a NotBlank validator for the task property.

```
export default class TodoModel<T extends Todo = Todo> extends ObjectModel<T> { 
  static createEmptyValue: () => Todo;

  get id(): NumberModel {
    return this[_getPropertyModel]('id', NumberModel, [true]);
  }

  get done(): BooleanModel {
    return this[_getPropertyModel]('done', BooleanModel, [false]);
  }

  get task(): StringModel {
    return this[_getPropertyModel]('task', StringModel, [true, new NotBlank()]);
  }
}
```

### Actual behavior
The model does not have a NotBlank validator for the task property.

```
class TodoModel<T extends Todo_1 = Todo_1> extends ObjectModel_1<T> {
    declare static createEmptyValue: () => Todo_1;
    get done(): BooleanModel_1 {
        return this[_getPropertyModel_1]("done", BooleanModel_1, [false]) as BooleanModel_1;
    }
    get id(): NumberModel_1 {
        return this[_getPropertyModel_1]("id", NumberModel_1, [true]) as NumberModel_1;
    }
    get task(): StringModel_1 {
        return this[_getPropertyModel_1]("task", StringModel_1, [true]) as StringModel_1;
    }
}
```

### Versions:
    - Vaadin / Hilla version: 2.0
    - Node version: v18.14.2
    - Java version: 17.0.6
    - OS version: Mac OS Ventura 13.2.1

platosha:
Fixes #830
Fixes #806


cromoteca (changes requested):
```
[ERROR] Failures: 
[ERROR]   JacksonPropertyModelTests.should_HaveCorrectType:49 expected: <true> but was: <false>
[ERROR]   JacksonPropertyModelTests.should_HaveCorrectType:49 expected: <true> but was: <false>
[ERROR]   JacksonPropertyModelTests.should_HaveCorrectType:49 expected: <true> but was: <false>
[ERROR]   JacksonPropertyModelTests.should_HaveCorrectType:49 expected: <true> but was: <false>
[ERROR]   JacksonPropertyModelTests.should_HaveCorrectType:49 expected: <true> but was: <false>
[ERROR]   JacksonPropertyModelTests.should_HaveCorrectType:49 expected: <true> but was: <false>
[ERROR]   JacksonPropertyModelTests.should_HaveCorrectType:49 expected: <true> but was: <false>
```
As test errors are deeply related to new features, I think that they should pass before reviewing.

cromoteca (approved):

cromoteca:
I've spotted Java 17 somewhere in the PR, but I guess that's fine as target is 2.0, right?

packages/java/parser-jvm-core/src/test/java/dev/hilla/parser/models/jackson/JacksonPropertyShared.java
Lodin, line 1: I guess, it should have name `JacksonPropertySharedTests.java`. I spent some time trying to understand why the `Map` that is definitely for tests is placed in the code 😅

packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/nodes/TypedNode.java
```diff
@@ -0,0 +1,20 @@
+package dev.hilla.parser.plugins.backbone.nodes;
+
+import java.util.List;
+import java.util.function.UnaryOperator;
+
+import dev.hilla.parser.models.SignatureModel;
+import io.swagger.v3.oas.models.media.Schema;
+import jakarta.annotation.Nonnull;
+
+public interface TypedNode extends AnnotatedNode {
```
Lodin, line 10: I wonder if we could have better naming for these nodes. Now it sounds strange because `TypedNode` is not actually typed, it contains a type. And `AnnotatedNode` also contains a model with annotations, not annotated itself. What do you think?

Lodin (changes requested):

packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/nodes/TypedNode.java
platosha: I took `AnnotatedModel` as an example for the `AnnotatedNode` interface. The idea is similar: this is something that provides access to annotations and type information.

In practice the node implementation delegates such access to its contained source, but the exact way of delegating is an implementation detail of a particular node class.

Any suggestions for a better name on your mind?

Lodin: After some consideration, I think it is ok to have this naming. If necessary, we could refactor it any time later

Lodin (approved):

PR #870, opened 2023-03-31
chore(tests): validation with two IT test groups in parallel

platosha:


cromoteca:
If you don't mind using an external script, you could split the packages in two groups automatically using a script like this one:
```bash
#!/bin/bash

# Get the list of test modules and keep only those that have packaging type jar or war
module_list=$(find packages/java/tests -name pom.xml | xargs -n 1 -I {} bash -c 'packaging=$(mvn help:evaluate -Dexpression=project.packaging -q -DforceStdout -f {}); if [[ "$packaging" =~ ^[jw]ar$ ]]; then echo {}; fi' 2>/dev/null)

# Extract even elements
module_list_even=$(echo $module_list | awk '{for(i=1;i<=NF;i+=2)print $i}')

# Extract odd elements
module_list_odd=$(echo $module_list | awk '{for(i=2;i<=NF;i+=2)print $i}')

# Create comma separated list of modules
module_list_even=$(echo $module_list_even | sed -e 's/ /,/g')
module_list_odd=$(echo $module_list_odd | sed -e 's/ /,/g')

#echo "Even modules: $module_list_even"
#echo "Odd modules: $module_list_odd"
#exit 0

# execute mvn verify on even modules
mvn verify -pl $module_list_odd
mvn verify -pl $module_list_even
```
It's a bit slow as it calls Maven for each module, but will adapt in case of new modules.

cromoteca (approved):

PR #885, opened 2023-04-06
✖️ Increase timeout

cromoteca:
This PR will not be merged and only exists for #884.


cromoteca:
Found a better approach

PR #886, opened 2023-04-06
fix: remove `EndpointRequestInit` import if not used

### LINKED ISSUE #795: Unused import in generated endpoints can lead to TypeScript compilation error
When an endpoint contains only reactive methods, the generated TypeScript file contains an unused import of `EndpointRequestInit_1`. This can lead to a compilation error if unused imports are not allowed in the configuration.

For example:

```typescript
import { EndpointRequestInit as EndpointRequestInit_1, Subscription as Subscription_1 } from "@hilla/frontend";
import client_1 from "./connect-client.default";
function getClock_1(): Subscription_1<string> { return client_1.subscribe("ReactiveEndpoint", "getClock", {}); }
function getClockCancellable_1(): Subscription_1<string> { return client_1.subscribe("ReactiveEndpoint", "getClockCancellable", {}); }
export { getClock_1 as getClock, getClockCancellable_1 as getClockCancellable };
```

```
[ERROR] frontend/generated/ReactiveEndpoint.ts(1,33): error TS6133: 'EndpointRequestInit_1' is declared but its value is never read.
```

cromoteca:
Fixes #795.

This PR is unfinished.

### DIFF
--- a/packages/ts/generator-typescript-plugin-push/src/PushProcessor.ts
+++ b/packages/ts/generator-typescript-plugin-push/src/PushProcessor.ts
@@ -5,14 +5,19 @@ import ts from 'typescript';
 
 const initParameterTypeName = 'EndpointRequestInit';
 
-export default class PushProcessor {
+export type EndpointOperations = {
+  methodsToPatch: string[];
+  removeInitImport: boolean;
+};
+
+export class PushProcessor {
   readonly #dependencies = new DependencyManager(new PathManager());
-  readonly #methods: readonly string[];
+  readonly #operations: EndpointOperations;
   readonly #source: ts.SourceFile;
   readonly #subscriptionId: ts.Identifier;
 
-  constructor(source: ts.SourceFile, methods: readonly string[]) {
-    this.#methods = methods;
+  constructor(source: ts.SourceFile, operations: EndpointOperations) {
+    this.#operations = operations;
     this.#source = source;
 
     const { imports, paths } = this.#dependencies;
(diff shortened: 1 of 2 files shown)


platosha (approved):

Lodin:
I'm thinking about a bit different approach. This PR aims removing `EndpointRequestInit` if it is not used but if we take a closer look, it is a problem of removing unused imports at all. So, it is more likely a scope of responsibility for `DependencyManager` from `generator-typescript-utils`. 

So, here is my idea. What if we create a kind of a reference counter for the `Identifier` returned by an `ImportManager#add`? So, if it is used anywhere, we increase counter (or simply put a flag `used` in there). And if this flag is set, we allow that import to go to the generated code; otherwise, it should be skipped. 

I believe, that approach is more robust and would allow us to remove any unused imports. 

What do you think, @cromoteca, @platosha?

cromoteca:
The problem here is that the import is created and then could be removed, so that counter would probably be incremented at first anyway. Then we should decrease it, and then remove the import.

Even better, what if that import is only created when needed and we move the `push`-related generation to first-class instead of patching existing methods?

Lodin:
Hmm, that sounds nice! 🤔
Yeah, we could traverse across all the `Identifier`s connected with using non-variable and non-declared stuff, find missing stuff and do an import. The only problem here is that we have no idea from where we have to import it. 

So, probably, there could be a solution in the middle. We use the current system of adding dependencies but when the TS file is gonna generate, we could traverse the AST and search for `Identifier`s declared in the import. If the import is not used anywhere, we could easily remove it.

Lodin (approved):
After the internal discussion, we decided to accept this PR as a hotfix, and later we could consider more robust solution.

PR #911, opened 2023-04-21
fix(form): error message not displayed in invalid fields

### LINKED ISSUE #887: Error message is not displayed below field since 1.3.8
### Description of the bug 
The error message is not displayed below the field in hilla version 1.3.8 and later.

My best guess is that it broke with this commit:  https://github.com/vaadin/hilla/commit/10ebc6afc170f9629bf0abb3f3983d5acb98b2f3

* Check out the basics tutorial project
  git clone https://github.com/vaadin/hilla-basics-tutorial.git
* Reset the branch to hilla 1.2.7
  git reset --hard 1804c05a91fd7e37131df59edeec64e395d45844
* Set hilla version to 1.3.8 in `pom.xml`
  `<hilla.version>1.3.8</hilla.version>`
* Set spring boot parent to 2.7.10 in `pom.xml`
  `<version>2.7.10</version>`
* Set the execute permission for the mvnw script
  chmod +x mvnw
* Run the project
  ./mvnw

### Expected behavior
The error message should be shown below the field when clicking in the field and tabbing out of it. (The screen shot is from hilla 1.3.7)

Screenshot 2023-04-06 at 20 47 34

### Actual behavior
The error message is not shown below the field when clicking in the field and tabbing out of it.

Screenshot 2023-04-06 at 20 45 05

### Versions:
- Vaadin / Hilla version: 1.3.8
- Node version: v18.14.2
- Java version: 17.0.6
- OS version: Mac OS Ventura 13.3

platosha:
Fixes #887

### DIFF
--- a/packages/ts/form/src/Field.ts
+++ b/packages/ts/form/src/Field.ts
@@ -102,23 +102,60 @@ export abstract class AbstractFieldStrategy<T = any> implements FieldStrategy<T>
     this._validityFallback = {
       ...defaultValidity,
       valid,
-      customError: !valid,
+      ...(valid ? {} : this._detectValidityError()),
     };
     return valid;
   }
+
+  private _detectValidityError(): Readonly<Partial<ValidityState>> {
+    if (!('inputElement' in this.element)) {
+      // Not a Vaadin component field
+      return { customError: true };
+    }
+
+    const inputElement = this.element.inputElement as FieldElement<string>;
+
+    if (this.element.value === '') {
+      if (inputElement.value === '') {
+        return { valueMissing: true };
+      }
+      // Some value is entered, but not meaningful to the
+      // web component — assume parse error.
+      return { badInput: true };
+    }
+    // Unknown constraint violation
+    return { customError: true };
+  }
 }
 
-export class VaadinFieldStrategy extends AbstractFieldStrategy {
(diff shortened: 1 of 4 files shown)


platosha:
See also: #912

packages/ts/form/src/Field.ts
```diff
@@ -255,7 +292,9 @@ export const field = directive(
         this.fieldState = fieldState;
 
         const updateValueFromElement = () => {
-          if (fieldState.strategy.checkValidity()) {
+          fieldState.strategy.checkValidity();
+          // When bad input is detected, skip reading new value in binder state
+          if (!fieldState.strategy.validity.badInput) {
```
cromoteca, line 297: Just to understand what happens here: if in a date picker I change the initial value of `21/04/2023` to `21/04/2023a`, that value is not stored in the binder and the original one is safe, but if I change it to `21/04/202`, it is stored, right?

platosha: Yes, assuming that `21/04/202` is a valid format for the `<vaadin-date-picker>`.

platosha: Another easy way to break `<vaadin-date-picker>` parser that is even easier to reproduce is to enter too many digits, like `21/04/20231111`.

cromoteca (approved):

PR #947, opened 2023-05-12
Add Maven Wrapper

cromoteca:
Let's stabilize the Maven version used in Hilla and also ease building the project on external systems just in case.


platosha:
Let's leave it out for now and get back to this once there is a motivation for adding the wrapper.

PR #952, opened 2023-05-15
feat: scaffold application coming from Spring Initializr

cromoteca:
Adds Maven goals and Gradle tasks to bootstrap a newly created application.

Related to https://github.com/vaadin/flow/pull/16778.

Closes https://github.com/vaadin/hilla/issues/954.


packages/java/hilla-plugin-base/src/main/java/dev/hilla/plugin/base/InitFileExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+        } else if (content.contains("hilla-spring-boot-starter")) {
+            return Frameworks.LIT;
+        } else {
+            throw new IllegalArgumentException("No hilla starter found");
+        }
+    }
+
+    private String findSpringBootApplicationPackage() throws IOException {
+        return Files.walk(projectDirectory)
+                .filter(path -> path.toString().endsWith(".java"))
```
taefi, line 136: Could we change the logic here so that instead of looking into all directories, we just look into `[project-root]/src` and then remove the condition for paths ending in ".java" for the future, for instance, to support Kotlin? Although, currently we don't fully support Kotlin out of the box. Just a suggestion.

pom.xml
```diff
@@ -21,6 +21,7 @@
     <module>packages/java/engine-runtime</module>
     <module>packages/java/maven-plugin</module>
     <module>packages/java/gradle-plugin</module>
+    <module>packages/java/hilla-plugin-base</module>
```
platosha, line 24:
```suggestion
    <module>packages/java/plugin-base</module>
```

I would simply name it `plugin-base`, this would look more like other technical artifact names (`generator-...`, `parser-...`) and less like something for the actual users.

platosha: On the other hand, maybe I'm wrong, sticking to a common pattern with, e. g., `flow-plugin-base`, is good too.

cromoteca: Well, in the Hilla project we decided not to use `hilla` in module names, if I recall correctly.

taefi: Oh, my bad! ;)

platosha (approved):
Approving this given the time constraints. Let us make another PR for any improvements.

platosha (approved):
Approving this given the time constraints. Let us make another PR for any improvements.

platosha (approved):
Approving this given the time constraints. Let us make another PR for any improvements.

PR #967, opened 2023-05-19
test(TypeScript): prevent hangups of TypeScrpt tests on CI

platosha:


platosha:
Here's the summary of my findings on this issue: https://github.com/bcoe/c8/issues/454#issuecomment-1554997984

cromoteca (approved):

PR #973, opened 2023-05-23
feat: use Spring Validator when available

### LINKED ISSUE #964: Autowiring in Validators in Hilla Endpoint DTO Validation
### Description of the bug 
When annotating a DTO which is submitted to an Endpoint with a Custom Constraint Validator that DTO is validated by dev.hilla.EndpointInvoker#validator. That works as expected. If that Validator has an @Autowire dependency it is not resolved and is therefore null.

Call any endpoint with any DTO. That DTO is annotated by a custom constraint with a custom validator with an @Autowire dependency

### Expected behavior
The dependency is resolved and therefore not null

### Actual behavior
That dependency is not resolved according to Spring Spec

### Versions:
    - Vaadin / Hilla version: 1.2.10
    - Node version: 18.15.0
    - Java version: 1.8.0_352
    - Application Server (if applicable): Spring 2.7.0

### Working with spring
If you call a Spring service with org.springframework.validation.annotation.Validated and the same DTO as parameter with javax.validation.Valid [0] spring resolves the dependency as expected.

### Hibernate workaround
Hibernate has the same problem which we fixed by implementing a custom org.springframework.context.annotation.Configuration which returns a factory org.springframework.context.annotation.Bean that implements org.springframework.beans.factory.config.AutowireCapableBeanFactory

### Private Repo
As always we have shared a reproducer example in our private repo

[0]
@Validated
public class Service {
    public void serviceCall(@Valid final IEntity2 entity2) {

cromoteca:
The `EndpointInvoker` gets the Spring `ApplicationContext` as constructor parameter, so we can use the Spring Validator Bean instead of the explicit one.

Closes #964.

### DIFF
--- a/packages/java/endpoint/src/main/java/dev/hilla/EndpointInvoker.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/EndpointInvoker.java
@@ -68,8 +68,7 @@ public class EndpointInvoker {
     private final EndpointRegistry endpointRegistry;
     private final ExplicitNullableTypeChecker explicitNullableTypeChecker;
     private final ServletContext servletContext;
-    private final Validator validator = Validation
-            .buildDefaultValidatorFactory().getValidator();
+    private final Validator validator;
 
     /**
      * Creates an instance of this bean.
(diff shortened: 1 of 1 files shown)


platosha (approved):

cromoteca:
I see in the issue that you're using Hilla 1.2. That version is not maintained anymore. We can consider backporting down to 1.3 if that can help.

gitgdako:
Hi. We are currently evaluating the possibility to upgrade to Hilla 2, but that might take some time. In the meanwhile we intend to use the latest stable 1.x, so making that feature available there would be of great use to us

gitgdako:
Hi. In what 1.x version can we expect this to be included? We will be staying on 1.x (latest with regular upgrades) for a while

platosha:
@gitgdako thanks for asking. Now that this change was [backported to the 1.3 branch](https://github.com/vaadin/hilla/pull/1034), it will be included in the next maintenance release, v1.3.15.

gitgdako:
Code is working as expected in 1.3.15
Imo this issue is resolved

PR #989, opened 2023-05-26
chore: replace Lerna with Nx

Lodin:


platosha (changes requested):
Please mind to keep the release scripts consistent between Hilla versions. For example, consider adding `npm run publish` as a common entrypoint for both Nx and Lerna.

Lodin:
Closing in favor of #1365.

PR #998, opened 2023-06-01
feat: always scan the root package of current module/project for endpoints

### LINKED ISSUE #994: Consider always scaning current project/module root package for endpoints
Currently, the packages from current project (or current module of a multi-module project) is scanned for endpoints, only if there is no configuration for the exposed packages from dependencies (or other modules). If the user configure the maven/gradle plugin to scan and include a single external endpoint from some dependency (e.g. SSO Kit), then they should also always list the a high-level package from the current project/module which is not convenient and redundant IMO.

Configuring the for exposed packages from other modules or from dependencies should not affect the parser behavior for scanning the packages in the current project (or current module of a multi-module project), as most of the times, the use case is to use some external utility endpoints, and not to hide the endpoints of the current project/module. 

We should consider always scanning the current project/module's packages, plus the additional exposed packages from the dependencies or other modules.

taefi:
## Description

This is a fixed-term solution for always including the root package of current 
module/project while scanning for endpoints. 

Finding the root package of the current module can be done in different ways,
and in this PR it is relying on the existence of a class annotated with 
@SpringBootApplication.

The better approach for scanning the endpoints could rely on spring beans instead, 
however, at the moment making such a breaking change is not possible, so until
then, this can be considered as a partial solution.

Fixes #994 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/Parser.java
```diff
@@ -1,18 +1,12 @@
...
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
```
Artur-, line 4: Expand imports, maybe we could have a check for this...

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/Parser.java
```diff
@@ -285,24 +286,29 @@ public OpenAPI execute() {
...
         var storage = new SharedStorage(config);
 
-        var classGraph = new ClassGraph().enableAnnotationInfo()
-                .ignoreClassVisibility()
-                .overrideClassLoaders(config.getClassLoader());
+        var buildDirectories = config.getClassPathElements().stream()
+                .filter(e -> !e.endsWith(".jar")).toList();
 
-        Collection<String> packages = config.exposedPackages;
+        Collection<String> packages = getConfig().getExposedPackages();
```
Artur-, line 292:
```suggestion
        Collection<String> packages = config.getExposedPackages();
```

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/Parser.java
```diff
@@ -482,4 +488,27 @@ public Collection<Plugin> getPlugins() {
...
+        public Optional<String> findRootPackage(List<String> buildDirectories) {
+
+            var currentModuleClassGraph = new ClassGraph()
+                    .enableAnnotationInfo().ignoreClassVisibility()
+                    .overrideClassLoaders(ClassLoader.getSystemClassLoader())
+                    .overrideClasspath(buildDirectories);
+
+            String springBootApplicationAnnotation = "org.springframework.boot.autoconfigure.SpringBootApplication";
+
+            try (var scanResult = currentModuleClassGraph.scan()) {
```
Artur-, line 503: Does scanning everything again have some performance impact?

Artur- (changes requested):

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/Parser.java
taefi: This is only looking into the `classes` directory and is not scanning the dependencies.

Issue #1012, opened 2023-06-08
Running a Hilla app in IntelliJ doesn't pick up the correct JDK


OlliTietavainenVaadin:
### Description of the bug 
If I try to start a Hilla app from IntelliJ Idea, it fails to start if I have an older JDK defined in my JAVA_HOME system property, even if the IntelliJ project is configured in Project Structure  / Platform Settings -> SDKs to use a modern JDK such as 17 or 20. 

### Minimal reproducible example
Create a new app with the Hilla CLI
### Expected behavior
The Hilla app should use the JDK from the project configuration and start normally. 
### Actual behavior

When starting the server, there are logs such as:
```
[...]

PR #1044, opened 2023-06-15
fix: use Java path of currently running java process to execute configure goal/task

taefi:
## Description

This helps users with older java versions set
in their JAVA_HOME globally, but having correct
java version set in their project build files
and/or in their IDEs configurations.

Now the same Java executable path that is used
to run the project (e.g. from IDE) will be used
to run the hilla:configure (Maven) or hillaConfiure (Gradle).

Fixes: #1012

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


packages/java/engine-core/src/main/java/dev/hilla/engine/commandrunner/CommandRunner.java
```diff
@@ -151,4 +154,39 @@ private void executeCommand(String executable, String[] arguments,
...
+     * has a value, but returning the Java path from project's configurations
+     * relies on IDE's functionality for setting the "java.home" system property
+     * based on the settings of the project.
+     *
+     * @return A String "path/to/java/home" of current running application or
+     *         the path returned by {@code System.getProperty("java.home")}
+     */
+    private String getCurrentJavaProcessJavaHome() {
+        return ProcessHandle.current().info().command().map(javaExecPath -> {
+            String pathToExclude = IS_WINDOWS ? "\\bin\\java" : "/bin/java";
```
cromoteca, line 171:
```suggestion
            String pathToExclude = IS_WINDOWS ? "\\bin\\java.exe" : "/bin/java";
```

platosha (approved):

PR #1055, opened 2023-06-20
feat: Include information on the frontend framework used by Hilla in MAU tracking data

mvysny:
## Description

Today, we can only calculate Lit usage by excluding those using @hilla/react-components from all Hilla users (using @hilla/frontend). If we add more frontend frameworks in the future, this process becomes more difficult and error-prone.

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


packages/java/endpoint/src/main/java/dev/hilla/HillaStats.java
```diff
@@ -0,0 +1,70 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
```
taefi, line 2:
```suggestion
 * Copyright 2000-2023 Vaadin Ltd.
```

packages/java/endpoint/src/test/java/dev/hilla/HillaStatsTest.java
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
```
taefi, line 2:
```suggestion
 * Copyright 2000-2023 Vaadin Ltd.
```

packages/java/endpoint/src/main/java/dev/hilla/EndpointController.java
```diff
@@ -158,6 +158,9 @@ public void registerEndpoints() {
             endpointBeans.forEach((name, endpointBean) -> endpointRegistry
                     .registerEndpoint(endpointBean));
         }
+
+        // Usage statistics
+        HillaStats.report();
```
taefi, line 163: The only question I have: 
Does it make sense to wrap this method call in a `try { } catch(Throwable t) { log(..., t); }` to make sure a runtime error in `UsageStatistics.markAsUsed` is not affecting the application runtime behavior? Or do we intentionally want it to fail in case it wasn't able to gather the statistics?

packages/java/endpoint/src/test/java/dev/hilla/HillaStatsTest.java
```diff
@@ -0,0 +1,97 @@
...
+    private ClassLoader oldContextClassLoader;
+
+    @Rule
+    public TemporaryFolder temporary = new TemporaryFolder();
+
+    @Before
+    public void rememberContextClassLoader() throws Exception {
+        oldContextClassLoader = Thread.currentThread().getContextClassLoader();
+        UsageStatistics.resetEntries();
+        fakeHilla(false);
```
taefi, line 47: It seems this makes the last test fail, so I suggest to manually call it in each the test:
```suggestion
```

packages/java/endpoint/src/test/java/dev/hilla/HillaStatsTest.java
```diff
@@ -0,0 +1,97 @@
...
+            classpath.add(fakeJar("hilla-react", "2.1.1"));
+        }
+        final ClassLoader classLoader = new URLClassLoader(
+                classpath.toArray(new URL[0]), null);
+        Thread.currentThread().setContextClassLoader(classLoader);
+    }
+
+    @Test
+    public void testLitIsReportedByDefault() {
+        HillaStats.report();
```
taefi, line 77:
```suggestion
        fakeHilla(false);
        HillaStats.report();
```

taefi (changes requested):

packages/java/endpoint/src/main/java/dev/hilla/EndpointController.java
mvysny: The `UsageStatistics.markAsUsed()` is called from multitude of places (also from Flow and from Vaadin Components), and it's always called directly (without being wrapped in a `try{} catch` block). I'd therefore suggest to follow that pattern; if need be, the `UsageStatistics.markAsUsed()` function itself should handle exceptions internally.

packages/java/endpoint/src/test/java/dev/hilla/HillaStatsTest.java
mvysny: Actually I think the problematic part was the memoization of `Platform.hillaVersion`, since that value depends on the context classloader. I've added a code which erases this memoization, it fixed the test.

taefi (dismissed):

taefi:
@mvysny the added tests are still failing.

platosha (dismissed):

mvysny:
@taefi sure, sorry for so many failed tests. The tests do pass in my Intellij unfortunately, so there's no other way but to push the code to the PR. I have added more verbose error logging which should shed some light into the nature of the failure.

mvysny:
The tests are failing in a random manner. @taefi are perhaps the tests run in parallel? If yes, we need to configure this test suite to run tests sequentially, since the tests are modifying a singleton map.

PR #1059, opened 2023-06-20
fix: fix package extraction failure when package declaration is not first line

### LINKED ISSUE #1048: Package search fails in Init App goal if the package is not at the first line
Spring-Initializr-generated apps have the package at the first line. When using `mvn hilla:init-app` on other applications, like the Spring Pet Clinic, it fails since there's a comment at the top of the file.

The regular expression can be improved to search for the package elsewhere in the file.

taefi:
## Description
This changes the package extraction approach for the class 
annotated with `@SpringBootApplication` by excluding all type
of java comments (single and multi-line), javadocs, and string 
literals, and then tries to extract the package declaration.  

Fixes #1048

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/packages/java/plugin-base/src/main/java/dev/hilla/plugin/base/HillaAppInitUtility.java
+++ b/packages/java/plugin-base/src/main/java/dev/hilla/plugin/base/HillaAppInitUtility.java
@@ -161,10 +161,12 @@ public class HillaAppInitUtility {
         try {
             var content = Files.readString(path);
             var regexToExcludeCommentsAndStringLiterals = "//.*|(\"(?:\\\\[^\"]|\\\\\"|.)*?\")|(?s)/\\*.*?\\*/";
-            var codeWithoutComments = content.replaceAll(regexToExcludeCommentsAndStringLiterals, "");
+            var codeWithoutComments = content
+                    .replaceAll(regexToExcludeCommentsAndStringLiterals, "");
 
             var regexToExtractPackage = "package\\b\\s+([a-zA-Z_][\\w.]*)\\s*;";
-            var matcher = Pattern.compile(regexToExtractPackage).matcher(codeWithoutComments);
+            var matcher = Pattern.compile(regexToExtractPackage)
+                    .matcher(codeWithoutComments);
             if (matcher.find()) {
                 return matcher.group(1);
             }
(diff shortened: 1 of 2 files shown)


packages/java/plugin-base/src/test/java/dev/hilla/plugin/base/HillaAppInitUtilityTest.java
```diff
@@ -158,18 +162,50 @@ public void withCorrectLitDependencies_scaffoldForLitSucceeds()
...
 
     private void createSpringBootApplicationClass() throws IOException {
 
         var content = """
-                package my;
+                /*
+                * Some comments that might also contains a package
+                * declaration statement:
+                * package my.package1.app;
+                */
```
platosha, line 191: Let's test this case also:
```suggestion
                /*
                Some comments that might also contains a package
                declaration statement:
                package my.package1.app;
                */
```

packages/java/plugin-base/src/main/java/dev/hilla/plugin/base/HillaAppInitUtility.java
```diff
@@ -159,8 +159,13 @@ private static String findSpringBootApplicationPackage(
     private static String extractPackage(Path path) {
         try {
             var content = Files.readString(path);
-            var matcher = Pattern.compile("^\\s*package\\s+([\\w.]+)\\s*;")
-                    .matcher(content);
+            var regexToExcludeCommentsAndStringLiterals = "//.*|(\"(?:\\\\[^\"]|\\\\\"|.)*?\")|(?s)/\\*.*?\\*/";
```
cromoteca, line 162: I was thinking that, instead of taking risks with regexps, we can also get the package from the relative path.

taefi: Working with paths also can work, however, with paths we need to consider skipping the `/src/main/java/` part or `src\main\java` in case of Windows. Then we might consider keeping it as Kotlin friendly as possible by considering to skip `src/main/kotlin` or `src\main\kotlin` for Windows, and I'd say it is not much more convenient either. (Supporting Kotlin is not our priority at the moment, anyway).

I'm not a fan of Regex myself, however, I'd prefer it over dealing with OS dependent paths :) 
I think there is not much of a risk if we have enough tests in place to make sure it works all known corner cases.

platosha (approved):

PR #849, opened 2023-03-16
refactor(1.3): cherry pick karma PR (#808)

Lodin:

PR #1075, opened 2023-06-28
feat: support ES-module application by adding .js to generated imports (#817) (CP: 1.3)

platosha:
* Add .js to imports

* Fix formatting

* Add missing .js

* Fix .js added twice to some imports

* Remove leaked dev logs

* test(generator): update snapshots for other plugins

---------

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16282,7 +16282,7 @@
     },
     "packages/ts/form": {
       "name": "@hilla/form",
-      "version": "1.3.8",
+      "version": "1.3.15",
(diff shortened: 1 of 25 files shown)


platosha:
Keeping the code in better sync with the main branch simplifies cherry-picks for maintenance fixes.

My motivation for cherry-picking this feature is to enable picking #849, which depends on some changes from here. In addition, there are ESlint code improvements coming up.

Lodin (approved):

PR #1076, opened 2023-06-28
build: add Windows run and skip-success workflows

### LINKED ISSUE #1073: Add Windows run to GitHub Validation workflow
Let us update the validation workflow to introduce testing on Windows with the goal of catching high-impact bugs.

Tasks:
- [ ] Run a Windows build / validation job in parallel with existing Linux jobs
- [ ] Do not depend on the `Build` due to platform differences
- [ ] Make sure at least the Gradle, Endpoint, and Security IT modules are run on Windows
- [ ] Remove the `paths-ignore` patterns that are currently blocking the validation for non-code changes

platosha:
Fixes #1073

### DIFF
--- a/.github/workflows/skip-validation.yml
+++ b/.github/workflows/skip-validation.yml
@@ -16,30 +16,9 @@ on:
       - 'packages/**/README.md'
 
 jobs:
-  check:
-    name: Check
-    runs-on: ubuntu-latest
-    timeout-minutes: 25
-    steps:
-      - uses: actions-cool/check-user-permission@main
-        id: checkUser
-        with:
-          username: ${{github.triggering_actor}}
-          require: 'write'
-      - name: Fail on external workflow triggering
-        if: ${{ steps.checkUser.outputs.require-result != 'true' }}
-        run: |
-          echo "🚫 **${{ github.actor }}** is an external contributor, a **${{github.repository}}** team member has to review this changes and re-run this build" \
-            | tee -a $GITHUB_STEP_SUMMARY && exit 1
-      - name: Check secrets
-        run: |
-          [ -z "${{secrets.TB_LICENSE}}" ] \
-            && echo "🚫 **TB_LICENSE** is not defined, check that **${{github.repository}}** repo has a valid secret" \
-            | tee -a $GITHUB_STEP_SUMMARY && exit 1 || exit 0
-
   test-all:
     name: Tests Summary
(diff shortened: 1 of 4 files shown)


.github/workflows/skip-validation.yml
```diff
@@ -0,0 +1,45 @@
+name: Validation
```
Lodin, line 1: Didn't know that we can set the same workflow name for several files. I thought it is going to be some `if-else` expressions withing the code 🤔

.github/workflows/skip-validation.yml
```diff
@@ -0,0 +1,45 @@
...
+  pull_request_target:
+    paths:
+      - 'hilla-logo.svg'
+      - 'README.md'
+      - '.github/ISSUE_TEMPLATE/*'
+      - 'packages/**/README.md'
+
+jobs:
+  check:
+    name: Check
```
Lodin, line 20: By the way, shouldn't it be a separate workflow? I mean, it is about user permission checking, I believe, it has nothing in common with the testing workflow. Isn't it so?

packages/java/tests/spring/pom.xml
```diff
@@ -203,6 +203,15 @@
       </modules>
     </profile>
 
+    <profile>
+      <id>it-modules-some</id>
```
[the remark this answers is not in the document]
platosha: I tried it, but in the context of a run title it looks a bit repetitive: `ITs (windodws-latest, windows)`

Consider the entire list of run titles:
image

.github/workflows/skip-validation.yml
platosha: Apparently it's a [documented workaround](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks).

.github/workflows/skip-validation.yml
platosha: The intention here is to prevent running any further steps in the testing workflow until the checks are satisfied. A separate workflow would bring more complexity around triggering one workflow from the other. I don't think it's worth it, compared to just few extra steps in the first job in the same file.

Now that you brought this up, however, I realised that the skip workflow does not run any testing steps anyway. We don't need to check any permissions here then. I removed this unnecessary job.

Lodin (approved):

PR #1084, opened 2023-06-30
POC: feat: add support for re-generating endpoints without restart

### LINKED ISSUE #863: Improve the Hotswap / JRebel Sync for Endpoints and Typescript generation

Our application takes 5-10 minutes to start, we want to add new / change endpoints and use them in the frontend without having to restart the backend. Currently one would have to restart the application to achieve this, new endpoint methods wouldn't be available as they are registered on startup.

Using a hotswap solution like the Hotswap Agent or JRebel we want to have a possiblity to access the new endpoints and be able to regenerate the typescript hilla sources.
The Hilla devmode Gizmo or a general Hilla-Framework-Endpoint could be a place to add a trigger for the regeneration, sync of the endpoint registry.

We discussed this approach with @TatuLund .

We could run mvn hilla:generate to regenerate the typescript sources, but our hilla module is a child module in a big structure and there are issues on running it just for the module. Even if that would work, it would still potentially take up to 20 minutes. For the re-registration of the endpoints we can't think of any alternative than to try to adapt core parts of the hilla framework (the dev.hilla.EndpointRegistry)

taefi:
## Description

This feature is based on having either HotSwap Agent or JRebel set in the development environment for
reloading classes so that Parser and Generator can re-generate the needed ts files based on the latest compiled classes.

In this approach, a filesystem WatchService will start monitoring the classes directory (in a separate thread) for changes if:

- Application is running in Dev mode, and
- Dev mode live reload is enabled, and
- The backend support for live reload is not Spring Boot DevTools

When a change in the classes is detected, the Parser and then Generator will update the openapi.json, and since this file reside under classes itself, then its changes also detected which will trigger the re-registration of the Endpoints, without a need to restart the application.

**Limitations**:
The limitations are dictated by the limitations of JRebel or HotSwapAgent to redefining the classes at runtime. Some preliminary local tests shows that HotSwapAgent is working much smoother than JRebel. For example, JRebel sometimes fails to redefine the method renames in the memory, although, Hilla is reloading the change based on the latest changes completely fine. HowSwapAgent shows a faster reload as well compared to JRebel.

If Only Spring Boot DevTools is in use in an application, the watch service will not register for changes, so the developer experience will not change.

Developer, do not need to do any setup or changes to their project in order to enable this feature, other than setting up the HotSwapAgent or JRebel in their development environment.

Fixes #863 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [ ] New and existing tests are passing locally with my change.

### DIFF
--- a/packages/java/endpoint/src/main/java/dev/hilla/EndpointControllerConfiguration.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/EndpointControllerConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2000-2022 Vaadin Ltd.
+ * Copyright 2000-2023 Vaadin Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -18,6 +18,7 @@ package dev.hilla;
 
 import java.lang.reflect.Method;
 
+import dev.hilla.internal.hotswap.HotSwapConfiguration;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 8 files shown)


tarekoraby:
> This feature is based on having either HotSwap Agent or JRebel set in the development environment for
reloading classes

@taefi, I understand that for large apps like the one described in https://github.com/vaadin/hilla/issues/863, HotSwap Agent or JRebel would be needed to make the process faster. But for smaller apps, will this POC also work if I'm using the live reload of  Spring Boot Developer Tools?

taefi:
@tarekoraby thanks for the comment. 
I think the following part of the description is addressing your concern already:

> If Only Spring Boot DevTools is in use in an application, the watch service will not register for changes, so the developer experience will not change.

If not, let me know to discuss further.

taefi:
Closed as not relevant anymore.

PR #1104, opened 2023-07-19
chore: enable integration tests to use java 20 as latest

taefi:
At the moment, latest final java version is 20,
but the endpoints-latest-java IT module was still
using java 17. This will set the latest java version 
to be used in that specific module to 20.


Lodin (changes requested):
I believe you also need to change the default version in CI as well [[link]](https://github.com/vaadin/hilla/blob/main/.github/actions/setup/action.yml#L12)

packages/java/tests/spring/endpoints-latest-java/pom.xml
```diff
@@ -12,9 +12,9 @@
     <packaging>jar</packaging>
 
     <properties>
-        <maven.compiler.source>17</maven.compiler.source>
-        <maven.compiler.target>17</maven.compiler.target>
-        <maven.compiler.release>17</maven.compiler.release>
+        <maven.compiler.source>20</maven.compiler.source>
```
Artur-, line 15: I would jump to 21 directly

taefi: Well it seems that GH Actions doesn't support installing the 21 yet.

Artur-: True, it will probably not appear until next week

Lodin: Is 21 LTS? I would prefer to stick with LTS version to not ship code that cannot compile with LTS

PR #1130, opened 2023-08-04
refactor(hilla-frontend): apply vaadin eslint rules for the package

Lodin:


packages/ts/hilla-frontend/src/types.d.ts
```diff
@@ -0,0 +1,23 @@
...
+    Flow?: VaadinFlow;
+    connectionIndicator?: ConnectionIndicator;
+    connectionState?: ConnectionStateStore;
+    registrations?: VaadinRegistration[];
+  }
+
+  interface Window {
+    Vaadin?: Vaadin;
+  }
+}
```
cromoteca, line 23: Can it clash with https://github.com/vaadin/sso-kit/blob/5c44f9c60a2c7799b47e66ae28b57d7324329711/sso-kit-client/react/src/index.ts#L18-L35 ?

cromoteca (changes requested):

packages/ts/hilla-frontend/src/types.d.ts
cromoteca: I was probably wrong on requesting changes: if needed, that's more on the kit side.

Lodin: It's ok 🙂 It's better to block merging to have an answer than to merge a wrong code.

> Can it clash
Well, it shouldn't, as far as I know. However, it's a good idea to make a separate check.

Lodin: Let's have it defined as a local interface

packages/ts/generator-typescript-core/src/Plugin.ts
```diff
@@ -1,5 +1,7 @@
-import type { Logger } from '@hilla/generator-typescript-utils/LoggerFactory.js';
+// eslint-disable-next-line import/no-duplicates
```
cromoteca, line 1: Not big deal, but it hurts me to see a rule that would forbid a necessary language syntax.

The `no-duplicates` docs says:

> If the core ESLint version is good enough (i.e. you're not using Flow and you are using [import/extensions](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md)), keep it and don't use this.

Could that allow to avoid this thing that I consider a bug?

cromoteca (changes requested):

packages/ts/generator-typescript-core/src/Plugin.ts
Lodin: 🤔 Well, that's a good point. However, I also have some unnecessary dupes removed with this rule. Another option is to get rid of default imports completely in `utils` library (which also looks a nice option to me). What do you think?

cromoteca: I don't know, `LoggerFactory.ts` looks fine as it is. TBH, my preferred option would be to keep things as they are and file a bug in `eslint-plugin-import`.

packages/ts/hilla-frontend/src/Connect.ts
```diff
@@ -247,26 +251,26 @@ export interface MiddlewareContext extends EndpointCallMetaInfo {
...
-  invoke(context: MiddlewareContext, next: MiddlewareNext): Promise<Response> | Response;
+  invoke(context: MiddlewareContext, next?: MiddlewareNext): MaybePromise<unknown>;
 }
 
 /**
  * An async callback function that can intercept the request and response
  * of a call.
  */
-type MiddlewareFunction = (context: MiddlewareContext, next: MiddlewareNext) => Promise<Response> | Response;
+export type MiddlewareFunction = (context: MiddlewareContext, next?: MiddlewareNext) => MaybePromise<unknown>;
```
platosha, line 273: Why `next` is optional? It is always provided for the middleware, and the middleware should normally invoke it.

cromoteca (approved):

packages/ts/hilla-frontend/src/Connect.ts
```diff
@@ -247,26 +249,26 @@ export interface MiddlewareContext extends EndpointCallMetaInfo {
...
  */
 export interface MiddlewareClass {
   /**
-   * @param context The information about the call and request
-   * @param next Invokes the next in the call chain
+   * @param context - The information about the call and request
+   * @param next - Invokes the next in the call chain
    */
-  invoke(context: MiddlewareContext, next: MiddlewareNext): Promise<Response> | Response;
+  invoke(context: MiddlewareContext, next?: MiddlewareNext): MaybePromise<unknown>;
```
platosha, line 264: Same API change here, `next` should be mandatory:

```suggestion
  invoke(context: MiddlewareContext, next: MiddlewareNext): MaybePromise<unknown>;
```

platosha (changes requested):

platosha (approved):

platosha:
@Lodin this needs a rebase. Could you, please, update the branch?

platosha (approved):

PR #1143, opened 2023-08-16
refactor(form): apply Vaadin ESLint rules

platosha:
Depends on #1130


packages/ts/form/src/Field.ts
```diff
@@ -1,13 +1,16 @@
+/* eslint-disable accessor-pairs,sort-keys */
 import { type ElementPart, noChange, nothing, type PropertyPart } from 'lit';
 import { directive, Directive, type DirectiveParameters, type PartInfo, PartType } from 'lit/directive.js';
+import { BooleanModel } from '../Models';
```
cromoteca, line 4:
```suggestion
import { BooleanModel } from '../Models.js';
```

cromoteca (changes requested):

Lodin:
Have you decided to postpone moving to ES native private members? I see several places where you could do that safely in the code.

packages/ts/form/src/Validators.ts
```diff
@@ -310,7 +310,7 @@ export class Digits<T> extends AbstractValidator<T> {
 
   override validate(value: T): boolean {
     return (
-      String(toFloat(String(value))).replace(/(.*)\.\d+/u, '$1').length <= this.integer &&
+      String(toFloat(String(value))).indexOf('.') <= this.integer &&
```
cromoteca, line 313: This change seems OK to me, anyway I question both implementation about negative number, as the minus sign seems to be taken into account here. 

There are no tests for negative numbers indeed.

I'm going to open an issue for that.

Lodin: I believe that we shouldn't do such kind of changes in this PR. Could we ignore the Sonar quality gates and create an issue for these issues?

cromoteca: Except that it's blocking merging

cromoteca: Concerning the minus sign yes, I opened a different issue.

Lodin: Hmm, I was sure we could ignore Sonar if it's better for us 🤔

Lodin: @platosha what do you think?

platosha: let's address in a separate PR

platosha: Sonar is red but not blocking I think

Lodin (dismissed):

packages/ts/form/src/Validators.ts
cromoteca: OK so I'm going to revert my changes.

cromoteca (approved):

PR #1146, opened 2023-08-16
feat: add support for re-generating endpoints without restart

### LINKED ISSUE #863: Improve the Hotswap / JRebel Sync for Endpoints and Typescript generation

taefi:
## Description

This feature is based on having either HotSwap Agent or JRebel set in the development environment for
reloading classes so that Parser and Generator can re-generate the needed TS files based on the latest compiled classes.

In this approach, an initiated service at startup polls for the changes of the endpoints and their associated classes, and by comparing the openapi contents with the previous state it will regenerate the needed TS files which makes them available to the views without restarting the application.

The polling service will start monitoring the changes (in a separate thread) if:

- Application is running in Dev mode, and
- Dev mode live reload is enabled, and
- The backend support for live reload is not Spring Boot DevTools

When a development environment meet the above criteria, the hot-reload feature is enabled by default. However, it can be disabled by setting the following property to false:

- `hilla.endpoint.hot-reload.enabled`

The default monitoring poll interval is 5 seconds, which is configurable by setting the following property:

- `hilla.endpoint.hot-reload.pollInterval`

**Limitations**:
The limitations are dictated by the limitations of JRebel or HotSwapAgent to redefining the classes at runtime. Some preliminary local tests shows that HotSwapAgent is working much smoother than JRebel. For example, JRebel sometimes fails to redefine the method renames in the memory, although, Hilla is reloading the change based on the latest changes completely fine. HowSwapAgent shows a faster reload as well compared to JRebel.

If Only Spring Boot DevTools is in use in an application, the watch service will not register for changes, so the developer experience will not change.

Developer, do not need to do any setup or changes to their project in order to enable this feature, other than setting up the HotSwapAgent or JRebel in their development environment.

Fixes #863 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


packages/java/endpoint/src/main/java/dev/hilla/internal/hotswap/EndpointHotSwapListener.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+        ParserProcessor parser = new ParserProcessor(engineConfiguration,
+                this.getClass().getClassLoader());
+        parser.process();
+        GeneratorProcessor generator = new GeneratorProcessor(
+                engineConfiguration, "node");
+        generator.process();
+
+        this.endpointController.registerEndpoints();
+        reload(event.browserLiveReload());
```
Artur-, line 70: Why should you reload the browser if an endpoint changes?

Artur-: Vite watches the frontend files and will do hot replace or reload or whatever is needed for that

packages/java/endpoint/src/main/java/dev/hilla/internal/hotswap/HotSwapConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class HotSwapConfiguration {
+
+    /**
+     * The interval for polling the changes in seconds.
+     */
+    @Value("${hilla.endpoint.hot-reload.pollInterval:5}")
```
Artur-, line 31: You should aim to declare properties in a way so that tooling can pick it up and offer auto complete, like in https://github.com/vaadin/flow/blob/main/vaadin-spring/src/main/java/com/vaadin/flow/spring/VaadinConfigurationProperties.java

packages/java/endpoint/src/main/java/dev/hilla/internal/hotswap/EndpointHotSwapListener.java
taefi: IMO, It is more convenient for the developer to see where in view's code might be broken due to the changes right away, rather than start entering data and clicking button to see errors in the browser's console. 

Reloading the browser automatically triggers the clientside devtools to show the error in the view code with nice styles and has no bad effects on anything (to my understanding). Do you see a pitfall cause by reloading the page?

Artur-: Why does not Vite handle this?

taefi: I was experiencing some unreliability when Vite wasn't conveniently hot-reloading the page (as it supposed to). However, at that time, I tried this with FSWatch service to regenerate the TS files, so there was a chance of things happening not in the same order as I expected... I'll give it a try to remove the BrowserReload and see what happens.

taefi: BTW, I had to disable the Vaadin HotSwap plugin since it is reloading the browser with a different timing and in a separate thread that we need to in Hilla.

tarekoraby: > Why should you reload the browser if an endpoint changes?

What is the harm in doing that? If no significant harm is made, why bother?

packages/java/endpoint/src/main/java/dev/hilla/internal/hotswap/HotSwapConfiguration.java
taefi: Done

packages/java/endpoint/src/main/java/dev/hilla/internal/hotswap/EndpointHotSwapListener.java
taefi: I disabled the browser reload and Vite is not triggering the hot-replace/reload. 
Currently, I don't know why it isn't working, since the generated files have already changed under `frontend/generated` directory, and even the IDE is detecting the change (e.g. renamed endpoint method). I suggest to ignore this for now and improve it later.

Artur- (approved):

PR #1151, opened 2023-08-17
chore: address SonarCloud DoS-related issues

cromoteca:
Note: **does not** fix https://github.com/vaadin/hilla/issues/1150

PR #1152, opened 2023-08-17
fix: don't count minus sign in integer digits

### LINKED ISSUE #1150: The `Digits` validator seems to be incorrect for negative numbers
### Description of the bug 

By looking at the [current implementation](https://github.com/vaadin/hilla/blob/4752a86a745fdb3b096a877241f79f452fe62fc2/packages/ts/form/src/Validators.ts#L310), the regular expression seems to count the minus sign as a digit. Is that correct?

I verified the Java `Digits` validation and it does not count the minus sign.

There are no tests for negative numbers.

I extended [the tests](https://github.com/vaadin/hilla/blob/4752a86a745fdb3b096a877241f79f452fe62fc2/packages/ts/form/test/Validators.test.ts#L247-L252) as such:
```typescript
      assert.isTrue(validator.validate('11.111'), 'Exact number of digits');
      assert.isTrue(validator.validate('1.1'), 'Less digits');
      assert.isTrue(validator.validate('1'), 'Less digits and no fraction');
      assert.isFalse(validator.validate('1.1111'), 'More fractional digits');
      assert.isFalse(validator.validate('111.111'), 'More integer digits');
      assert.isFalse(validator.validate('111.1111'), 'More integer and fractional digits');
      assert.isTrue(validator.validate('-11.111'), 'Exact number of digits, negative number');
      assert.isTrue(validator.validate('-1.1'), 'Less digits, negative number');
      assert.isTrue(validator.validate('-1'), 'Less digits and no fraction, negative number');
      assert.isFalse(validator.validate('-1.1111'), 'More fractional digits, negative number');
      assert.isFalse(validator.validate('-111.111'), 'More integer digits, negative number');
      assert.isFalse(validator.validate('-111.1111'), 'More integer and fractional digits, negative number');
```
and it fails indeed.

Hilla version: 2.1.5

cromoteca:
Closes #1150.

Depends on #1151.

### DIFF
--- a/packages/ts/form/src/Validators.ts
+++ b/packages/ts/form/src/Validators.ts
@@ -310,7 +310,7 @@ export class Digits<T> extends AbstractValidator<T> {
 
   override validate(value: T): boolean {
     return (
-      String(toFloat(String(value))).replace(/(.*)\.\d+/u, '$1').length <= this.integer &&
+      String(toFloat(String(value))).indexOf('.') <= this.integer &&
       // eslint-disable-next-line camelcase
       isDecimal(String(value), { decimal_digits: `0,${this.fraction}` })
     );
--- a/packages/ts/form/src/Validators.ts
+++ b/packages/ts/form/src/Validators.ts
@@ -310,7 +310,7 @@ export class Digits<T> extends AbstractValidator<T> {
 
   override validate(value: T): boolean {
     return (
-      String(Math.abs(toFloat(String(value)))).indexOf('.') <= this.integer &&
+      String(Math.floor(Math.abs(toFloat(String(value))))).length <= this.integer &&
       // eslint-disable-next-line camelcase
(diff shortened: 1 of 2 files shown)


packages/ts/form/src/Validators.ts
```diff
@@ -310,7 +310,7 @@ export class Digits<T> extends AbstractValidator<T> {
 
   override validate(value: T): boolean {
     return (
-      String(toFloat(String(value))).indexOf('.') <= this.integer &&
+      String(Math.abs(toFloat(String(value)))).indexOf('.') <= this.integer &&
```
platosha, line 313: `.indexOf()` returns `-1` if there is no decimal separator. Should we do this instead?
```suggestion
      String(Math.floor(Math.abs(toFloat(String(value))))).length <= this.integer &&
```

platosha (dismissed):

platosha (approved):

PR #1157, opened 2023-08-21
feat: display React form value and validation state

platosha:
Depends on #1149

### DIFF
--- a/packages/ts/react-form/src/index.ts
+++ b/packages/ts/react-form/src/index.ts
@@ -108,15 +108,13 @@ function useFields<T, M extends AbstractModel<T>>(node: BinderNode<T, M>): Field
     return ((model: AbstractModel<any>) => {
       const n = getBinderNode(model);
 
-      const fieldState: FieldState<unknown> = registry.has(model)
-        ? registry.get(model)!
-        : {
-          value: undefined,
-          required: false,
-          invalid: false,
-          errorMessage: '',
-          strategy: undefined,
-        };
+      const fieldState: FieldState<unknown> = registry.get(model) ?? {
+        value: undefined,
+        required: false,
+        invalid: false,
+        errorMessage: '',
+        strategy: undefined,
+      };
 
       if (!registry.has(model)) {
         registry.set(model, fieldState);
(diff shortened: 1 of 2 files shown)


Lodin (approved):

packages/ts/react-form/src/index.ts
```diff
@@ -85,56 +101,96 @@ function getBinderNodeControls<T, M extends AbstractModel<T>>(
...
-        if (!strategy || strategy.model !== model) {
-          strategy = getDefaultFieldStrategy(fld, model);
-          registry.current.set(fld, strategy);
+        if (fieldState.required !== n.required) {
+          fieldState.required = n.required;
+          fieldState.strategy.required = n.required;
         }
 
-        const elementValue = strategy.value;
+        const firstError: ValueError<any> | undefined = n.ownErrors[0];
```
Lodin, line 137:
```suggestion
        const [firstError] = n.ownErrors;
```

packages/ts/react-form/src/index.ts
```diff
@@ -85,56 +101,96 @@ function getBinderNodeControls<T, M extends AbstractModel<T>>(
...
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        const errorMessage = firstError?.message || '';
```
Lodin, line 139:
```suggestion
        const errorMessage = firstError?.message ?? '';
```

packages/ts/react-form/src/index.ts
Lodin: or we could also use:
```suggestion
        const firstError = n.ownErrors.at(0);
```

Lodin (changes requested):

packages/ts/react-form/src/index.ts
```diff
@@ -85,56 +101,96 @@ function getBinderNodeControls<T, M extends AbstractModel<T>>(
...
+
+      const fieldState: FieldState<unknown> = registry.has(model)
+        ? registry.get(model)!
+        : {
+          value: undefined,
+          required: false,
+          invalid: false,
+          errorMessage: '',
+          strategy: undefined,
+        };
```
Lodin, line 119:
```suggestion
      const fieldState: FieldState<unknown> = registry.get(model) ?? {
        value: undefined,
        required: false,
        invalid: false,
        errorMessage: '',
        strategy: undefined,
      };
```

Lodin (changes requested):

PR #1162, opened 2023-08-23
fix(ts): make Binder work in non-browser environment

platosha:


packages/ts/hilla-frontend/package.json
```diff
@@ -26,16 +26,24 @@
   },
   "exports": {
     ".": {
-      "default": "./index.js"
+      "default": "./index.js",
+      "_types": "./index.d.ts"
```
Lodin, line 30: Why underscore?

platosha: 🤦‍♂️

Lodin (approved):

PR #1163, opened 2023-08-24
refactor: Move endpoint re-generation code to a new class

Artur-:
Also uses the node executable from FrontendTools to work in all cases


packages/java/endpoint/src/main/java/dev/hilla/EndpointCodeGenerator.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
```
taefi, line 2:
```suggestion
 * Copyright 2000-2023 Vaadin Ltd.
```

packages/java/endpoint/src/main/java/dev/hilla/EndpointCodeGenerator.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * Gets the singleton instance.
+     * 
+     * @throws IOException
+     */
+    public static EndpointCodeGenerator getInstance() {
+        return instance;
+    }
```
taefi, line 72: I suggest avoiding the singleton anti-pattern. We should be able to manage passing a singleton instance through the Spring configuration and constructor injection where needed.

taefi (changes requested):

packages/java/endpoint/src/main/java/dev/hilla/EndpointCodeGenerator.java
Artur-: The consumer of this would be e.g. a Hotswap agent plugin

taefi: Well, nothing else can be done I guess.

Artur-: Hotswap agent cannot be aware of Spring Boot so it is better to do the hacky Spring context "lookup" inside here instead of in a HA plugin

Artur-: I removed it for now, as it is not strictly needed for this change

taefi (approved):

PR #1171, opened 2023-08-24
fix(react-form): consistent render updates

platosha:


Lodin (approved):
Wow, that's quite interesting solution. If I understand correctly, you're trying to save the value set during the `constructor` run, right?

PR #1172, opened 2023-08-25
refactor: Extract util for finding all classes used in openapi.json

Artur-:
Removes the Aot processor that finds endpoint classes as they are also detected from openapi.json

PR #1173, opened 2023-08-25
feat: Callback method for hotswap solutions

Artur-:
Hotswapper.onHotswap can be called by a list of changed classes (as strings) by any hotswap solution and Hilla will react to the changed classes in an appropriate way, for instance by re-generating the endpoint metadata and TypeScript. The hotswap solution is not supposed to do any pre-filtering of the changed classes, this is done by the method itself.


Artur-:
Based on #1172  and #1163

packages/java/endpoint/src/main/java/dev/hilla/EndpointCodeGenerator.java
```diff
@@ -36,8 +37,12 @@ public class EndpointCodeGenerator {
     private final EndpointController endpointController;
     private final VaadinContext context;
     private Path buildDirectory;
+
     private ApplicationConfiguration configuration;
     private String nodeExecutable;
+    private Set<String> classesUsedInOpenApi = null;
+
+    private static EndpointCodeGenerator instance = null;
```
cromoteca, line 45: Where's its value initialized?

Artur-: Feels like something has gone missing during the rebase

Artur-: Added

cromoteca (approved):

PR #1174, opened 2023-08-25
Chore: build with flow 24.1-SNAPSHOT

ZheSun88:
As what has been told in the monthly meeting, hilla 2.2 will be released with flow 24.1


Artur-:
Do we even need snapshot or can we just use the latest stable?

ZheSun88:
then with every flow release, we need update here.. and we cannot catch the error if it comes from flow snapshot. 

during the release step, we already has the step to set the flow.version

Artur-:
I don't think there is any need to use Hilla as a test for Flow snapshots. You see any potential problems when you update the dependency version, like with all other dependencies

manolo:
Let's merge as it's so as we can release the alpha, when builds in TC etc, are modified, we will need to update this and have fixed versions instead

manolo (dismissed):

PR #1178, opened 2023-08-25
fix(TypeScript): consistent validation updates in React forms

### LINKED ISSUE #1169: React Form Binder: error message missing when deleting many characters at once and another validator is in place
I have a Java field annotated as such:
```java
    @NotBlank(message = "First name is mandatory")
    @Size(max = 10, message = "First name cannot be longer than 10 characters")
```
Now, if I delete all characters in the corresponding React form field, it triggers the first message correctly. If I write some characters and delete them, is triggers the message again. If I type more than 10 characters, it triggers the second message. But now, if I delete all characters at once, no error message is shown.

This issue happens on current `main`.

platosha:
Fixes #1169


cromoteca (approved):

cromoteca:
I'll open a separate issue for the flashing effect I've seen while testing manually.

PR #1194, opened 2023-08-30
fix(form): gracefully handle validator throwing an error

sissbruecker:
## Description

Adds error handling to validator executions, so that binder validation and update can still complete in case a validator unexpectedly throw an error.

## Type of change

- Bugfix

### DIFF
--- a/packages/ts/form/src/Validation.ts
+++ b/packages/ts/form/src/Validation.ts
@@ -97,28 +97,33 @@ export async function runValidator<T>(
   if (!binderNode.required && !new Required().validate(value) && !(model instanceof NumberModel)) {
     return [];
   }
-  return (async () => validator.validate(value, binderNode.binder))().then((result) => {
-    if (result === false) {
-      return [{ message: interpolateMessage(validator.message), property: binderNode.name, validator, value }];
-    }
-    if (result === true || (Array.isArray(result) && result.length === 0)) {
-      return [];
-    }
-    if (Array.isArray(result)) {
-      return result.map((result2) => ({
-        message: interpolateMessage(validator.message),
-        ...setPropertyAbsolutePath(binderNode.name, result2),
-        validator,
-        value,
-      }));
-    }
-    return [
-      {
-        message: interpolateMessage(validator.message),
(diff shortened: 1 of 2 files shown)


platosha (dismissed):

packages/ts/form/src/Validation.ts
```diff
@@ -97,28 +97,33 @@ export async function runValidator<T>(
...
-        ...setPropertyAbsolutePath(binderNode.name, result as ValidationResult),
-        validator,
-        value,
-      },
-    ];
-  });
+  return (async () => validator.validate(value, binderNode.binder))()
+    .catch((error) => {
+      console.error(`${binderNode.name} - Validator ${validator.constructor.name} threw an error:`, error);
+      return true;
```
taefi, line 103: Isn't returning `true` here mean the validation has passed?

taefi (changes requested):

packages/ts/form/src/Validation.ts
sissbruecker: Yes, currently it is implemented so that a faulty (client-side) validator does not cause an invalid state / prevent form submission. Whether that is the right way to go is of course up for debate 🙂. It feels like both approaches are rather opinionated.

taefi: Well, I was thinking of a situation that in a near future, we make some breaking/behavioral changes in a major/minor version that causes validation exceptions in user codes, and as the result their applications allows inconsistent data to be saved. Of course, the applications should run tests before rolling out new production deployment, but in my experience, I'd feel much safer to see failures in case an exception occurs in a validation rather than letting it pass. As you mentioned, it might be opinionated anyway.

sissbruecker: I'm open to changing it so that a broken validator prevents submitting the form.

@platosha What do you think?

sissbruecker: We agreed in a discussion to change the behavior so that a validator throwing an exception is also reported as a validation error. Changed the implementation accordingly.

taefi (approved):

PR #1215, opened 2023-09-07
fix(parser-jvm-plugin-backbone): Properly expand generic types

### LINKED ISSUE #1214: Childendpoints from a typed @EndpointExposed Baseclasses have wrong method parameter types
### Description of the bug 
When you have a typed Endpoint annotated with @EndpointExposed and multiple children extend this endpoint, it is possible that wrong method parameters are used for the second and later endpoints if they have the same method name.

Have a typed exposed endpoint with the type as a method parameter:
```
@EndpointExposed
public abstract class BaseEndpoint<T> {

    public String getAll(T param) {
        return null;
    }

}
```

Then have two endpoints extending the baseendpoint with different types:

```
@Endpoint
@AnonymousAllowed
public class HelloWorldEndpoint extends BaseEndpoint<String>{

}
```

```
@Endpoint
@AnonymousAllowed
public class SecondEndpoint extends BaseEndpoint<Long>{
}
```

Now generate the endpoints. Both endpoints have the same method parameter type:

HelloWorldEndpoint.ts

`async function getAll_1(param: number | undefined, init?: EndpointRequestInit_1): Promise<string | undefined> { return client_1.call("HelloWorldEndpoint", "getAll", { param }, init); }`

SecondEndpoint.ts
`async function getAll_1(param: number | undefined, init?: EndpointRequestInit_1): Promise<string | undefined> { return client_1.call("SecondEndpoint", "getAll", { param }, init); }`

This only happens with methods from the baseclass. If both endpoints implement their own methods, they have the correct parameter types. 

Reproducable Repo:

https://github.com/AirbornePanda/hilla_endpoint_error_example

### Expected behavior
The get_All function of the HelloWorldEndpoint.ts should have string as method parameter type.
### Actual behavior
The get_All function of the HelloWorldEndpoint.ts has number as method parameter type.
### Versions:
    - Vaadin / Hilla version: 2.2.0.beta1
    - Node version: v18.16.1
    - Java version: 19
    - OS version: Win 10
    - Browser version (if applicable): Irrelevant
    - Application Server (if applicable): Spring Boot
    - IDE/Editor (if applicable): IntelliJ

Artur-:
When a method is defined in the super class, the definition needs to be duplicated for each implementing class so that generics can be set separately for each instance

Fixes #1214

### DIFF
--- a/packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/EndpointExposedPlugin.java
+++ b/packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/EndpointExposedPlugin.java
@@ -11,6 +11,7 @@ import dev.hilla.parser.core.NodePath;
 import dev.hilla.parser.models.AnnotationInfoModel;
 import dev.hilla.parser.models.ClassInfoModel;
 import dev.hilla.parser.models.ClassRefSignatureModel;
+import dev.hilla.parser.models.MethodInfoModel;
 import dev.hilla.parser.models.SignatureModel;
 import dev.hilla.parser.models.TypeArgumentModel;
 import dev.hilla.parser.models.TypeParameterModel;
@@ -19,6 +20,7 @@ import dev.hilla.parser.plugins.backbone.nodes.EndpointExposedNode;
 import dev.hilla.parser.plugins.backbone.nodes.EndpointNode;
 import dev.hilla.parser.plugins.backbone.nodes.EndpointNonExposedNode;
 import dev.hilla.parser.plugins.backbone.nodes.EndpointSignatureNode;
+import dev.hilla.parser.plugins.backbone.nodes.MethodNode;
 import dev.hilla.parser.plugins.backbone.nodes.TypeSignatureNode;
 
 public final class EndpointExposedPlugin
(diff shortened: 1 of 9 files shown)


packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/EndpointExposedPlugin.java
```diff
@@ -35,6 +37,10 @@ public void exit(NodePath<?> nodePath) {
     @Override
     public Node<?, ?> resolve(@Nonnull Node<?, ?> node,
             @Nonnull NodePath<?> parentPath) {
+        if (node instanceof MethodNode
+                && parentPath.getNode() instanceof EndpointExposedNode) {
```
cromoteca, line 41: What if the `EndpointExposed` class is not the parent one, but it's up in the hierarchy?

cromoteca: The [POC Jackson parser](https://github.com/cromoteca/jackson-parser/blob/8813055c665142f951280e2b25f31590a31de99f/generator/src/main/java/com/cromoteca/parser/Parser.java#L57-L66) I wrote when testing Jackson for the generator does a better job in my opinion about how to find methods.

Artur-: The thought also occured to me that EndpointExposed is probably not the only case. If a more generic solution is implemented, this fix can be reverted and the test kept.

platosha (approved):

PR #1220, opened 2023-09-08
refactor(parser-jvm): move synthetic method filtering to BackbonePlugin

Lodin:
This PR is a small fix for #1217.

I moved filtering of synthetics methods from the model to `BackbonePlugin` because the design idea of the model is to provide the fullest info about the reflection item. Filtering synthetic methods breaks this idea, so I believe that the best place for this filtering is in the `BackbonePlugin`.

### DIFF
--- a/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoReflectionModel.java
+++ b/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/models/ClassInfoReflectionModel.java
@@ -217,7 +217,6 @@ final class ClassInfoReflectionModel extends ClassInfoModel
     @Override
     protected List<MethodInfoModel> prepareMethods() {
         return Arrays.stream(origin.getDeclaredMethods())
-                .filter(method -> !method.isSynthetic())
                 .map(MethodInfoModel::of).sorted(MethodInfoModel.METHOD_ORDER)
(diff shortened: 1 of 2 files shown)


cromoteca:
I disagree: from the parser/generator point of view, synthetic methods should be ignored as soon as possible.

Lodin:
@cromoteca why it is so required? Cannot there be any case that would require using synthetic methods somewhere in plugins?

cromoteca:
> why it is so required? Cannot there be any case that would require using synthetic methods somewhere in plugins?

Synthetic methods are added to classes by many tools, including the JDK itself. If we start dealing with them, our classes get "polluted" in many ways. I strongly recommend to stick to "real" methods from the start, so that nothing will interfere with parsing.

As far as I remember, we already dealt with them in the past, but maybe some code changes reintroduced them.

Lodin:
> As far as I remember, we already dealt with them in the past, but maybe some code changes reintroduced them.

Yeah, I definitely remember that we had a filter for it somewhere but I cannot find it anywhere.

> Synthetic methods are added to classes by many tools, including the JDK itself. If we start dealing with them, our classes get "polluted" in many ways. I strongly recommend to stick to "real" methods from the start, so that nothing will interfere with parsing.

Hmm... Ok, let's keep it simple then. If we ever have a necessity, we could just get it back anytime.

PR #1221, opened 2023-09-08
feat: Introduce BrowserCallable as a synonym for Endpoint

Artur-:
This is to test what feels more natural to use

The terminology would be that you mark a service to be available for the browser, using `@BrowerCallable`

For now, you need to put `@EndpointExposed` on the super class when using `@BrowserCallable`. For that, we should come up with a better name also

### DIFF
--- a/packages/java/endpoint/src/main/java/dev/hilla/BrowserCallable.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/BrowserCallable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2000-2022 Vaadin Ltd.
+ * Copyright 2000-2023 Vaadin Ltd.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
--- a/packages/java/endpoint/src/main/java/dev/hilla/BrowserCallable.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/BrowserCallable.java
@@ -28,6 +28,8 @@ import org.springframework.stereotype.Component;
  * For each class, a corresponding TypeScript class is generated in
  * {@code frontend/generated} with TypeScript methods for invoking the methods
  * in this class.
+ * <p>
+ * This is an alias for {@link Endpoint}.
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
--- a/packages/java/endpoint/src/main/java/dev/hilla/BrowserCallable.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/BrowserCallable.java
@@ -20,13 +20,17 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+import org.springframework.stereotype.Component;
+
 /**
- * Annotation marks the class itself should not be treated as an
- * {@link Endpoint} but inheritable methods will be added to the
- * decsendant Endpoints.
+ * Makes the methods of the annotated class available to the browser.
+ * <p>
+ * For each class, a corresponding TypeScript class is generated in
+ * {@code frontend/generated} with TypeScript methods for invoking the methods
+ * in this class.
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
-public @interface EndpointExposed {
-
+@Component
+public @interface BrowserCallable {
 }
--- a/packages/java/endpoint/src/main/java/dev/hilla/EndpointExposed.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/EndpointExposed.java
@@ -22,7 +22,7 @@ import java.lang.annotation.Target;
 
 /**
  * Annotation marks the class itself should not be treated as an
- * {@link Endpoint} but inheritable methods will be added to the descendant
+ * {@link Endpoint}, but inheritable methods will be added to the descendant
  * Endpoints.
  */
 @Target(ElementType.TYPE)
--- a/packages/java/endpoint/src/main/java/dev/hilla/EndpointExposed.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/EndpointExposed.java
@@ -22,7 +22,7 @@ import java.lang.annotation.Target;
 
 /**
  * Annotation marks the class itself should not be treated as an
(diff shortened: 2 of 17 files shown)


packages/java/endpoint/src/main/java/dev/hilla/EndpointExposed.java
```diff
@@ -22,8 +22,8 @@
 
 /**
  * Annotation marks the class itself should not be treated as an
- * {@link Endpoint}. Instead the inheritable methods will be added to the
- * decsendant Endpoints.
+ * {@link Endpoint} but inheritable methods will be added to the decsendant
```
cromoteca, line 25:
```suggestion
 * {@link Endpoint} but inheritable methods will be added to the descendant
```

Artur-:
This would still need some changes in the generator code

cromoteca:
I was thinking of this:

`@BrowserCallable` === `@EndpointExposed`
`@Service @BrowserCallable` === `@Endpoint`

That would allow to use `@BrowserCallable` even at method level.

Artur-:
That would be really unexpected if methods in `@BrowserCallable` classes would not be callable from the browser

tarekoraby:
> @Service @BrowserCallable === @Endpoint

The typical use case of Hilla is to expose the whole class, not a method. Let's not make the typical use case extra cumbersome by requiring two annotations. Martin hates boilerplate code.

packages/java/endpoint/src/main/java/dev/hilla/BrowserCallable.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright 2000-2022 Vaadin Ltd.
```
taefi, line 2:
```suggestion
 * Copyright 2000-2023 Vaadin Ltd.
```

taefi (changes requested):

packages/java/endpoint/src/main/java/dev/hilla/BrowserCallable.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.Target;
+
+import org.springframework.stereotype.Component;
+
+/**
+ * Makes the methods of the annotated class available to the browser.
+ * <p>
+ * For each class, a corresponding TypeScript class is generated in
+ * {@code frontend/generated} with TypeScript methods for invoking the methods
+ * in this class.
```
taefi, line 30: Shouldn't it be mentioned that this is an alias for the existing `@Endpoint` ? Otherwise, people are confused what's the difference...

Artur-: Added cross references

packages/java/endpoint/src/main/java/dev/hilla/EndpointExposed.java
```diff
@@ -22,8 +22,8 @@
 
 /**
  * Annotation marks the class itself should not be treated as an
- * {@link Endpoint}. Instead the inheritable methods will be added to the
- * decsendant Endpoints.
+ * {@link Endpoint} but inheritable methods will be added to the descendant
```
taefi, line 25:
```suggestion
 * {@link Endpoint}, but inheritable methods will be added to the descendant
```

packages/java/endpoint/src/main/java/dev/hilla/EndpointUtil.java
```diff
@@ -31,7 +31,7 @@
 import org.springframework.web.util.pattern.PathPatternParser;
 
 /**
- * A util class related to {@link Endpoint}.
+ * A util class related to published classes.
```
taefi, line 34: Not sure if `published classes` gives any concrete message about what are we talking about. 
Maybe for now we could say:
```
A utility class related to {@link Endpoint} and {@link BrowserCallable}.
```
or 
```
A utility class related to browser-callable services.
```

Artur-: Reworded

packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/ScanResult.java
```diff
@@ -16,9 +18,10 @@ public io.github.classgraph.ScanResult get() {
...
 
-    public List<? extends Class<?>> getClassesWithAnnotation(
-            String annotation) {
-        return result.getClassesWithAnnotation(annotation).stream()
-                .map(ClassInfo::loadClass).collect(Collectors.toList());
+    public List<Class<?>> getClassesWithAnnotation(String... annotations) {
+        Stream<ClassInfo> c = Stream.of(annotations)
+                .flatMap(annotation -> result
+                        .getClassesWithAnnotation(annotation).stream());
+        return c.map(ClassInfo::loadClass).collect(Collectors.toList());
```
taefi, line 25:
```suggestion
        return c.map(ClassInfo::loadClass).toList();
```

packages/java/parser-jvm-plugin-backbone/src/main/java/dev/hilla/parser/plugins/backbone/EndpointPlugin.java
```diff
@@ -67,8 +68,10 @@ public NodeDependencies scan(@Nonnull NodeDependencies nodeDependencies) {
             var endpointAnnotationName = getStorage().getParserConfig()
                     .getEndpointAnnotationName();
             var endpoints = rootNode.getSource()
-                    .getClassesWithAnnotation(endpointAnnotationName).stream()
-                    .map(ClassInfoModel::of).collect(Collectors.toList());
+                    .getClassesWithAnnotation(endpointAnnotationName,
+                            "dev.hilla.BrowserCallable")
+                    .stream().map(ClassInfoModel::of)
+                    .collect(Collectors.toList());
```
taefi, line 74:
```suggestion
                    .toList();
```

platosha (dismissed):

taefi (dismissed):

Artur-:
Still some formatting..

taefi (approved):

taefi:
On the matter adding an alias for or renaming `@EndpointExposed` (which comes later), I'd suggest to name it `@BrowserCallableSuperclass`. I might seem to be a little long at first sight, but as it is somewhat similar to `@MappedSuperclass` of JPA, I think it would be easier to relate the ideas of and conclude that it is not a browser-callable on its own, but the public methods are available through the inheriting classes.

PR #1222, opened 2023-09-08
chore: Fix build

Artur-:
If a newer antrun plugin is available, the build fails with "You are using 'tasks' which has been removed from the maven-antrun-plugin. Please use 'target' and refer to the >>Major Version Upgrade to version 3.0.0<< on the plugin site"

<!-- PLEASE READ AND FOLLO


ZheSun88:
let us remove those plugins, it is not necessary, as there is no {{version}} in the generated `versions.json`.

ZheSun88 (approved):

PR #1234, opened 2023-09-11
feat: add react-routing helpers

sissbruecker:
## Description

Extracts routing types and hooks from the starter into a new `@hilla/react-routing` package. The main features are:
- A `RouteObjectWithMetadata` type that extends the `react-router-dom` route type with metadata for menu icon and title, as well as for whether the route requires authentication and which roles are allowed to access the route. Specifically the type overrides the `handle` property, which should be used for custom application-specific data, with metadata currently used by the starter.
- A `useRouteMetadata` hook that returns the metadata for the current route. The hook will return the metadata for the current route if it has metadata, or the nearest parent route that has metadata. Otherwise returns `undefined`.

Changes compared to the starter:
- This duplicates the `AccessProps` type as there currently is none in the Hilla codebase that could be imported. Not a blocker IMO and can be cleaned up when auth-related helpers are extracted.
- The role type has been changed to use a `string`, whereas the starter uses an application-specific `Role` enum.

## Type of change

- Feature

### DIFF
--- a/packages/ts/react-routing/src/index.tsx
+++ b/packages/ts/react-routing/src/index.tsx
@@ -10,21 +10,23 @@ export type AccessProps = Readonly<{
   rolesAllowed?: readonly string[];
 }>;
 
-export type ViewMeta = Readonly<{ handle?: AccessProps & MenuProps }>;
+export type RouteMetadata = AccessProps & MenuProps;
+
+type HandleWithMetadata = { handle?: RouteMetadata };
 
 type Override<T, E> = E & Omit<T, keyof E>;
 
-export type IndexHillaRouteObject = Override<IndexRouteObject, ViewMeta>;
-export type NonIndexHillaRouteObject = Override<
-  Override<NonIndexRouteObject, ViewMeta>,
+export type IndexRouteObjectWithMetadata = Override<IndexRouteObject, HandleWithMetadata>;
+export type NonIndexRouteObjectWithMetadata = Override<
+  Override<NonIndexRouteObject, HandleWithMetadata>,
   {
-    children?: HillaRouteObject[];
+    children?: RouteObjectWithMetadata[];
   }
 >;
-export type HillaRouteObject = IndexHillaRouteObject | NonIndexHillaRouteObject;
+export type RouteObjectWithMetadata = IndexRouteObjectWithMetadata | NonIndexRouteObjectWithMetadata;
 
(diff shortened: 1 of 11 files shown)


packages/ts/react-routing/test/useRouteMetadata.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+  icon: 'nested',
+  requireAuthentication: true,
+  rolesAllowed: ['ROLE_NESTED'],
+};
+
+const routes: RouteObjectWithMetadata[] = [
+  {
+    path: '/main',
+    element: <CaptureMetadata />,
+    handle: mainMetadata,
```
Artur-, line 30: Why is it metadata is specified as `handle` and not e.g. `metadata`?

packages/ts/react-routing/src/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+type Override<T, E> = E & Omit<T, keyof E>;
+
+export type IndexRouteObjectWithMetadata = Override<IndexRouteObject, HandleWithMetadata>;
+export type NonIndexRouteObjectWithMetadata = Override<
+  Override<NonIndexRouteObject, HandleWithMetadata>,
+  {
+    children?: RouteObjectWithMetadata[];
+  }
+>;
+export type RouteObjectWithMetadata = IndexRouteObjectWithMetadata | NonIndexRouteObjectWithMetadata;
```
sissbruecker, line 26: One downside with this type is that developers would not be able to add any other custom property to `handle` anymore. A future improvement could make it generic, otherwise I'd suggest to keep it simple for now.

packages/ts/react-routing/test/useRouteMetadata.spec.tsx
sissbruecker: The types and hook are just a glorified wrapper around the ones from `react-router-dom`. The `useMatches` hook used internally does not return the full route object but only some properties of it, one of which is `handle`. So if we add some custom property to the route itself we would not be able to retrieve it that way, but would have to come up with our own approach of accessing all routes (no default way to do that in `react-router-dom`), and then identifying the current one.

sissbruecker:
Marking this as a draft for now, as we want to look into changing the starter to just use basic react-router API without adding too much complexity in terms of helpers.

sissbruecker:
Closing as we decided not to add any helpers at all for now.

Issue #703, opened 2022-12-28
Page response type ignored


simasch:
I have this endpoint method:

```java
@Nonnull
public Page<@Nonnull Person> list(String filter, Pageable pageable) {
```

This generates:

```typescript
function _list(
 filter: string | undefined,
 pageable: Pageable | undefined,
 __init?: EndpointRequestInit
): Promise<Array<Person>>
{
 return client.call('PersonEndpoint', 'list', {filter, pageable}, __init);
}
```

As you can see the return type Page is ignored.

Why is that? If I would like the result as Page I cannot do that and have to create my own result type .

platosha:
It was introduced in https://github.com/vaadin/flow/pull/11648 in order to support returning `Page` from endpoints somehow. The idea back then was that a simple array result is enough.

The old generator was struggling to emit the `Page` directly, as it is coming from a dependency, which source is not available, and reflection support was lacking.

Since this is unexpected, we could improve this for the new generator, and the old one could use a better (more similar to Page) internal representation as a workaround.

Until this is fixed, please use a DTO as a workaround.

platosha:
Hm-m, actually it would be preferable to fix this issue in a way that keeps the same behavior in the old and new generator. I’m marking this as a blocker, so that it is fixed before the 2.0 major release.

platosha:
After one more take on this, it turns out that generic entities such as Page were not supported. Fixing this issue is blocked by https://github.com/vaadin/hilla/issues/579

Sadly it is not possible to fix in the 2.0 beta at the moment. Let us revisit this after adding support for generic entities.

thiagonuic:
+1 for fixing this.

platosha:
After a discussion with the team, we figured that a good way of supporting `Page` would apply to all iterative types, as at the moment we treat `Page<T>` as an instance of `Iterable<T>` and that's why it is converted to array.

We need:
- Support for generating types extending Array
- Client runtime mapping between array with additional properties and JSON representation

platosha:
Let's simplify the implementation to the following idea: instead of trying to apply it on every iterable class, we introduce a transfer type DTO for `Page` specifically, which is an object with items array property and other metadata properties from `Page`.

On the TypeScript, let us try to achieve an object extending array with additional metadata properties. In the model generator, it would be enough to just map it to `ArrayModel`. If that idea would not seem achievable, we could accept the breaking change and make a simple non-iterable object with items instead.

PR #1251, opened 2023-09-11
feat: Add list support for auto grid

Artur-:


packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -0,0 +1,97 @@
...
+    let size;
+    if (items.length === pageSize) {
+      size = (pageNumber + 1) * pageSize + 1;
+      if (size < (grid as any)._cache.size) {
+        // Only allow size to grow here to avoid shrinking the size when scrolled down and sorting
+        size = undefined;
+      }
+    } else {
+      size = pageNumber * pageSize + items.length;
+    }
```
sissbruecker, line 59: We could also return the size from the crud service and then just use it here. Related to that, the current signature of `CrudService` would prevent use from ever doing so as it only returns a list.

How about something like this:
```java
public class PageResult<T> {
    private final List<T> content;
    private final long totalElements;

    public PageResult(Page<T> page) {
        this.content = page.getContent();
        this.totalElements = page.getTotalElements();
    }
}

public interface CrudService<T> {
    PageResult<T> list(Pageable pageable);
}

@EndpointExposed
public class CrudRepositoryService<T, ID> implements CrudService<T> {
    ...
    @Override
    public PageResult<T> list(Pageable pageable) {
        return new PageResult(repository.findAll(pageable));
    }
}
```

packages/ts/react-grid/test/autogrid.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+    it('sets a data provider', async () => {
+      const result = render(<TestAutoGrid />);
+      const grid = result.container.querySelector('vaadin-grid');
+      expect(grid?.dataProvider).to.not.be.undefined;
+    });
+    it('data provider provides data', async () => {
+      const result = render(<TestAutoGrid />);
+      const grid: GridElement = result.container.querySelector('vaadin-grid')!;
+      grid.requestContentUpdate();
+      await sleep(1);
```
sissbruecker, line 59: Interesting that the `flushGrid` helper doesn't work here. Still waiting for the next frame seems to be enough, which might be better than arbitrary timeouts. The `requestContentUpdate` doesn't make a difference and can be removed.

sissbruecker (commented):
All of the TS files light up like a christmas tree for me with ESLint, so we should do something about those.

packages/ts/react-grid/src/types/Order.ts
```diff
@@ -0,0 +1,9 @@
+import type Direction_1 from './Direction.js';
```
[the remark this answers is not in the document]
Artur-: All these files are copied from a project, where they are auto generated from the Java files. Not sure at all how we should handle them but probably not make manual changes to them if they are auto generated + copied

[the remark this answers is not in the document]
Artur-: But this file particular file seems to be in the wrong folder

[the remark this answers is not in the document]
Artur-: Ah sorry, they are moved in the next PR I think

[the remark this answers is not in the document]
sissbruecker: Alright.

packages/ts/react-grid/src/types/Direction.ts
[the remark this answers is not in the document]
sissbruecker: Obsolete if the files are auto-generated.

packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+// import Sort from "Frontend/generated/dev/hilla/mappedtypes/Sort";
+// import Direction from "Frontend/generated/org/springframework/data/domain/Sort/Direction";
+type Sort = any;
+enum Direction {
+  ASC = 'ASC',
+  DESC = 'DESC',
+}
+
+export type _AutoGridProps<TItem> = {
```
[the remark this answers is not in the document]
Artur-: This is also fixed in the next PR, let me move the fix here

packages/ts/react-grid/src/autogrid.tsx
Artur-: Yes this is done intentionally for now so you do not have to worry about the performance hit of making an extra query and the more complex return type. If you return `PageResult` then the frontend code gets more complex. What is in autogrid does not matter but the crud services will be used by people outside autogrid also. 

The `Page` class in Spring Boot actually can contain the total elements so it is kind of your `PageResult` already. Hilla just ignores that #703

packages/ts/react-grid/test/autogrid.spec.tsx
Artur-: This is also fixed in the next PR, let me move the fix here

Artur-:
Rebased without changes first

packages/ts/react-grid/src/autogrid.tsx
sissbruecker: What would be your long-term plan for this? A second method for counting that you either implement or not? IMO this hurts usability as you can not scroll to the end of the grid, you have to go through all pages first.

packages/ts/react-grid/test/autogrid.spec.tsx
Artur-: Removed `requestContentUpdate`

packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -0,0 +1,97 @@
...
+  service: CrudService<TItem>;
+  model: ModelConstructor<TItem, any>;
```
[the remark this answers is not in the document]
Artur-: Fixed

packages/ts/react-grid/src/autogrid.tsx
Artur-: Should we leave this as an additional feature, to optionally include the size? The current one only implements infinite scrolling

packages/ts/react-grid/test/autogrid.spec.tsx
Artur-: I tried adding the `nextRender` helper from https://github.com/vaadin/testing-helpers/blob/fb02eb62812de679e1534967ada8dbfd90795a95/src/utils.ts#L67 but depending on `@vaadin/testin-helpers` fail the tests with
```
10:16:45 AM [vite] Internal server error: Failed to resolve import "/__web-dev-server__web-socket.js" from ".vite/deps/@vaadin_testing-helpers.js?v=0c03f7fe". Does the file exist?
```

and copying the helper into the project causes
```
  Uncaught ReferenceError: JSCompiler_renameProperty is not defined
```
because of the Polymer import

Artur-: Using `flushGrid` now, it seems to work

packages/ts/react-grid/src/autogrid.tsx
Artur-: Whether not being able to scroll directly to the end or not is a problem is quite use case dependent. Some people search instead of scroll, other people don't. We must have another method for the case as it has different requirements

packages/java/endpoint/src/main/java/dev/hilla/crud/CrudService.java
```diff
@@ -0,0 +1,24 @@
+package dev.hilla.crud;
+
+import java.util.List;
+
+import org.springframework.data.domain.Pageable;
+
+/**
+ * A browser service that can create, read and update a given type of object.
```
taefi, line 8: Something called "browser service" is communicating something else IMO. I suggest to change it to: 
```suggestion
 * A browser-callable service that can create, read and update a given type of object.
```

taefi (changes requested):

packages/java/endpoint/src/main/java/dev/hilla/crud/CrudRepositoryService.java
```diff
@@ -0,0 +1,32 @@
+package dev.hilla.crud;
+
+import java.util.List;
+
+import dev.hilla.EndpointExposed;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.jpa.repository.JpaRepository;
+
+/**
+ * A browser service that delegates crud operations to a JPA repository.
```
taefi, line 10:
```suggestion
 * A browser-callable service that delegates crud operations to a JPA repository.
```

packages/ts/react-grid/src/autogrid.tsx
sissbruecker: Let's cover this later with a separate API then.

packages/ts/react-grid/test/autogrid.spec.tsx
sissbruecker: What fixes it is the `await`, even though `flushGrid` doesn't return a promise. If you remove the `await` the test fails. I would suggest to wait for an animation frame instead to make this more explicit.

Artur-: which helper would that be and from where?

sissbruecker: We use a `nextFrame` helper in WCs, which comes from `node_modules/@vaadin/testing-helpers/dist/utils.js`

It's basically this:
```js
/**
 * Resolves after requestAnimationFrame.
 *
 * @return {Promise<void>} Promise resolved after requestAnimationFrame
 */
export async function nextFrame() {
    return new Promise((resolve) => {
        requestAnimationFrame(() => {
            resolve();
        });
    });
}
```

Artur-: Updated

taefi (dismissed):

sissbruecker (approved):

PR #1252, opened 2023-09-11
refactor(form): improve form DX

Lodin:
This PR improves the overall DX of `@hilla/form` along with fixing some bugs.

The most noticeable changes are: 

1. Improve TS support in all form components:
  - Reduce amount of `as` casts.
  - Reduce amount of `any`. According to TS recommendations, `any` was replaced preferably with `unknown`, in some cases with `never`.
  - `tsc` now can infer more types without the user support.
2. Models syntax is now more user-friendly.
  - Child types are created using a callback instead of a separate constructor + array of parameters.
  - Less magic in `createEmptyValue` method: no static inheritance, each class has its own method.
  - Due to these changes, the TS support became way better. 
3. Updates for internal methods implementation to reduce cognitive load.
  - Multiple validation methods are now generators instead of arrays.
  - Now private, protected and public members are separated at the JS level:
    - Public members are just regular string methods.
    - Protected members are symbol methods.
    - Private members are new ECMA private members (with `#`).
4. For tests, the usage of `Sinon` spies is increased. They replace multiple manual implementations.

I apologize for the PR size. Unfortunately, it'd be hard to make small independent PRs for these changes since they are pulling one another.

There are going to be a couple of follow-on PRs that should be merged into this one to make all tests pass.

1. The first one will address the model generation. Since the structure has changed, it requires an update for the generator as well.
2. The second one will update `react-form` package to reflect changed made.


packages/ts/form/src/Binder.ts
```diff
@@ -10,7 +12,7 @@ import type { AbstractModel, ModelConstructor } from './Models.js';
  * @typeParam T - Type of the value that binds to a form
  * @typeParam M - Type of the model that describes the structure of the value
  */
-export class Binder<T, M extends AbstractModel<T>> extends BinderRoot<T, M> {
+export class Binder<M extends AbstractModel> extends BinderRoot<M> {
```
sissbruecker, line 15: This looks like a breaking change. Assuming that I have some code like this in my application:
```ts
private doSomethingWithBinder(binder: Binder<Person, PersonModel>) { ... }
```
It will not compile anymore.

sissbruecker (commented):
Can you please elaborate, maybe with some small code examples, how this improves DX for Hilla users, specifically for React developers? It mostly looks like it just changes how types internally are set up, which doesn't seem very valuable to deal with at the moment.

Lodin:
@sissbruecker, the problem is that there were several times I faced that issue during the development of `react-form`. It happens when you have the current implementation and try to use `unknown` instead of `any` everywhere (which is a TS recommended practice).

For example, the following code won't compile:
```ts
export function useFormPart<M extends AbstractModel<unknown>>(
  model: M,
): UseFormPartResult<ReturnType<M['valueOf']>, M> {
  const binderNode = getBinderNode(model);
  const field = useFields(binderNode);
  return {
    ...getFormPart(binderNode),
    field,
  };
}
```
It raises the following error:
```
TS2344: Type 'M' does not satisfy the constraint 'AbstractModel<ReturnType<M["valueOf"]>>'.
Type 'AbstractModel<unknown>' is not assignable to type 'AbstractModel<ReturnType<M["valueOf"]>>'.
The types returned by 'valueOf()' are incompatible between these types.
Type 'unknown' is not assignable to type 'ReturnType<M["valueOf"]>'.
```
So, you are forced to use `any` everywhere which reduces the type safety and could cause Eslint errors.

Then, about the same `useFormPart` hook. It has a different signature than `useForm`.
```ts
function useFormPart<M extends AbstractModel<any>>(model: M): UseFormPartResult<ReturnType<M['valueOf']>, M>;

useForm<T, M extends AbstractModel<T>>(
  Model: ModelConstructor<T, M>,
  config?: BinderConfiguration<T>,
): UseFormResult<T, M>;
```
When I was working on `react-form`, that part confused me. So I can imagine that it could easily confuse the developer who starts to work with the `@hilla/form` even though it could be an edge use case.

Third: this PR also reworks and simplifies the model implementation. Now you can create model by hands without additional cognitive load because everything is done via callbacks instead of constructors + array of their arguments.

Also, using only the specific model as an argument makes API much clearer and more responsive. And the internal cleanup increases the stability of the library.

platosha (dismissed):

packages/ts/form/src/Models.ts
```diff
@@ -27,29 +23,32 @@ export function hasFromString<T>(model: AbstractModel<T>): model is AbstractMode
 
 export type Value<M> = M extends AbstractModel<infer T> ? T : never;
 
+export interface HasValue<T> {
+  value?: T;
+}
+
+export type ModelParent<T> = AbstractModel | HasValue<T>;
+
+export type ModelConstructor<M> = new (parent: { value: undefined }, key: 'value', optional: boolean) => M;
```
Lodin, line 32: It looks like an `EmptyModelConstructor`, not just a `ModelConstructor`. What about changing the name?

Lodin: Also, I consider `value` is a misleading name. I spent quite a time to understand that it is about the empty parent, not the real model. Let's have more special name here, like `$value$` I introduced.

platosha: Let's address the `value` name confusion later.

platosha (dismissed):

packages/ts/form/src/Models.ts
```diff
@@ -34,57 +21,38 @@ export function hasFromString<T>(model: AbstractModel<T>): model is AbstractMode
   return _fromString in model;
 }
 
-export interface HasValue<T> {
-  value?: T;
-}
+export type Value<M> = M extends AbstractModel<infer T> ? T : never;
+
+export const _modelDetachedParent = { $value: undefined };
```
Lodin, line 26: Let's not use underscore for anything except Symbols.

```suggestion
export const modelDetachedParent = { $value$: undefined };
```

packages/ts/form/src/Models.ts
```diff
@@ -34,57 +21,38 @@ export function hasFromString<T>(model: AbstractModel<T>): model is AbstractMode
...
+
+export type ModelParent = AbstractModel | BinderNode | typeof _modelDetachedParent;
 
-export type ModelParent<T> = AbstractModel<any> | HasValue<T>;
-export type ModelValue<M extends AbstractModel<unknown>> = ReturnType<M['valueOf']>;
+export type EmptyModelConstructor<M> = new (
```
Lodin, line 30: If everything else is named as `detached model`, let's be consistent here as well.
```suggestion
export type DetachedModelConstructor<M> = new (
```

packages/ts/form/src/Models.ts
```diff
@@ -34,57 +21,38 @@ export function hasFromString<T>(model: AbstractModel<T>): model is AbstractMode
...
+  parent: typeof _modelDetachedParent,
+  key: '$value$',
+  optional: boolean,
+) => M;
 
-export interface ModelConstructor<T, M extends AbstractModel<T>> {
-  new (parent: ModelParent<T>, key: keyof any, optional: boolean, ...args: any[]): M;
-  createEmptyValue(): T;
+export function _createDetachedModel<M extends AbstractModel>(type: EmptyModelConstructor<M>): M {
```
Lodin, line 36: Let's not use underscore for anything except Symbols.
```suggestion
export function createDetachedModel<M extends AbstractModel>(type: EmptyModelConstructor<M>): M {
```

Lodin (commented):
There one error and several small style fixes

sissbruecker (dismissed):

sissbruecker (dismissed):

sissbruecker (dismissed):

sissbruecker (approved):

PR #1257, opened 2023-09-12
refactor(react-form): update code according to #1252

Lodin:
Supplement PR for #1252


Lodin:
Unfortunately, tests are not going to pass because they require #1256 to be merged as well.

Issue #438, opened 2022-06-08
Support polymorphic return types through @JsonSubTypes


Artur-:
When sending different types of events to the client, I want to be able to use one Flux endpoint method for all different types of events as they are happening in a logical order that I want to preserve. On the client side I want to be able to differentiate between the event types and handle them separately.

**Describe the solution you'd like**

Ideally I would want to define my endpoint method to return an abstract type, e.g.
```
public Flux<AbstractEvent> subscribe() { ... }
```
and nothing else. This does not seem realistic as Hilla then would need to find out all possible subtypes of `AbstractEvent`. While it can work for a limited case it would not work very well if returning `Flux<Object>`. Also with only this there is no information available on the client side about which type of event was actually sent, unless the event instance manually include that data.

Jackson supports `@JsonSubTypes` which should help here for both of the problems. If I annotate my `AbstractEvent` as
```
    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY)
    @JsonSubTypes({
            @JsonSubTypes.Type(value = AddEvent.class, name = "add"),
            @JsonSubTypes.Type(value = UpdateEvent.class, name = "update"),
            @JsonSubTypes.Type(value = DeleteEvent.class, name = "delete") })
```
then 
1. Hilla should generate TS for all the mentioned sub types: `AddEvent`, `UpdateEvent`, `DeleteEvent`
2. Jackson will add the `@type` property (because of use=`NAME`) when serializing so you can distinguish the event types on the client side
3. Hilla should include the `@type` property in the generated TS to make it easy to access

samie:
Do you have TS example? Or how would the TS usage look like for `AbstractEvent`?

cromoteca:
Before thinking about an implementation, some investigation about serialization in reactive endpoints is needed. Consider this endpoint:

```java
@Endpoint
@AnonymousAllowed
public class SubtypesEndpoint {
  public Flux<@Nonnull AbstractEvent> subscribe() {
    return Flux.just(new AddEvent(), new UpdateEvent(), new DeleteEvent());
  }

  public @Nonnull AbstractEvent single() {
    return new AddEvent();
  }

  @JsonTypeInfo(use = JsonTypeInfo.Id.MINIMAL_CLASS)
  public static class AbstractEvent {
    public String name;
  }

  public static class AddEvent extends AbstractEvent {
    public String item;
  }

  public static class UpdateEvent extends AbstractEvent {
[...]

PR #1259, opened 2023-09-12
feat: Filtering for auto grid

Artur-:


packages/java/endpoint/src/main/java/dev/hilla/crud/CrudRepositoryService.java
```diff
@@ -13,21 +18,34 @@
...
+    protected JpaSpecificationExecutor<T> getRepository() {
+        return repository;
     }
 
     @Override
-    public List<T> list(Pageable pageable) {
-        return repository.findAll(pageable).getContent();
+    public List<T> list(Pageable pageable, @Nullable Filter filter) {
+        Specification<T> spec = jpaFilterConverter.toSpec(filter, entityClass);
+        return ((JpaSpecificationExecutor<T>) repository)
```
sissbruecker, line 46: Cast seems to be unnecessary.

packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
```diff
@@ -0,0 +1,64 @@
...
+        this.filter = filter;
+        this.entity = entity;
+        this.javaType = javaType;
+    }
+
+    @Override
+    public Predicate toPredicate(Root<T> root, CriteriaQuery<?> query,
+            CriteriaBuilder criteriaBuilder) {
+        String value = filter.getFilterValue();
+        Path<String> propertyPath = root.get(filter.getPropertyId());
```
sissbruecker, line 36: Looks like this will throw should be path / property not exist. Intended, or should we continue and log a warning?

packages/java/endpoint/src/main/java/dev/hilla/crud/filter/Filter.java
```diff
@@ -0,0 +1,13 @@
+package dev.hilla.crud.filter;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonSubTypes.Type;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "t")
+@JsonSubTypes({ @Type(value = OrFilter.class, name = "o"),
+        @Type(value = AndFilter.class, name = "a"),
+        @Type(value = PropertyStringFilter.class, name = "p") })
```
sissbruecker, line 10: As developers can construct and pass their own filters I would suggest to use more readable names here: `type`, `or`, `and`, `propertyString`.

packages/java/tests/spring/react-grid-test/frontend/views/ReadOnlyGridOrFilter.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const [filter, setFilter] = useState<Filter | undefined>(undefined);
+
+  return (
+    <div>
+      <TextField
+        id="filter"
+        style={{ width: '20em' }}
+        label="Search for first or last name"
+        onValueChanged={(e) => {
+          const firstNameFilter: any = {
```
sissbruecker, line 18: Can we type the filters here?

packages/java/endpoint/src/main/java/dev/hilla/crud/filter/OrFilter.java
```diff
@@ -0,0 +1,19 @@
+package dev.hilla.crud.filter;
+
+import java.util.List;
+
+/**
+ * A filter that requires at least on of its children to pass.
```
sissbruecker, line 6:
```suggestion
 * A filter that requires at least one of its children to pass.
```

packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -77,17 +83,23 @@ function createColumns(model: ModelConstructor<unknown, AbstractModel<unknown>>)
...
-      children={children}
-    ></Grid>
-  );
+  const ref = useRef<GridElement<TItem>>(null);
+  const dataProviderFilter = useRef<Filter | undefined>(undefined);
+
+  useEffect(() => {
+    const grid = ref.current!;
+    grid.dataProvider = createDataProvider(grid, service, dataProviderFilter);
+  });
```
sissbruecker, line 96: This causes the data provider to be created and assigned with every render. You need to add an empty dependencies array if you only want to run this only once.

packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -77,17 +83,23 @@ function createColumns(model: ModelConstructor<unknown, AbstractModel<unknown>>)
...
+
+  useEffect(() => {
```
sissbruecker, line 98: As this file will probably pile up with `useEffect` calls, I would suggest to add a small comment over each on what they do and when they are run.

packages/java/endpoint/src/main/java/dev/hilla/crud/filter/Filter.java
```diff
@@ -0,0 +1,13 @@
...
+public interface Filter {
```
sissbruecker, line 11: I would consider adding a `type` property to `Filter`, and then using an `Enum` for the values, if only so that we get the enum generated on the client and developers don't have to use strings and can actually use the `Filter` type and sub-types when defining filters. I've seen https://github.com/vaadin/hilla/issues/438, which would be a better solution, but this seems like an easy improvement in the meantime.

packages/java/tests/spring/react-grid-test/src/test/java/dev/hilla/test/reactgrid/ReadOnlyGridOrFilterIT.java
```diff
@@ -0,0 +1,33 @@
+package dev.hilla.test.reactgrid;
+
+import org.junit.Test;
+
+import com.vaadin.flow.component.textfield.testbench.TextFieldElement;
+
+public class ReadOnlyGridOrFilterIT extends AbstractGridTest {
```
sissbruecker, line 7: ITs currently don't cover the and filter and the equals matcher. Would be nice to add them, or alternatively write unit tests for the spec conversion.

packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
Artur-: added a test

packages/java/tests/spring/react-grid-test/src/test/java/dev/hilla/test/reactgrid/ReadOnlyGridOrFilterIT.java
Artur-: Added unit tests

packages/ts/react-grid/src/autogrid.tsx
Artur-: Fixed

packages/ts/react-grid/src/autogrid.tsx
Artur-: Done

packages/java/endpoint/src/main/java/dev/hilla/crud/filter/Filter.java
sissbruecker: As discussed let's postpone this for now, to either improve the generator to support `@JsonSubTypes` or create a custom TS API.

sissbruecker (approved):

sissbruecker (approved):

PR #1275, opened 2023-09-14
chore: ignore same files as eslint when prettifying

cromoteca:
Git commit hook prettifies files that are ignored by ESLint in `react-grid`. Let's create a similar config for Prettier.


cromoteca:
Thanks @sissbruecker

sissbruecker (approved):

Issue #1273, opened 2023-09-14
[autogrid] option to customize which columns to show and in which order


Artur-:
It should be possible to define which columns are shown and in which order, e.g. as
`<AutoGrid visibleColumns={"firstName","lastName","foo","bar"}>`

This would generate columns only for the given properties in the given order

PR #1279, opened 2023-09-14
feat: customize visible columns in auto grid

sissbruecker:
## Description

Allows to customize the visible columns and their order in auto grid.

Closes https://github.com/vaadin/hilla/issues/1273

## Type of change

- Feature

### DIFF
--- a/packages/ts/react-grid/src/autogrid.tsx
+++ b/packages/ts/react-grid/src/autogrid.tsx
@@ -90,7 +90,13 @@ function createColumns(model: ModelConstructor<unknown, AbstractModel<unknown>>,
   ));
 }
 
-export function AutoGrid<TItem>({ service, model, filter, visibleColumns, ...gridProps }: AutoGridProps<TItem>): JSX.Element {
+export function AutoGrid<TItem>({
+  service,
+  model,
+  filter,
+  visibleColumns,
+  ...gridProps
+}: AutoGridProps<TItem>): JSX.Element {
   // This cast should go away with #1252
   const children = createColumns(model as ModelConstructor<unknown, AbstractModel<unknown>>, visibleColumns);
 
@@ -105,9 +111,11 @@ export function AutoGrid<TItem>({ service, model, filter, visibleColumns, ...gri
 
   useEffect(() => {
     // Update the filtering, whenever the filter changes
-    const grid = ref.current!;
-    dataProviderFilter.current = filter;
-    grid.clearCache();
+    const grid = ref.current;
+    if (grid) {
+      dataProviderFilter.current = filter;
+      grid.clearCache();
+    }
   }, [filter]);
 
   return <Grid {...gridProps} ref={ref} children={children}></Grid>;
--- a/packages/ts/react-grid/src/autogrid.tsx
+++ b/packages/ts/react-grid/src/autogrid.tsx
@@ -14,13 +14,14 @@ import type { CrudService } from './crud';
 import type Filter from './types/dev/hilla/crud/filter/Filter';
 import type Sort from './types/dev/hilla/mappedtypes/Sort';
 import Direction from './types/org/springframework/data/domain/Sort/Direction';
-import { getProperties } from './utils.js';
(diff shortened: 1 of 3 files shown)


packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -71,21 +72,29 @@ function createDataProvider<TItem>(grid: GridElement<TItem>, service: CrudServic
...
+  const children = createColumns(model as ModelConstructor<unknown, AbstractModel<unknown>>, visibleColumns);
 
   return (
     <Grid
       {...gridProps}
       ref={(grid) => {
-        grid!.dataProvider = createDataProvider(grid!, service);
+        if (grid) {
+          grid.dataProvider = createDataProvider(grid, service);
+        }
```
sissbruecker, line 97: This resulted in a null reference during test teardown 🤷‍♂️

packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -98,9 +111,11 @@ export function AutoGrid<TItem>({ service, model, filter, ...gridProps }: AutoGr
...
   useEffect(() => {
     // Update the filtering, whenever the filter changes
-    const grid = ref.current!;
-    dataProviderFilter.current = filter;
-    grid.clearCache();
+    const grid = ref.current;
+    if (grid) {
+      dataProviderFilter.current = filter;
+      grid.clearCache();
+    }
```
sissbruecker, line 118: All tests fail for me without this check

```
ERROR LOG: 'Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.'

  @hilla/react-grid
    useAutoGrid
      ✗ creates columns based on model
        Error: Uncaught TypeError: Cannot read properties of null (reading 'useRef') (http://localhost:9877/packages/ts/react-grid/.vite/deps/chunk-U33FUF2H.js?v=c646efa1:1074)
            at Object.invokeGuardedCallbackDev (packages/ts/react-grid/.vite/deps/chunk-HMFLAOO4.js?v=a669691a:3699:24)
            at invokeGuardedCallback (packages/ts/react-grid/.vite/deps/chunk-HMFLAOO4.js?v=a669691a:3733:39)
            at beginWork$1 (packages/ts/react-grid/.vite/deps/chunk-HMFLAOO4.js?v=a669691a:19761:15)
            at performUnitOfWork (packages/ts/react-grid/.vite/deps/chunk-HMFLAOO4.js?v=a669691a:19197:20)
            at workLoopSync (packages/ts/react-grid/.vite/deps/chunk-HMFLAOO4.js?v=a669691a:19133:13)
            at renderRootSync (packages/ts/react-grid/.vite/deps/chunk-HMFLAOO4.js?v=a669691a:19112:15)
            at performConcurrentWorkOnRoot (packages/ts/react-grid/.vite/deps/chunk-HMFLAOO4.js?v=a669691a:18674:83)
            at flushActQueue (packages/ts/react-grid/.vite/deps/chunk-DFDKWVIJ.js?v=a669691a:1812:32)
            at act (packages/ts/react-grid/.vite/deps/chunk-DFDKWVIJ.js?v=a669691a:1747:19)
```
Hook calls in the component look fine though 🤔

Artur-: Not sure how you run the tests because after removing the if, I get
```
  @hilla/react-grid
    useAutoGrid
      ✓ creates columns based on model
      ✓ sets a data provider, but only once
      ✓ data provider provides data
      ✓ does not pass its own parameters to the underlying grid
      ✓ passes filter to the data provider
      customize columns
        ✓ should only show configured columns in specified order
        ✓ should ignore unknown columns
```

Artur- (approved):
Otherwise 👍

PR #1289, opened 2023-09-15
fix(react-form): enable eslint checks

Lodin:


packages/ts/react-form/src/index.ts
```diff
@@ -5,31 +5,32 @@ import {
...
 
-const $wnd = window as any;
+const $wnd = window as VaadinWindow;
 
 $wnd.Vaadin ??= {};
 $wnd.Vaadin.registrations ??= [];
 $wnd.Vaadin.registrations.push({
   is: '@hilla/react-form',
-  version: /* updated-by-script */ '2.2.0-beta1',
+  version: __VERSION__,
```
Artur-, line 29: Is the script still able to update the value? Where did the default value go?

Lodin: The `__VERSION__` is replaced by esbuild to the version from `package.json` during the build stage. We already use this idea in `@hilla/form`. It allows us to maintain the version in a single place and avoid unnecessary version-update commits.

Artur-: Good then

Artur- (approved):

PR #1297, opened 2023-09-18
test: Remove extra junit version definition

Artur-:


packages/java/hilla-react/pom.xml
```diff
@@ -42,7 +42,7 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <version>4.13.1</version>
+            <version>4.13.2</version>
```
sissbruecker, line 45: As the dependency is already in the root pom, can the version be removed here? That's how other nested modules are configued. 

Also the module has no tests at all, so the dependency could also be removed I guess.

Artur-: Good, removed. JUnit 4  should not be used for anything new either

sissbruecker (approved):

PR #1300, opened 2023-09-18
feat: Add support for generated header filters

Artur-:
Only adds filter fields for string properties for now


packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -78,27 +84,83 @@ function createDataProvider<TItem>(
...
     .map((name) => properties.find((prop) => prop.name === name))
     .filter(Boolean) as PropertyInfo[];
 
-  return effectiveProperties.map((p) => (
-    <GridSortColumn path={p.name} header={p.humanReadableName} key={p.name} autoWidth></GridSortColumn>
-  ));
+  return effectiveProperties.map((p) => {
+    const column = <GridSortColumn path={p.name} header={p.humanReadableName} key={p.name} autoWidth></GridSortColumn>;
+    if (options.headerFilters) {
+      const headerRenderer = useCallback(
```
sissbruecker, line 101: Using a hook like `useCallback` here is problematic, as this can easily lead to a different number of hook calls during rendering, which is not allowed. This can happen for example when changing the visible columns or turning header filters on or off. 

Alternatively we could:
- Create a single header renderer for all columns and memoize that once outside of any loops or conditions. The renderer would then need to access column-specific data (like property name) from the column instance itself which should already be passed in the props of the header renderer somehow. To store the property name we could use a data attribute.
- Or keep creating a separate renderer for each column. Then they would need to be stored in a map, which in turn is stored in a `useMemo` or a ref somewhere outside of any loops or conditions.

Another issue of memoizing the renderers is that the `setPropertyFilter` reference is always the initial one, it will never be updated on subsequent renders. That also means that `setPropertyFilter` will always reference the initial `internalFilter` state. That currently works, because the function mutates the filter object. But if we would add some other feature that calls `setInternalFilter`, then calling `setPropertyFilter` from one of the header renderers would undo that change.

Artur-: Changed to use one renderer. Does that resolve the problems?

Artur-: The header renderer needs to be memoized in some way because of https://github.com/vaadin/react-components/issues/146

Artur-: Is there a standard react way for handling communication with the singleton renderer then? Should the headerFilterRenderer be stored as state?

sissbruecker: The current hook usage looks safe due to the change to `useColumns` where you're signaling that it must be called everytime, and if someone want's to change that they will have to deal with refactoring it's content's too.

`setPropertyFilter` is still an issue, as the memoized header renderer still only get's the reference to the initial version of that function, which aways references the initial `internalFilter` state. To fix that you'd have to put `setPropertyFilter` into a ref on every render so that the header renderer can access the latest reference from there. The safest way would probably be to move all inputs of the renderer into a ref - for example when switching models, the renderer would need to access the updated `properties`, which it could only through a ref.

packages/ts/react-grid/src/autogrid.tsx
sissbruecker, line 1: The current UI is a bit odd in that the filter fields are displayed above the header. Taken from the CRUD web component, here is a variant that would render the filters in the second header row:

<details>

<summary>Code</summary>

```tsx
  const headerFilterRenderer = useCallback((column: any) => {
    // eslint-disable-next-line
    if (!column?.original) {
      return null;
    }
    // eslint-disable-next-line
    const { path } = column.original;
    const propertyInfo: PropertyInfo = properties.find((p) => p.name === path)!;

    return createFilterField(propertyInfo, {
      onInput: (e: { target: { value: string } }) => {
        const fieldValue = e.target.value;
        const filterValue = fieldValue;

        const filter = {
          propertyId: propertyInfo.name,
          filterValue,
          matcher: Matcher.CONTAINS,
        };

        // eslint-disable-next-line
        (filter as any).t = 'propertyString';
        setPropertyFilter(filter);
      },
    });
  }, []);

  const headerSorterRenderer = useCallback((column: any) => {
    // eslint-disable-next-line
    if (!column?.original) {
      return null;
    }
    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
    const { original } = column;
    // eslint-disable-next-line
    const path = original.dataset.path;
    const propertyInfo: PropertyInfo = properties.find((p) => p.name === path)!;

    return <GridSorter path={path}>{propertyInfo.humanReadableName}</GridSorter>;
  }, []);

  return effectiveProperties.map((p) => {
    if (options.headerFilters) {
      return (
        <GridColumnGroup key={`group${p.name}`} data-path={p.name} headerRenderer={headerSorterRenderer}>
          <GridColumn path={p.name} headerRenderer={headerFilterRenderer}></GridColumn>
        </GridColumnGroup>
      );
    }
    return <GridSortColumn path={p.name} header={p.humanReadableName} key={p.name} autoWidth></GridSortColumn>;
  });
```

</details>

packages/ts/react-grid/src/autogrid.tsx
Artur-: Change property filter to be a ref and also the properties and added test that the model can be changed, the header filters dynamically turned on and off, and that filters also work after a model change

packages/ts/react-grid/src/autogrid.tsx
Artur-: Moved below, thanks

packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -78,27 +91,108 @@ function createDataProvider<TItem>(
...
   service,
   model,
   filter,
   visibleColumns,
+  headerFilters,
   ...gridProps
 }: AutoGridProps<TItem>): JSX.Element {
+  const [internalFilter, setInternalFilter] = useState<AndFilter>({ ...{ t: 'and' }, children: [] });
+
+  const setHeaderPropertyFilter = useRef((propertyFilter: PropertyStringFilter) => {
```
sissbruecker, line 174: This setup has the same issue as the previous one 😅. The ref is only initialized on the first render, and then never updated again. So the function in the ref will always reference the initial `internalFilter` state.

I would suggest to just redefine the function here on every render as before, pass it to `useColumns`, and there store it in a ref on each render.

packages/ts/react-grid/test/autogrid.spec.tsx
```diff
@@ -126,26 +100,138 @@ describe('@hilla/react-grid', () => {
...
+      });
+      it('removes filters if turning header filters off', async () => {
+        const result = render(<AutoGrid service={personService} model={PersonModel} headerFilters></AutoGrid>);
+        await nextFrame();
+        await nextFrame();
+        result.rerender(<AutoGrid service={personService} model={PersonModel}></AutoGrid>);
+        await nextFrame();
+        await nextFrame();
+        const grid: GridElement = result.container.querySelector('vaadin-grid')!;
+        expect(getHeaderCellContent(grid, 0, 0).innerText).to.equal('First name');
```
sissbruecker, line 177: Should it also clear the filter state and reload from data provider?

There's also some weirdness if you enter a filter value and then toggle header filters. The grid will still be filtered, but the filter fields are empty because we don't initialize them from the initial filter. Maybe resetting filter state when disabling header filters is easiest for now.

packages/ts/react-grid/test/autogrid.spec.tsx
```diff
@@ -57,21 +22,30 @@ export async function nextFrame(): Promise<void> {
   });
 }
 
+async function assertColumns(result: RenderResult, ...ids: string[]) {
+  const grid = result.container.querySelector('vaadin-grid')!;
+  await nextFrame();
+  const columns = grid.querySelectorAll('vaadin-grid-column');
+  expect(columns.length).to.equal(ids.length);
+  for (let i = 0; i < ids.length; i++) {
+    expect(getHeaderCellContent(grid, 0, i).innerText).to.equal(_generateHeader(ids[i]));
```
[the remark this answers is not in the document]
Artur-: Yes but is it needed if the point is to check the visible columns? The header generation should rather be removed if we are checking the headers and not the data

packages/ts/react-grid/src/autogrid.tsx
Artur-: so just like `propertiesRef`...

Artur-: the pattern of
```
  const setPropertyFilterRef = useRef(setPropertyFilter);
  setPropertyFilterRef.current = setPropertyFilter;
```
just looks wrong..

packages/ts/react-grid/test/autogrid.spec.tsx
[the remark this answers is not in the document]
sissbruecker: I don't see tests that check that the correct data is rendered after switching models for example. Checking the path could be a cheap way to do that.

packages/ts/react-grid/src/autogrid.tsx
sissbruecker: Yeah, but that's how it works. The argument to `useRef` will only ever be used in the first call of the hook 🤷‍♂️.

packages/ts/react-grid/test/autogrid.spec.tsx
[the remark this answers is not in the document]
Artur-: Let's add that back then

packages/ts/react-grid/test/autogrid.spec.tsx
Artur-: Sounds reasonable to remove all filters if you remove the filter row

packages/ts/react-grid/test/autogrid.spec.tsx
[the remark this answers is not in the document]
Artur-: Fixed

PR #1301, opened 2023-09-18
chore: disable esbuild minification to avoid broken source maps

Lodin:
Trying to use packages minified by `esbuild` I found that source maps are broken. If the minification is disabled, everything works as expected. The problem is somewhere inside Vite remapping: it incorrectly remaps existing source maps during the build preparation.


Artur-:
I am not sure what this fixes

Lodin:
Ok, let's put it on hold until the error appears again

Issue #1271, opened 2023-09-14
Include the java type in the generated model metadata


Artur-:
To be able to e.g. format better in an autogrid, the model metadata needs to contain the Java type

cromoteca:
Did you think of any preferred format for that metadata?

cromoteca:
Some months ago I made a prototype at https://github.com/cromoteca/jackson-parser that uses a [reviver](https://github.com/cromoteca/hilla/blob/42bf5f1e554b2a341b7f0683ff59f862cc27754f/packages/ts/hilla-frontend/src/Connect.ts#L422-L442) to transform a JSON like

```json
{
    "date": {
        "__hilla_type": "Date",
        "value": "2023-09-14T12:59:44.632+02:00"
    },
    "uuid": "e8aee4dd-f894-47a8-b8da-61f018b4052f",
    "map": {
        "__hilla_type": "Map",
        "entries": [
            {
                "key": 3,
                "value": 4.0
            },
            {
                "key": 0,
                "value": 0.1
            },
            {
                "key": 1,
                "value": 2.0
            }
        ]
    },
    "bigInteger": {
        "__hilla_type": "BigInt",
        "value": "1234567890123456789012345678901234567890"
    },
    "other": {
        "__hilla_type": "Date",
        "value": "2023-09-14T13:16:24.633+02:00"
    }
}
```
to
```javascript
{ 
  date: Date Thu Sep 14 2023 12: 59: 44 GMT +0200(Central European Summer Time),
  uuid: "e8aee4dd-f894-47a8-b8da-61f018b4052f",
  map: Map(3), // now it's a record which loses Map generic types
  bigInteger: 1234567890123456789012345678901234567890n,
  other: Date Thu Sep 14 2023 13: 16: 24 GMT +0200(Central European Summer Time)
}
```
And a replacer to do the same in the opposite direction.

I'm not saying we should change our JSON format now (it would break everything). What I mean is that it would be nice to add type information in a way that can be used when dealing with JSON.

Would it be fine for you to find type information in the data rather than in the model? I guess no.

Artur-:
Sound inefficient to send the types always with the data. As the Java class structure is fixed, so should the type of values be so it should be enough to include it once in the model.

PR #1304, opened 2023-09-18
feat: add java type to parser

sissbruecker:
## Description

Adds the fully qualified Java type name to the generated `openapi.json`. This is part of getting the Java type information into the generated models in order to use them for CRUD features like choosing a proper column renderer or form field type.

Part of https://github.com/vaadin/hilla/issues/1271

### DIFF
--- a/packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
+++ b/packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
@@ -154,7 +154,8 @@ public final class ModelPlugin extends AbstractPlugin<PluginConfiguration> {
         } else if (signature instanceof ClassRefSignatureModel classRefSignatureModel) {
             typeName = classRefSignatureModel.getName();
         } else if (signature instanceof ArraySignatureModel arraySignatureModel) {
-            AnnotatedArrayType o = (AnnotatedArrayType) arraySignatureModel.get();
+            AnnotatedArrayType o = (AnnotatedArrayType) arraySignatureModel
+                    .get();
             typeName = o.toString();
         }
 
--- a/packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
+++ b/packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
@@ -1,5 +1,6 @@
 package dev.hilla.parser.plugins.model;
 
+import java.lang.reflect.AnnotatedArrayType;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
@@ -16,6 +17,7 @@ import dev.hilla.parser.core.PluginConfiguration;
 import dev.hilla.parser.models.AnnotatedModel;
 import dev.hilla.parser.models.AnnotationInfoModel;
 import dev.hilla.parser.models.AnnotationParameterModel;
+import dev.hilla.parser.models.ArraySignatureModel;
 import dev.hilla.parser.models.BaseSignatureModel;
 import dev.hilla.parser.models.ClassRefSignatureModel;
 import dev.hilla.parser.models.SignatureModel;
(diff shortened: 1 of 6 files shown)


packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
```diff
@@ -66,6 +69,13 @@ public void enter(NodePath<?> nodePath) {
         }
 
         var schema = typedNode.getTarget();
+        if (signature instanceof BaseSignatureModel) {
+            schema.addExtension(JAVA_TYPE_KEY,
```
Artur-, line 73: Will this leak the class name for custom beans into the client side? Can we avoid that somehow?

Lodin (commented):
Is adding that part really necessary? I mean, from the already generated model, it's pretty easy to extract the current type name. Even for `int` or `double` we can understand the type from the current OpenAPI structure.

sissbruecker:
> Is adding that part really necessary? I mean, from the already generated model, it's pretty easy to extract the current type name. Even for int or double we can understand the type from the current OpenAPI structure.

It seems to cover most cases, but not all of them. For example `LocalTime` gets generated as:
```
          "localTime" : {
            "type" : "string",
            "format" : "date-time",
            "nullable" : true
          },
```

Which is the same as `LocalDateTime`:
```
          "localDateTime" : {
            "type" : "string",
            "format" : "date-time",
            "nullable" : true
          },
```

There might be others?

packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
sissbruecker: We could include-list the types that we explicitly support with column renderers / input field types.

Artur-: I think we should limit it for now BUT we will want the information later, in some form, to know e.g that ”company” in the Person class refers to the Company class

sissbruecker: At least for the schema that information is already there, so we could expose it in the generator when necessary.

Artur-: So you are saying we should include everything in openapi.json, which is not published, but selectively include in generated TS?

packages/java/parser-jvm-plugin-model/src/test/resources/dev/hilla/parser/plugins/model/validation/openapi.json
```diff
@@ -138,8 +147,10 @@
             "nullable": true,
             "items": {
               "type": "string",
-              "nullable": true
+              "nullable": true,
+              "x-java-type": "java.lang.String"
             },
+            "x-java-type": "java.util.List",
```
Artur-, line 153: Should we leave generics out until we need them?

Artur- (dismissed):

packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
sissbruecker: For now I've added filtering for the Java types so that only primitives and any type from `java.**` will be included in the schema.

packages/java/parser-jvm-plugin-model/src/test/resources/dev/hilla/parser/plugins/model/validation/openapi.json
sissbruecker: The item type information is sort of there, in the `items` object:
```
"aStringList" : {
  "type" : "array",
  "nullable" : true,
  "items" : {
    "type" : "string",
    "nullable" : true,
    "x-java-type" : "java.lang.String"
  },
  "x-java-type" : "java.util.List"
},
```

Artur- (approved):

PR #1305, opened 2023-09-18
fix: check acl based on @Endpoint auth config not @EndpointExposed

### LINKED ISSUE #1250: AccessControl should come from the endpoint class, not a possible EndpointExposed class
### Description of the bug 
If you create a reusable `CrudEndpoint` class marked as `@EndpointExposed`, it makes no sense that access control is checked from that class. Instead you want to have e.g. a class with public info marked as `@AnonymousAllowed` and a class with private info as `@RolesAllowed("admin")`. All with the same super class and the methods only defined in the super class

```java
@EndpointExposed
public class SuperDuper {

public String hello() {return "Hello";}
}
```

```java
@Endpoint
@AnonymousAllowed
public class MyHello extends SuperDuper {
}
```

### Expected behavior
Anybody can call `MyHello.hello();`

### Actual behavior
Nobody can call `MyHello.hello();`
### Versions:
    - Vaadin / Hilla version: 2.2

taefi:
## Description
No matter a method is defined in a class annotated with `@EndpointExposed`
or in an Endpoint that extends it, the access control for the inherited methods
is checked based on the Endpoint class (the one annotated with either 
`@Endpoint` or `@BrowserCallable`). 

Inherited methods from a parent class annotated with `@EndpointExposed` 
would be treated by the class level access control rules of the children Endpoint .
If the access control rules for an inherited method should not follow the rules
of the child endpoint, that method should be overridden and annotated with 
access control annotations explicitly.

To make this policy even more clear, an exception should be thrown while starting 
the application that explicitly mentions that classes annotated with 
`@EndpointExposed` should not contain any of access control annotations and
the exception is preventing the confusion. See #1314

Fixes #1250

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


packages/java/endpoint/src/main/java/dev/hilla/EndpointInvoker.java
```diff
@@ -45,11 +32,27 @@
...
+import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
+import org.springframework.lang.NonNullApi;
+import org.springframework.util.ClassUtils;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.security.Principal;
+import java.util.*;
```
Artur-, line 52: This should be expanded and your IDE conf changed to use real imports

packages/java/endpoint/src/test/java/dev/hilla/auth/EndpointAccessCheckerTest.java
```diff
@@ -399,4 +417,258 @@ public void test() {
...
+            @AnonymousAllowed
+            public String sayHello() {
+                return "Hello from Test Endpoint";
+            }
+
+            public void test() {
+            }
+        }
+
+        shouldPassInherited(Test.class);
```
Artur-, line 462: Similarly here and below, e.g.
```
assertAccessGranted(Test.class, "sayHello");
assertAccessDenied(Test.class);
```

packages/java/endpoint/src/test/java/dev/hilla/auth/EndpointAccessCheckerTest.java
```diff
@@ -399,4 +417,258 @@ public void test() {
...
+            }
+        }
+
+        class Test extends ParentEndpoint {
+
+            public void test() {
+            }
+        }
+
+        shouldFailInherited(Test.class);
```
Artur-, line 436: Could this be more explicit so we are sure what we test? Like
```
assertAccessDenied(Test.class, "sayHello");
assertAccessDenied(Test.class);
```

packages/java/endpoint/src/test/java/dev/hilla/auth/EndpointAccessCheckerTest.java
```diff
@@ -399,4 +417,258 @@ public void test() {
             CurrentInstance.clearAll();
         }
     }
+
+    @Test
+    public void should_fail_When_Endpoint_is_not_annotated() throws Exception {
```
Artur-, line 422: It is not super clear what is being tested here. If we are testing that access is checked from  the correct method/class, maybe we do not need to test separately for all annotations and then test naming will become clearer, e.g. 
```
should_pass_When_Endpoint_overridden_method
```
is clearer that it tests how access is checked compared to `should_pass_When_Endpoint_overridden_method_is_AnonymousAllowed` where you don't get the relevance of `AnonymousAllowed` in the test and start wondering if checking is done using different methods/classes for different annotations

Artur- (changes requested):

packages/java/endpoint/src/test/java/dev/hilla/auth/EndpointAccessCheckerTest.java
taefi: Well, I definitely agree with the "naming should be improved" part, but in the first two tests it is testing with an Endpoint that is **not** annotated and inherits a method from a super class. In this case, being overridden is not gonna simply grant the access. If I change it to something like "access_granted_when_inherited_method_is_overridden_and_annotated" it can be still vague since annotating with "@DenyAll" doesn't count, however, as it is the only case, we can ignore and simplify since we are not testing the effect of annotations, but the correct target for access checking. All combinations of annotations don't require testing here, as they are tested separately.

packages/java/endpoint/src/test/java/dev/hilla/auth/EndpointAccessCheckerTest.java
```diff
@@ -48,14 +48,24 @@ private void createAnonymousContext() {
         when(requestMock.getUserPrincipal()).thenReturn(null);
     }
 
-    private void shouldPass(Class<?> test) throws Exception {
-        Method method = test.getMethod("test");
-        assertNull(checker.check(method, requestMock));
+    private void assertAccessGranted(Class<?> test, String methodName)
+            throws Exception {
+        Method method = test.getMethod(methodName);
+        if (method.getDeclaringClass().equals(test)) {
```
Artur-, line 54: Why is this logic needed here? It is like you code the test according to how the code should work?

packages/java/endpoint/src/test/java/dev/hilla/auth/EndpointAccessCheckerTest.java
Artur-: But then the test should not be about "access granted" but instead of which access control annotation is used?

packages/java/endpoint/src/test/java/dev/hilla/auth/EndpointAccessCheckerTest.java
taefi: This part is not related to the `EndpointAccessChecker` logic. This logic is tested in`EndpointInvokerTest`.

packages/java/endpoint/src/test/java/dev/hilla/auth/EndpointAccessCheckerTest.java
taefi: I think we are almost saying the same thing. I just emphasized that something like "should_pass_When_Endpoint_overridden_method" is not clear either and suggested a new one. The unnecessary tests that were re-checking the logic based on which access control annotation is used are removed now.

packages/java/endpoint/src/test/java/dev/hilla/EndpointInvokerTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+        endpointRegistry.registerEndpoint(test);
+
+        endpointInvoker.invoke("TestEndpoint", "sayhello", body, principal,
+                requestMock::isUserInRole);
+
+        Mockito.verify(endpointAccessChecker, Mockito.times(0))
+                .check(any(Method.class), any(), any());
+        Mockito.verify(endpointAccessChecker, Mockito.times(1))
+                .check(any(Class.class), any(), any());
```
platosha, line 133: There is some ambiguity regarding which class is used here. Should we assert that it is `TestEndpoint`, and not `ParentEndpoint`?

packages/java/endpoint/src/test/java/dev/hilla/EndpointInvokerTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+        TestEndpoint test = new TestEndpoint();
+
+        endpointRegistry.registerEndpoint(test);
+
+        endpointInvoker.invoke("TestEndpoint", "sayhello", body, principal,
+                requestMock::isUserInRole);
+
+        Mockito.verify(endpointAccessChecker, Mockito.times(1))
+                .check(any(Method.class), any(), any());
```
platosha, line 167: Would be nice to verify that the method being checked is `TestEndpoint::sayHello` and not `ParentEndpoint::sayHello`.

Artur- (approved):

PR #1308, opened 2023-09-18
test: Basic autogrid sort test

Artur-:


packages/ts/react-grid/test/autogrid.spec.tsx
```diff
@@ -63,6 +73,21 @@ describe('@hilla/react-grid', () => {
       expect(columns[2].path).to.equal('email');
       expect(columns[2].header).to.equal('Email');
     });
+    it('creates sortable columns', async () => {
+      const result = render(<TestAutoGrid />);
+      const grid: GridElement = result.container.querySelector('vaadin-grid')!;
+      await nextFrame();
+      await nextFrame();
```
sissbruecker, line 80: One await seems to be enough from my testing.

sissbruecker (approved):

packages/ts/react-grid/test/autogrid.spec.tsx
Artur-: It is a bit interesting because most of the time it is for this. For header filters it is not. However this test also failed every now or then without it..

Issue #1266, opened 2023-09-14
[autogrid] Hide @Id and @Version columns automatically


Artur-:
Any `@Id` or `@Version` column is for internal use and should not be shown as grid columns

PR #1311, opened 2023-09-19
feat: add annotations to schema

sissbruecker:
## Description

Adds annotations for properties to the schema. The annotations are filtered by an include list, which currently only includes `jakarta.persistence.Id` and `jakarta.persistence.Version`.

Part of https://github.com/vaadin/hilla/issues/1266

### DIFF
--- a/packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
+++ b/packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
@@ -2,6 +2,7 @@ package dev.hilla.parser.plugins.model;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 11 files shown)


packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
```diff
@@ -53,20 +72,31 @@ private static boolean isValidationConstraintAnnotation(
...
 
+    private static boolean isIncludedAnnotation(
+            AnnotationInfoModel annotation) {
+        return INCLUDED_ANNOTATIONS.contains(annotation.getName());
+    }
+
     @Override
     public void enter(NodePath<?> nodePath) {
-        if (!(nodePath.getNode() instanceof TypedNode)) {
+        if (!(nodePath.getNode()instanceof TypedNode typedNode)) {
```
Lodin, line 82:
```suggestion
        if (!(nodePath.getNode() instanceof TypedNode typedNode)) {
```

packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
```diff
@@ -53,20 +72,31 @@ private static boolean isValidationConstraintAnnotation(
...
             return;
         }
 
         var schema = typedNode.getTarget();
-        addConstraintsToSchema((AnnotatedNode) typedNode, schema);
+        addConstraintsToSchema(typedNode, schema);
+
+        // Add annotations from parent property model to schema
+        if (nodePath.getParentPath() != null && nodePath.getParentPath()
+                .getNode()instanceof PropertyNode propertyNode) {
```
Lodin, line 96:
```suggestion
                .getNode() instanceof PropertyNode propertyNode) {
```

Lodin (changes requested):
There are some stylistic issues to fix

packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
sissbruecker: Unfortunately that is how `mvn formatter:format` decides to format this.

Lodin: That's weird. Probably, they don't know what to do with that new syntax. But ok, let's have it that way then.

Lodin (dismissed):

Artur-:
Now you broke react-grid. Probably something with the Filter class and its annotations

packages/java/parser-jvm-plugin-model/src/main/java/dev/hilla/parser/plugins/model/ModelPlugin.java
Artur-: Oh look it was resolved with the latest formatter :)

Artur- (approved):

PR #1315, opened 2023-09-19
refactor(generator-typescript-plugin-model): remove obsolete validator generation

sissbruecker:
## Description

The generator currently contains two different ways of generating validators:
https://github.com/vaadin/hilla/blob/65d61dc3c4381755d32541404e9313f4eb49ce04/packages/ts/generator-typescript-plugin-model/src/ModelSchemaProcessor.ts#L235-L241

Where `getValidatorsFromAnnotations` looks up the `x-annotations` schema extension:
https://github.com/vaadin/hilla/blob/65d61dc3c4381755d32541404e9313f4eb49ce04/packages/ts/generator-typescript-plugin-model/src/ModelSchemaProcessor.ts#L279-L281

However that extension is never added to the schema, and there are no references in the codebase where a `x-annotations` property is ever being set.

Since this conflicts with the new feature proposed in #1311 , this PR would remove the unused functionality to free up the `x-annotations` extension for a different use.


sissbruecker:
@platosha When you find some time, could you please double-check that this is indeed not needed anymore?

Artur-:
All tests pass so it cannot be needed :)

platosha (approved):

PR #1321, opened 2023-09-20
chore: fix karma-vite configuration

Lodin:
This PR fixes Karma / Vite configuration to avoid broken CI


Artur-:
Which part is fixing the problem?

Lodin:
@Artur-, moving the Vite config to ge karma config, and some minor updates for it

.eslintrc
```diff
@@ -9,6 +9,7 @@
   "plugins": ["tsdoc"],
   "rules": {
     "@typescript-eslint/member-ordering": "off",
+    "@typescript-eslint/return-await": "off",
```
Artur-, line 12: Why do we want this and how is it related?

Lodin: It looks like a leftover from one of my attempts 🤔 We can remove it for now

Lodin: Done

packages/ts/react-form/test/useForm.spec.tsx
```diff
@@ -1,14 +1,18 @@
...
 import sinonChai from 'sinon-chai';
 import { useForm as _useForm, useFormPart } from '../src/index.js';
-import { type Login, LoginModel, UserModel } from './models.js';
+import { type Login, LoginModel, type UserModel } from './models.js';
 
 use(sinonChai);
+use(chaiDom);
+
+describe('@hilla/react-form', function () {
+  this.timeout(100000);
```
platosha, line 14: I guess this is leftover from debugging, let's clean it up.

```suggestion
```

scripts/build.ts
```diff
@@ -19,7 +19,7 @@ await build({
   },
   entryPoints: srcFiles.map((file) => new URL(file, root)).map(fileURLToPath),
   format: 'esm',
-  minify: true,
+  // minify: true,
```
platosha, line 22: I'm curious, what was the reason for disabling this? Was minified code problematic for debugging?

platosha (changes requested):

scripts/build.ts
Lodin: This is a debug leftover. Though, the minification causes Vite to hiccup on sourcemaps for some reason, so all the lines are messed up. That's why I'm considering disabling this option until this bug is fixed. What do you think?

package.json
```diff
@@ -30,7 +30,11 @@
   "license": "Apache-2.0",
   "overrides": {
     "esbuild": "$esbuild",
-    "prettier": "$prettier"
+    "prettier": "$prettier",
+    "@vaadin/vaadin-lumo-styles": "24.2.0-rc1",
```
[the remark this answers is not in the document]
Lodin: Good notice. There was some `dev` versions of the package left across web components. Let's check if it is correct now and file an issue if not.

scripts/build.ts
platosha: > This is a debug leftover. Though, the minification causes Vite to hiccup on sourcemaps for some reason, so all the lines are messed up. That's why I'm considering disabling this option until this bug is fixed. What do you think?

Agree. Also bug reports with real thing names in the error messages and stack traces are much nicer than what we have now with mangled names.

platosha: It's fine to address minification disabling in a separate PR though.

Lodin: Yep, will do a separate one

Lodin: Btw, mangled names are a bug I believe. I know why it happened and it would be better to fix it eventually.

platosha (approved):

PR #1322, opened 2023-09-20
test: copy setup from EndpointHotSwapListenerTest

Artur-:


Artur-:
I think this works only because all the tests reuse the same context as they have exactly the same setup. Then there is no conflict of the static variable...

PR #1323, opened 2023-09-21
feat: enabling default CRUD services just by extending CrudRepositoryService

taefi:
## Description

This enables the developers to define a `CrudRepositoryService` just by extending it and setting the correct generic type, no need to have a weird constructor that passes the same types as defined in generic type while defining the service class. 

Thus, the use would be like:
```java
@BrowserCallable
@AnonymousAllowed
public class PersonService extends CrudRepositoryService<Person, Long, PersonRepository> {
}
```  
instead of 
```java
@BrowserCallable
@AnonymousAllowed
public class PersonService extends CrudRepositoryService<Person, Long> {

    public PersonService(PersonRepository repository) {
        super(Person.class, repository);
    }
}
```
This way, it resembles the way developers are using/extending `JpaRepository`, so it would be easier to relate to, and of course, achieve the same thing with less lines of codes.

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


Artur- (commented):
Finding the entity type this way sounds great IMO. Taking away how the repository is autowired is more questionable. At least we should retain the constructor where you can manually specify both.

taefi:
> Finding the entity type this way sounds great IMO. Taking away how the repository is autowired is more questionable. At least we should retain the constructor where you can manually specify both.

Completely agree on having the constructor with args (apart from the default one) to give the freedom of choice to developers for passing any Repository instance they want.

Artur-:
Having generics for the repository type might still be good as you will likely want to add some of your own methods at some point and if `getRepository` returns the wrong type, you will be sad

packages/java/endpoint/pom.xml
```diff
@@ -219,6 +219,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>com.vaadin.external</groupId>
+            <artifactId>gentyref</artifactId>
+            <version>1.2.0.vaadin1</version>
```
Artur-, line 225: the version is specified in the main pom, use that here

packages/java/endpoint/src/main/java/dev/hilla/crud/CrudRepositoryService.java
```diff
@@ -1,49 +1,104 @@
...
+    /**
+     * Initializes the repository instance if it hasn't been initialized yet,
+     * otherwise returns the existing one.
+     * <p>
+     * This method uses ApplicationContext to obtain the Repository instance, so
+     * it is not suitable for use in the constructor.
+     *
+     * @return the repository instance
+     */
+    protected R getRepository() {
```
Artur-, line 60: Most of the time it is easier to init in a `@PostConstruct` so you find any problems immediately

packages/java/endpoint/src/test/java/dev/hilla/crud/CrudRepositoryServiceTest.java
```diff
@@ -0,0 +1,393 @@
...
+        CrudRepositoryServiceTest.CustomCrudRepositoryService.class,
+        CrudRepositoryServiceTest.CustomJpaRepository.class,
+        CrudRepositoryServiceTest.CustomJpaRepositoryService.class })
+@ContextConfiguration(classes = { CrudConfiguration.class })
+@RunWith(SpringRunner.class)
+@EnableAutoConfiguration
+public class CrudRepositoryServiceTest {
+
+    @Repository
+    static class CustomCrudRepository
```
Artur-, line 41: Might be easier to run as a `@DataJpaTest` like https://github.com/vaadin/hilla/blob/main/packages/java/endpoint/src/test/java/dev/hilla/crud/FilterTest.java#L22

packages/java/tests/spring/react-grid-test/src/main/java/dev/hilla/test/reactgrid/CompanyService.java
```diff
@@ -9,9 +9,10 @@
 @BrowserCallable
 @Service
 @AnonymousAllowed
-public class CompanyService extends CrudRepositoryService<Company, Long> {
+public class CompanyService
+        extends CrudRepositoryService<Company, Long, CompanyRepository> {
 
     public CompanyService(CompanyRepository repository) {
```
Artur-, line 15: I would remove this as this is not a test for `CrudRepositoryService`

Artur- (changes requested):

packages/java/endpoint/src/test/java/dev/hilla/crud/CrudRepositoryServiceTest.java
taefi: It seems that `@DataJpaTest` is only initializing the Persistence Layer context, and even using it in combination with `@RunWith(SpringRunner.class)` and `@ContextConfiguration(classes = { CrudConfiguration.class })` still does not support autowiring of the service objects.

packages/java/tests/spring/react-grid-test/src/main/resources/data.sql
```diff
@@ -48,6 +48,3 @@ INSERT INTO person (id, version, first_name, last_name, lucky_number) VALUES (47
 INSERT INTO person (id, version, first_name, last_name, lucky_number) VALUES (48, 0, 'Vivian', 'Xavier',19);
 INSERT INTO person (id, version, first_name, last_name, lucky_number) VALUES (49, 0, 'William', 'Young',22);
 INSERT INTO person (id, version, first_name, last_name, lucky_number) VALUES (50, 0, 'Xander', 'Zane',18);
-
```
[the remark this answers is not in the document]
taefi: That was a mistake :)

Artur- (approved):
The constructor in `CompanyService` still feels unnecessary

Artur- (dismissed):

Artur- (approved):

packages/java/tests/spring/native/pom.xml
```diff
@@ -28,6 +28,11 @@
         <dependency>
             <groupId>com.vaadin</groupId>
             <artifactId>vaadin-lumo-theme</artifactId>
+            <version>24.1.10</version>
```
Artur-, line 31: Why do you need the version here?

Artur-: It's wrong btw :)

Artur-: All the dependencies could be replaced by the standard dev.hilla/hilla dependencies. See how the other tests were changed

Artur- (changes requested):
the native deps should be fixed as the project will be broken when multiple versions of the components are mixed, otherwise lgtm

Artur- (approved):

taefi:
> the native deps should be fixed as the project will be broken when multiple versions of the components are mixed, otherwise lgtm

@Artur- I think I may not fully understand what should be fixed, did you mean to have the `package.json` and `package-lock.json` also committed for the native test app (as we didn't have the `@vaadin/bundles` before in the `package.json` and now with the newew Hilla version it is also added and can collide with `@hilla/react-components`?

Artur-:
I meant the Maven deps you already fixed

PR #1326, opened 2023-09-22
feat: Support header filters for number fields

### LINKED ISSUE #1284: [autogrid] header filter support for number columns
When the column contains numbers, the header filter should be a combination of a select and a text field
image

The select contains options for how to match the value (>, = or <), the text field value is used for matching

The above image is captured from the code
```jsx
          <Select
            renderer={() => (
              <ListBox>
                <Item value="GREATER_THAN" {...{ label: ">" }}>
                  &gt; Greater than
                </Item>
                <Item value="LESS_THAN" {...{ label: "<" }}>
                  &lt; Less than
                </Item>
                <Item value="EQUALS" {...{ label: "=" }}>
                  = Equals
                </Item>
              </ListBox>
            )}
            className={css.matchSelect}
            value="GREATER_THAN"
          ></Select>
          <TextField/>
```

and

```css
.matchSelect {
  --vaadin-field-default-width: 2em;
  margin-right: 3px;
}
.matchSelect > vaadin-select-value-button {
  --_lumo-text-field-overflow-mask-image: none !important;
}

.matchSelect::part(toggle-button) {
  display: none;
}
```

Note that the text field must support input in whatever format the grid column shows data in. If e.g. grid formats "1000" as "1,000" then filtering using both "1000" and "1,000" must work. Also the text in the field should be aligned the same way as the column data (left / right)

Artur-:
Fixes #1284


packages/java/endpoint/src/main/java/dev/hilla/crud/filter/PropertyStringFilter.java
sissbruecker, line 1: The `PropertyStringFilter` indicated to me that it filters string properties, only after some thinking I realize the name might be related to the filter value. Maybe just use `PropertyFilter`.

Was also surprised that we use a string value for filtering numbers, though it seems to work as expected.

packages/ts/react-grid/src/header-filter.tsx
```diff
@@ -0,0 +1,81 @@
...
+  useEffect(() => {
+    // Workaround for https://github.com/vaadin/react-components/issues/148
+    setTimeout(() => {
+      if (select.current) {
+        select.current.requestContentUpdate();
+      }
+    }, 1);
+  }, []);
+
+  useEffect(() => {
```
sissbruecker, line 24: Nit: I'd suggest to just use a function with arguments for value + matcher, and call that from the event handlers instead of relying on detecting state changes. It's a bit unexpected that you would get a `setPropertyFilter` call just from rendering the component initially. That way you can also remove `changed` checks in the filter implementation.

packages/java/endpoint/src/main/java/dev/hilla/crud/filter/PropertyStringFilter.java
Artur-: It relates to the filter value being a string rather than e.g. a date

Artur-: Let's see what is needed when we add date filters

packages/ts/react-grid/src/header-filter.tsx
Artur-: Where would you track the current values?

packages/ts/react-grid/src/header-filter.tsx
```diff
@@ -0,0 +1,81 @@
...
+            </ListBox>
+          )}
+          className={css.filterWithLessGreaterEquals}
+          value={matcher}
+        ></Select>
+        <TextField
+          placeholder="Filter..."
+          onInput={(e) => {
+            const fieldValue = ((e as InputEvent).target as TextFieldElement).value;
+            setFilterValue(fieldValue);
```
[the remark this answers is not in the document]
Artur-: Fixed

packages/ts/react-grid/src/header-filter.tsx
sissbruecker: I would still keep the states, but set them from said function:
```ts
const [matcher, setMatcher] = useState(Matcher.GREATER_THAN);
const [filterValue, setFilterValue] = useState('');

function updateFilter(newFilterValue: string, newMatcher: string) {
  setFilterValue(newFilterValue);
  setMatcher(newMatcher);

  const filter = {
    propertyId: context.propertyInfo.name,
    filterValue,
    matcher,
  };
  // eslint-disable-next-line
  (filter as any).t = 'propertyString';
  context.setPropertyFilter(filter);
} 
```

In cases where you don't want to change one of the states, for example in the select, just pass the current state as argument.

Artur-: Ok, what I thought. It's a bit fiddly because you cannot use both values from the state as they don't update immediately. But probably works better this way.

Fixed

sissbruecker: Sorry, I meant something like this:
```diff
@@ -9,18 +9,6 @@
 import Matcher from './types/dev/hilla/crud/filter/PropertyStringFilter/Matcher';
 import { PropertyInfo } from './utils.js';
 
-function triggerFilterUpdate(matcher: Matcher, filterValue: string, context: HeaderColumnContext) {
-  // Update filter
-  const filter = {
-    propertyId: context.propertyInfo.name,
-    filterValue,
-    matcher,
-  };
-  // eslint-disable-next-line
-  (filter as any).t = 'propertyString';
-  context.setPropertyFilter(filter);
-}
-
 export function HeaderFilter(): ReactElement {
   const context = useContext(HeaderColumnContext)!;
   const [matcher, setMatcher] = useState(Matcher.GREATER_THAN);
@@ -35,13 +23,26 @@
     }, 1);
   }, []);
 
+  function updateFilter(newFilterValue: string, newMatcher: Matcher) {
+    setFilterValue(newFilterValue);
+    setMatcher(newMatcher);
+    const filter = {
+      propertyId: context.propertyInfo.name,
+      filterValue: newFilterValue,
+      matcher: newMatcher,
+    };
+    // eslint-disable-next-line
+    (filter as any).t = 'propertyString';
+    context.setPropertyFilter(filter);
+  }
+
   if (context.propertyInfo.modelType === 'string') {
     return (
       <TextField
         placeholder="Filter..."
         onInput={(e: any) => {
           const fieldValue = ((e as InputEvent).target as TextFieldElement).value;
-          triggerFilterUpdate(Matcher.CONTAINS, fieldValue, context);
+          updateFilter(fieldValue, Matcher.CONTAINS);
         }}
       ></TextField>
     );
@@ -52,8 +53,7 @@
           ref={select}
           onValueChanged={(e) => {
             const newMatcher = e.detail.value as Matcher;
-            setMatcher(newMatcher);
-            triggerFilterUpdate(newMatcher, filterValue, context);
+            updateFilter(filterValue, newMatcher);
           }}
           renderer={() => (
             <ListBox>
@@ -75,8 +75,7 @@
           placeholder="Filter..."
           onInput={(e) => {
             const fieldValue = ((e as InputEvent).target as TextFieldElement).value;
-            setFilterValue(fieldValue);
-            triggerFilterUpdate(matcher, fieldValue, context);
+            updateFilter(fieldValue, matcher);
           }}
         />
       </>
```

Artur-: That's another way of writing the same thing :)

Artur-: A bit cleaner, applied

PR #1327, opened 2023-09-22
feat(react-auth): allow customizing user type

sissbruecker:
## Description

Working on updating the starters it was rather odd that the auth state contains a user object with two dozen properties that might exist, but you can't really be sure unless you double check in the backend. This proposes an alternative that would allow providing a custom user type to the auth module. It also does some minor improvements to how the context works, so that the auth provider doesn't require passing a value, and the `useAuth` hook is actually the hook that devs can call to get the auth state from the context.

To start with, apps would have some module like `frontend/util/auth.ts` that configures the auth module to use a specific user type:
```ts
import { configureAuth } from "@hilla/react-auth";
import { UserEndpoint } from "Frontend/generated/endpoints";

// Configure auth to use `UserEndpoint.getAuthenticatedUser`
const auth = configureAuth(UserEndpoint.getAuthenticatedUser);

// Export auth provider and useAuth hook, which are automatically typed to the result of `UserEndpoint.getAuthenticatedUser`
export const useAuth = auth.useAuth;
export const AuthProvider = auth.AuthProvider;
```

Adding the auth provider to the app then becomes:
```diff
import router from 'Frontend/routes.js';
- import { AuthContext, useAuth } from 'Frontend/useAuth.js';
+ import { AuthProvider } from 'Frontend/utils/auth';
import { RouterProvider } from 'react-router-dom';

export default function App() {
-   const auth = useAuth();
  return (
-     <AuthContext.Provider value={auth}>
+     <AuthProvider>
      <RouterProvider router={router} />
-     </AuthContext.Provider>
+    </ AuthProvider>
  );
}
```

And accessing the auth state then becomes:
```diff
- import { useContext } from 'react';
- import { AuthContext } from '@hilla/react-auth';
+ import { useAuth } from 'frontend/utils/auth';

- const { state } = useContext(AuthContext);
+ const { state } = useAuth();

// state.user is now the same type as return type from UserEndpoint.getAuthenticatedUser
```

### DIFF
--- a/packages/ts/react-auth/src/useAuth.tsx
+++ b/packages/ts/react-auth/src/useAuth.tsx
@@ -136,7 +136,7 @@ export type Authentication<TUser extends AuthUser> = Readonly<{
  * The hook that can be used to get the authentication state.
  * It returns the state of the authentication.
  */
-const AuthContext = createContext<Authentication<AuthUser>>({
+export const AuthContext = createContext<Authentication<AuthUser>>({
   state: initialState,
   async login() {
     throw new Error('AuthContext not initialized');
--- a/packages/ts/react-auth/src/useAuth.tsx
+++ b/packages/ts/react-auth/src/useAuth.tsx
@@ -1,5 +1,5 @@
 import { login as _login, logout as _logout, type LoginResult } from '@hilla/frontend';
-import { createContext, type Dispatch, useEffect, useReducer } from 'react';
+import { createContext, type Dispatch, type ReactNode, useContext, useEffect, useReducer } from 'react';
 
 type LoginFunction = (username: string, password: string) => Promise<LoginResult>;
 type LogoutFunction = () => Promise<void>;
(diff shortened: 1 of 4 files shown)


packages/ts/react-auth/src/useAuth.tsx
```diff
@@ -18,38 +18,20 @@ const LOGOUT = 'LOGOUT';
...
-  email?: string;
-  emailVerified?: boolean;
-  gender?: string;
-  birthdate?: string;
-  zoneinfo?: string;
-  locale?: string;
-  phoneNumber?: string;
-  phoneNumberVerified?: boolean;
-  roles?: string[];
+  roles: string[];
```
sissbruecker, line 21: Regarding roles: I just made them required for now, so you'd have to add them in order to use the package. I was thinking about adding an accessor function, but thought that might be more hassle than just having a `roles` property as convention.

cromoteca: SSO Kit doesn't have roles right now. I'm not sure that imposing them is a good idea, given that we're giving freedom about the `User` object.
What about allowing to define a custom role function while giving a default one that tries to access `roles` as a `string[]`?

cromoteca (approved):
I tested it in an app and everything works. I cannot find a better solution for `auth.ts`, so OK.

PR #1337, opened 2023-09-22
fix: add support for proxied services as endpoints

taefi:
## Description
If a service class that is annotated with `@Endpoint` or `@BrowserCallable` is proxied at runtime e.g. by
annotating it by `@Transactional`, the invoked endpoint at runtime would be proxied subclass of the actual endpoint. In this case, resolving the target for access checking should consider resolving the original endpoint class not the runtime proxy object.

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/packages/java/endpoint/src/main/java/dev/hilla/EndpointInvoker.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/EndpointInvoker.java
@@ -38,6 +38,7 @@ import jakarta.validation.Validation;
 import jakarta.validation.Validator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.cglib.proxy.Enhancer;
 import org.springframework.context.ApplicationContext;
 import org.springframework.http.ResponseEntity;
 import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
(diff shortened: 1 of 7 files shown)


packages/java/endpoint/src/main/java/dev/hilla/EndpointInvoker.java
```diff
@@ -384,6 +385,12 @@ private Object invokeVaadinEndpointMethod(String endpointName,
         var invokedEndpointClass = vaadinEndpointData.getEndpointObject()
                 .getClass();
 
+        // if the invoked endpoint is proxied, then take the original class:
+        // This is only to supports CGLIB proxies as Spring AOP proxies are.
+        if (Enhancer.isEnhanced(invokedEndpointClass)) {
+            invokedEndpointClass = invokedEndpointClass.getSuperclass();
+        }
```
mcollovati, line 392: If there's no way to avoid coupling with Spring, I would suggest using `ClassUtils.getUserClass`, as already done below in this method

```java
        Type[] javaParameters = getJavaParameters(methodToInvoke, ClassUtils
                .getUserClass(vaadinEndpointData.getEndpointObject()));
```

Artur- (approved):

PR #1343, opened 2023-09-25
refactor: introduce model options

sissbruecker:
## Description

Change model constructors to accept an optional options object instead of a variable number of validator arguments.

### DIFF
--- a/packages/ts/form/src/BinderNode.ts
+++ b/packages/ts/form/src/BinderNode.ts
@@ -25,7 +25,6 @@ import {
   type ArrayItemModel,
   ArrayModel,
   getObjectModelOwnAndParentGetters,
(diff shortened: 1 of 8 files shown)


Lodin (commented):
Just thought... What about adding all params into the options object, not only validators?

sissbruecker:
> Just thought... What about adding all params into the options object, not only validators?

@Lodin Anton was kind of against it, so for now we just move the validators over, which are kind of optional, as will be the other properties that we are going to add.

Lodin:
@sissbruecker actually, we discussed it a bit at Friday, and he came to some similar ideas.

@platosha, what do you think?

platosha (approved):

PR #1345, opened 2023-09-25
feat: Auto Form component

Artur-:
Can be used as
```
<ExperimentalAutoForm service={AppointmentService} model={AppointmentModel} />
```

to auto generate a form using AppointmentModel and allow submitting the result to the update method of the AppointmentService

### DIFF
--- a/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudService.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudService.java
@@ -12,7 +12,7 @@ import org.springframework.data.domain.Pageable;
  * <p>
  * Note! Not yet fully implemented but limited to read operations
  */
-public interface CrudService<T> {
+public interface CrudService<T, ID> {
 
     /**
      * Lists objects of the given type using the paging and sorting options
@@ -26,4 +26,25 @@ public interface CrudService<T> {
      */
     List<T> list(Pageable pageable, @Nullable Filter filter);
 
+    /**
+     * Updates (or creates) the given object and returns a potentially updated
+     * object.
+     * <p>
+     * If you store the object in a SQL database, the returned object might have
+     * a new id or updated consistency version.
+     * 
+     * @param value
+     *            the object to update/store
+     * @return the fresh object or {@code null} if no object was found to update
+     */
+    @Nullable
+    T update(T value);
+
+    /**
+     * Delets the object with the given id.
+     * 
+     * @param id
+     *            the id of the object to delete
+     */
+    void delete(ID id);
 }
--- a/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudService.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudService.java
@@ -32,7 +32,7 @@ public interface CrudService<T, ID> {
      * <p>
      * If you store the object in a SQL database, the returned object might have
      * a new id or updated consistency version.
-     * 
+     *
      * @param value
      *            the object to update/store
      * @return the fresh object or {@code null} if no object was found to update
@@ -42,7 +42,7 @@ public interface CrudService<T, ID> {
 
     /**
      * Delets the object with the given id.
-     * 
+     *
      * @param id
      *            the id of the object to delete
      */
--- a/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudService.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudService.java
@@ -27,18 +27,17 @@ public interface CrudService<T, ID> {
     List<T> list(Pageable pageable, @Nullable Filter filter);
 
     /**
-     * Updates (or creates) the given object and returns a potentially updated
-     * object.
+     * Saves the given object and returns the (potentially) updated object.
      * <p>
      * If you store the object in a SQL database, the returned object might have
      * a new id or updated consistency version.
      *
      * @param value
-     *            the object to update/store
+     *            the object to save
      * @return the fresh object or {@code null} if no object was found to update
      */
     @Nullable
-    T update(T value);
+    T save(T value);
 
     /**
      * Delets the object with the given id.
--- a/packages/java/tests/spring/react-grid-test/src/test/java/dev/hilla/test/reactgrid/AutoFormViewIT.java
+++ b/packages/java/tests/spring/react-grid-test/src/test/java/dev/hilla/test/reactgrid/AutoFormViewIT.java
@@ -30,11 +30,6 @@ public class AutoFormViewIT extends ChromeBrowserTest {
 
         // Field values remain
         Assert.assertEquals("It's me, Mario", getTextField("name").getValue());
-        // Custom error message shown
-        Assert.assertEquals(
-                "Something went wrong, please check all your values",
-                $("div").id("formerror").getText());
-
     }
 
     @Test
@@ -44,9 +39,11 @@ public class AutoFormViewIT extends ChromeBrowserTest {
         getTextField("time").setValue("2020-10-03T09:20:00.000Z");
         submit();
 
-        Assert.assertEquals(
-                "Thank you John Doe, your appointment has been reserved.",
-                $("*").id("form-submitted").getText());
+        waitUntil(driver -> {
+            return $("*").id("form-submitted").getText().equals(
+                    "Thank you John Doe, your appointment has been reserved.");
+
+        });
 
     }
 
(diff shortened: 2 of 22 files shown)


packages/ts/react-grid/src/autoform.tsx
```diff
@@ -0,0 +1,80 @@
...
+              setFormError('');
+              const newItem = await form.submit();
+              if (!newItem) {
+                // If update returns an empty object, then no update was performed
+                throw new Error('generic error');
+              } else if (onSubmit) {
+                onSubmit({ item: newItem });
+              } else {
+                form.clear();
+              }
```
sissbruecker, line 60: Don't quite understand this part: If there is a callback, then we don't clear the form, otherwise we do.

I think the form should be cleared when:
- There is no `item` and the form has been successfully submitted
- The `item` state changes to null. Might already be covered by `form.read`, not sure though.

packages/ts/react-grid/src/autoform.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const [formError, setFormError] = useState('');
+  useEffect(() => {
+    form.read(item);
+  }, [item]);
+
+  return (
+    <VerticalLayout theme="padding spacing">
+      {getProperties(model).map((propertyInfo) => (
+        <AutoFormField
+          key={`auto-form-field-${propertyInfo.name}`}
```
sissbruecker, line 38:
```suggestion
          key={propertyInfo.name}
```
Nit: The key is never rendered into the DOM, it just needs to be a unique value. As such we can remove any constants from it as those don't make it any more unique.

packages/ts/react-grid/test/autoform.spec.tsx
```diff
@@ -0,0 +1,155 @@
...
+        someNumber: 24451,
+        vip: false,
+      };
+      const result = render(<AutoForm service={personService} model={PersonModel} item={person} />);
+      await assertFormFieldValue(result, 'First name', person.firstName);
+      await assertFormFieldValue(result, 'Last name', person.lastName);
+      await assertFormFieldValue(result, 'Email', person.email);
+      await assertFormFieldValue(result, 'Some number', person.someNumber);
+    });
+    it('works without an item', async () => {
```
sissbruecker, line 67: Could add a test that the form gets updated when passing a different item or switching from an item to null.

packages/ts/react-grid/test/autoform.spec.tsx
```diff
@@ -0,0 +1,155 @@
...
+      const service: CrudService<Person> & HasLastFilter = createService<Person>(personData);
+
+      const person = await getItem(service, 1);
+      let submittedItem: Person | undefined;
+      const result = render(
+        <AutoForm
+          service={service}
+          model={PersonModel}
+          item={person}
+          onSubmit={({ item }) => {
```
sissbruecker, line 106: Nit: Could use a spy here.

packages/ts/react-grid/test/autoform.spec.tsx
sissbruecker, line 1: Clearing the form is lacking tests.

packages/java/endpoint/src/main/java/dev/hilla/crud/CrudService.java
```diff
@@ -26,4 +26,25 @@ public interface CrudService<T> {
...
+     * <p>
+     * If you store the object in a SQL database, the returned object might have
+     * a new id or updated consistency version.
+     *
+     * @param value
+     *            the object to update/store
+     * @return the fresh object or {@code null} if no object was found to update
+     */
+    @Nullable
+    T update(T value);
```
sissbruecker, line 41: How about using `save` here? Update makes it look like there should be a `create`, and you need to check the API docs to verify that this can also create items. `Save` is already familiar terminology from JPA.

packages/ts/react-grid/test/autoform.spec.tsx
```diff
@@ -0,0 +1,155 @@
...
+    it('uses values from an item', async () => {
+      const person = (await getItem(personService, 2))!;
+
+      const result = render(<AutoForm service={personService} model={PersonModel} item={person} />);
+      await assertFormFieldValue(result, 'First name', person.firstName);
+      await assertFormFieldValue(result, 'Last name', person.lastName);
+      await assertFormFieldValue(result, 'Email', person.email);
+      await assertFormFieldValue(result, 'Some number', person.someNumber);
+    });
+    it('submits a valid form', async () => {
```
sissbruecker, line 83: Could also test creating a new item.

packages/ts/react-grid/test/autoform.spec.tsx
```diff
@@ -0,0 +1,155 @@
...
+      };
+      const person = await getItem(service, 1);
+      const result = render(<AutoForm service={service} model={PersonModel} item={person} />);
+      await submit(result);
+      await nextFrame();
+      // eslint-disable-next-line
+      const formError = (result.container.querySelector('#formerror') as HTMLElement).innerText;
+      expect(formError).to.equal('Something went wrong, please check all your values');
+    });
+    it('calls onSubmitEerror and does not show error if the endpoint call fails', async () => {
```
sissbruecker, line 130: Could also test that `onSubmit` is not called when there is an error.

packages/java/tests/spring/react-grid-test/src/test/java/dev/hilla/test/reactgrid/AutoFormViewIT.java
```diff
@@ -0,0 +1,74 @@
...
+        getTextField("doctor").setValue("Evil");
+        getTextField("time").setValue("now");
+        submit();
+
+        // Field values remain
+        Assert.assertEquals("It's me, Mario", getTextField("name").getValue());
+        // Custom error message shown
+        Assert.assertEquals(
+                "Something went wrong, please check all your values",
+                $("div").id("formerror").getText());
```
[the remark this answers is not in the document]
Artur-: Removed

packages/ts/react-grid/test/autoform.spec.tsx
Artur-: Added

packages/ts/react-grid/test/autoform.spec.tsx
Artur-: Fixed

packages/ts/react-grid/test/autoform.spec.tsx
Artur-: Done

packages/ts/react-grid/test/autoform.spec.tsx
Artur-: Added

packages/ts/react-grid/src/autoform.tsx
Artur-: Yeah the thought was that if you handle submit yourself you might also want to control what happens to the form. But maybe it always should be cleared.

packages/ts/react-grid/src/autoform.tsx
```diff
@@ -0,0 +1,80 @@
...
+          form={form}
+        ></AutoFormField>
+      ))}
+      <div id="formerror" style={{ color: 'red' }}>
+        {formError}
+      </div>
+      <HorizontalLayout theme="spacing">
+        <Button
+          // eslint-disable-next-line @typescript-eslint/no-misused-promises
+          onClick={async () => {
```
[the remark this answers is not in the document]
Artur-: Done

packages/java/endpoint/src/main/java/dev/hilla/crud/CrudService.java
Artur-: That would maybe fit better, I will change it

Artur-: Done

packages/ts/react-grid/src/autoform.tsx
sissbruecker: Always clearing might be a problem when implementing the CRUD view, I would expect that a selected item stays in the form after saving. But we can also deal with this later.

packages/ts/react-grid/test/autoform.spec.tsx
sissbruecker: I think you missed to update it in this specific test.

packages/ts/react-grid/test/autoform.spec.tsx
sissbruecker: It's probably OK to leave this out.

packages/ts/react-grid/test/autoform.spec.tsx
Artur-: done

sissbruecker (approved):

PR #1352, opened 2023-09-26
feat: AutoCrud component

Artur-:


packages/ts/react-grid/src/autoform-field.tsx
```diff
@@ -5,12 +5,13 @@ import type { PropertyInfo } from './utils.js';
 type AutoFormFieldProps = {
   propertyInfo: PropertyInfo;
   form: UseFormResult<any>;
+  disabled?: boolean;
 };
-export function AutoFormField({ propertyInfo, form }: AutoFormFieldProps): JSX.Element {
+export function AutoFormField({ propertyInfo, form, disabled }: AutoFormFieldProps): JSX.Element {
```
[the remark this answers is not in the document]
Artur-: Changed the logic so that the selection is kept and form values are kept

packages/ts/react-grid/src/autocrud.tsx
```diff
@@ -0,0 +1,53 @@
...
+          service={service}
+          model={model}
+          onActiveItemChanged={(e) => {
+            const activeItem = e.detail.value;
+            (e.target as GridElement).selectedItems = activeItem ? [activeItem] : [];
+          }}
+          onSelectedItemsChanged={(e) => {
+            const selectedItem = e.detail.value[0];
+            setItem({ ...selectedItem });
+          }}
```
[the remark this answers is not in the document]
Artur-: Fixed

packages/ts/react-grid/src/autocrud.tsx
```diff
@@ -0,0 +1,53 @@
...
+        ></AutoGrid>
+        <ExperimentalAutoForm
+          disabled={!item}
+          service={service}
+          model={model}
+          item={item}
+          onSubmit={() => {
+            // Trigger grid data refresh
+            setRefreshTrigger(refreshTrigger + 1);
+          }}
```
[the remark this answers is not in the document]
Artur-: Fixed

packages/ts/react-grid/src/autocrud.tsx
```diff
@@ -0,0 +1,58 @@
...
+        ></AutoGrid>
+        <ExperimentalAutoForm
+          disabled={!item}
+          service={service}
+          model={model}
+          item={item}
+          onSubmit={({ item: submittedItem }) => {
+            setItem(submittedItem);
+            // Trigger grid data refresh
+            setRefreshTrigger(refreshTrigger + 1);
```
taefi, line 52: Even though it works, it looks hacky :)
Maybe, we could at least hide the logic in a method called `refresh` and call it here.

packages/ts/react-grid/src/autocrud.tsx
```diff
@@ -0,0 +1,58 @@
...
+              setItem({ ...selectedItem });
+            }
+          }}
+        ></AutoGrid>
+        <ExperimentalAutoForm
+          disabled={!item}
+          service={service}
+          model={model}
+          item={item}
+          onSubmit={({ item: submittedItem }) => {
```
taefi, line 49: This is called `onSubmit` but it seems to be more of an `afterSubmit`.

packages/ts/react-grid/src/autocrud.tsx
taefi: Shouldn't there be a `refresh`/`reload` function defined in `AutoGrid` and exposed for `AutoCrud` to call? However, in the end, that function should change some state in the `AutoGrid` so that the dependent effect is triggered.

packages/ts/react-grid/test/grid-test-helpers.ts
```diff
@@ -29,6 +33,31 @@ export const getHeaderCell = <T>(grid: Grid<T>, row: number, col: number): HTMLE
...
+      resolve(undefined);
+    });
+  });
+}
+
+export async function setActiveItem<T>(grid: Grid<T>, item: T | undefined): Promise<void> {
+  grid.activeItem = item;
+  await nextFrame();
+  await nextFrame();
+  await nextFrame();
```
taefi, line 52: What is the reason of `nextFrame` getting called here three times, and in many other places two time consecutively? What is it doing exactly?

taefi (changes requested):

packages/ts/react-grid/src/autocrud.tsx
```diff
@@ -0,0 +1,58 @@
...
+}>;
+
+interface AutoCrudContextType<TItem> {
+  item: TItem | undefined;
+}
+
+export const AutoCrudContext = createContext<AutoCrudContextType<any>>({ item: undefined });
+
+export function ExperimentalAutoCrud<TItem>({ service, model }: AutoCrudProps<TItem>): JSX.Element {
+  const [item, setItem] = useState<TItem | undefined>(undefined);
```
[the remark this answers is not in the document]
Artur-: What do you want to distinguish it from? To me it is the `item` that is being used by the CRUD. As grid uses both `selectedItem` and `activeItem` we should not confuse it with those. `currentItem` could work but that sounds a bit like there would be a `previousItem` also? It could be `currentItem`

packages/ts/react-grid/src/autocrud.tsx
Artur-: The next PR for delete introduces a context with a `refreshGrid` method but this would require some more refactoring then to be able to use the context. Can we postpone it a bit so it will be easier? It will probably be necessary to do the refactoring to support customization of the form and/or layout

packages/ts/react-grid/test/grid-test-helpers.ts
Artur-: It is a bad way of waiting for React to finish rendering. It basically waits for the next frame in the browser, using `requestAnimationFrame`. Sometimes React renders in two cycles, sometimes in more. It needs to be fixed in a completely different way in the future

packages/ts/react-grid/src/autocrud.tsx
Artur-: Yes, renamed

packages/ts/react-grid/src/autocrud.tsx
Artur-: > Shouldn't there be a refresh/reload function defined in AutoGrid and exposed for AutoCrud to call
As we are talking React, a component cannot really publish API...

Artur-: @sissbruecker  had a prototype for a different approach that could be added later also

packages/ts/react-grid/test/grid-test-helpers.ts
taefi: Is it possible to introduce a function e.g. called `waitForRender` to wrap two or three `nextFrame()` calls? Would it break the tests if there is a need for two cycles and we wait for three cycles?

Artur-: Waiting more is just … waiting more. I would not wrap them though but instead use act() or waitFor()

Artur-: Created #1418 for this

packages/ts/react-grid/src/autocrud.tsx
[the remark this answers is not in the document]
Artur-: As `AutoForm` also uses `item`, I would not rename it as part of this PR at least

packages/ts/react-grid/src/autocrud.tsx
taefi: Isn't there a way to avoid reloading the grid data from backend? For example, getting rid of that `refreshTrigger` state and pass the newly added/edited item and adding/updating that one directly into the existing data? Maybe, we don't want to get rid of `refreshTrigger` completely for cases that needs a reload to happen, but in this case it feels like an overhead.

Artur-: Reloading the data is a good first approach. It can always be improved although it requires quite a bit more code. Also by reloading you get changes done by other people. Create an issue and we won't forget it

packages/ts/react-grid/src/autocrud.tsx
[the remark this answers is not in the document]
taefi: The way I was picturing the AutoForm or a form in general, is a navigable form with <kbd> << </kbd> <kbd> < </kbd> [ x of n record(s) ] <kbd> > </kbd><kbd> >> </kbd> (pagination controls) so that I can move back and forward without relying on grid. 
Let's think about having the `currentItem` in that context, but not rename it here.

taefi (approved):

PR #1357, opened 2023-09-26
feat: add metadata to model options

sissbruecker:
## Description

Adds an optional metadata property to the model options passed to the model constructor. The metadata can contain:
- The java type name
- A list of annotations

Closes https://github.com/vaadin/hilla/issues/1271
Part of https://github.com/vaadin/hilla/issues/1266

### DIFF
--- a/packages/ts/generator-typescript-plugin-model/src/MetadataParser.ts
+++ b/packages/ts/generator-typescript-plugin-model/src/MetadataParser.ts
@@ -1,0 +1,57 @@
+import type { Schema } from '@hilla/generator-typescript-core/Schema.js';
+import ts, { type ObjectLiteralExpression, type PropertyAssignment } from 'typescript';
+
+interface Annotation {
+  name: string;
+  attributes?: Record<string, unknown>;
+}
+
+export type SchemaWithMetadata = Schema & {
+  'x-annotations'?: Annotation[];
+  'x-java-type'?: string;
+};
+
+export class MetadataParser {
+  parse(schema: Schema): ObjectLiteralExpression | null {
+    const schemaWithMetadata = schema as SchemaWithMetadata;
+
+    const properties = [
+      this.#createAnnotationsProperty(schemaWithMetadata),
+      this.#createJavaTypeProperty(schemaWithMetadata),
+    ].filter(Boolean) as PropertyAssignment[];
+
+    if (properties.length === 0) {
+      return null;
+    }
+
+    return ts.factory.createObjectLiteralExpression(properties);
+  }
+
+  #createAnnotationsProperty(schema: SchemaWithMetadata): PropertyAssignment | null {
+    const annotations = schema['x-annotations'];
+    const hasAnnotations = annotations && annotations.length > 0;
+    if (!hasAnnotations) {
+      return null;
+    }
+
+    const annotationLiterals = annotations.map((annotation) =>
+      ts.factory.createObjectLiteralExpression([
+        ts.factory.createPropertyAssignment('name', ts.factory.createStringLiteral(annotation.name)),
+      ]),
+    );
+
+    return ts.factory.createPropertyAssignment(
+      'annotations',
+      ts.factory.createArrayLiteralExpression(annotationLiterals),
+    );
+  }
+
+  #createJavaTypeProperty(schema: SchemaWithMetadata): PropertyAssignment | null {
+    const javaType = schema['x-java-type'];
+    if (!javaType) {
+      return null;
+    }
+
+    return ts.factory.createPropertyAssignment('javaType', ts.factory.createStringLiteral(javaType));
+  }
+}
--- a/packages/ts/generator-typescript-plugin-model/src/ModelSchemaProcessor.ts
+++ b/packages/ts/generator-typescript-plugin-model/src/ModelSchemaProcessor.ts
@@ -28,7 +28,7 @@ import ts, {
   type TypeNode,
   type TypeReferenceNode,
 } from 'typescript';
-import { MetadataParser } from './MetadataParser';
+import { MetadataParser } from './MetadataParser.js';
 import { createModelBuildingCallback, importBuiltInFormModel } from './utils.js';
 import { hasValidationConstraints, ValidationConstraintParser } from './ValidationConstraintParser.js';
 
--- a/packages/ts/generator-typescript-plugin-model/src/ModelSchemaProcessor.ts
+++ b/packages/ts/generator-typescript-plugin-model/src/ModelSchemaProcessor.ts
@@ -295,6 +295,6 @@ export class ModelSchemaExpressionProcessor extends ModelSchemaPartProcessor<rea
 
   #createMetadataProperty(schema: Schema): PropertyAssignment | null {
     const metadata = this.#metadataProcessor.process(schema);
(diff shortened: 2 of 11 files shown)


packages/ts/generator-typescript-plugin-model/src/MetadataParser.ts
```diff
@@ -0,0 +1,57 @@
...
+    const hasAnnotations = annotations && annotations.length > 0;
+    if (!hasAnnotations) {
+      return null;
+    }
+
+    const annotationLiterals = annotations.map((annotation) =>
+      ts.factory.createObjectLiteralExpression([
+        ts.factory.createPropertyAssignment('name', ts.factory.createStringLiteral(annotation.name)),
+      ]),
+    );
```
sissbruecker, line 41: Annotations can also have a generic object of attributes, but it wasn't clear how I could get those generated. There is something similar in the validation constraints parser but I wasn't able to adapt that so far. I would suggest to start without the attributes for now, as we don't have a use-case for them at the moment.

packages/ts/generator-typescript-plugin-model/src/ValidationConstraintParser.ts
sissbruecker, line 1: Renamed the file, class and types to reflect that this parses validation constraints rather than generic annotations.

packages/ts/generator-typescript-plugin-model/src/ValidationConstraintParser.ts
```diff
@@ -13,51 +13,45 @@ import ts, {
...
-  parse(annotation: Annotation | string): ts.NewExpression {
-    if (typeof annotation === 'string') {
-      const nameEndIndex = annotation.indexOf('(');
-      const simpleName = nameEndIndex >= 0 ? annotation.slice(0, nameEndIndex) : annotation;
-      const id = this.#importer(simpleName);
-
-      return template<NewExpression>(`${variableStatementVar} = new ${annotation}`, selector, [
-        transform((node) => (ts.isIdentifier(node) && node.text === simpleName ? id : node)),
-      ]);
-    }
```
sissbruecker, line 51: This is some left-over legacy code related to what has been removed in https://github.com/vaadin/hilla/pull/1315

packages/ts/generator-typescript-plugin-model/src/MetadataParser.ts
Lodin: Aren't they gonna be put in the same object? Like:
```ts
{
  name: "jakarta.persistense.Id",
  param1: 0,
  param2: "string",
}
```

packages/ts/generator-typescript-plugin-model/src/MetadataParser.ts
```diff
@@ -0,0 +1,57 @@
...
+  name: string;
+  attributes?: Record<string, unknown>;
+}
+
+export type SchemaWithMetadata = Schema & {
+  'x-annotations'?: Annotation[];
+  'x-java-type'?: string;
+};
+
+export class MetadataParser {
```
Lodin, line 14:
```suggestion
export class MetadataProcessor {
```
The `parser` parts existed just because we needed to parse a string representation of validation constraints. Now we get rid of it, so let's have it consistent.

packages/ts/generator-typescript-plugin-model/src/ModelSchemaProcessor.ts
```diff
@@ -272,7 +282,19 @@ export class ModelSchemaExpressionProcessor extends ModelSchemaPartProcessor<rea
...
+
+    const constraints = schema['x-validation-constraints'].map((constraint) =>
+      this.#validationConstraintParser.parse(constraint),
+    );
+    return ts.factory.createPropertyAssignment('validators', ts.factory.createArrayLiteralExpression(constraints));
+  }
+
+  #createMetadataProperty(schema: Schema): PropertyAssignment | null {
+    const metadata = this.#metadataParser.parse(schema);
+    return metadata ? ts.factory.createPropertyAssignment('metadata', metadata) : null;
```
Lodin, line 298: Let's shrink the name a bit to have less code in the output.
```suggestion
    return metadata ? ts.factory.createPropertyAssignment('meta', metadata) : null;
```

packages/ts/generator-typescript-plugin-model/src/ValidationConstraintParser.ts
```diff
@@ -13,51 +13,45 @@ import ts, {
...
 
 function selector<T extends Expression>([statement]: readonly Statement[]): T {
   return (statement as VariableStatement).declarationList.declarations[0].initializer as T;
 }
 
 const variableStatementVar = 'const a';
 
-export class AnnotationParser {
-  readonly #importer: AnnotationImporter;
+export class ValidationConstraintParser {
```
Lodin, line 39:
```suggestion
export class ValidationConstraintProcessor {
```

packages/ts/generator-typescript-plugin-model/test/model/AnnotationParser.spec.ts
```diff
@@ -24,13 +24,13 @@ function assertAnnotation(actual: NewExpression, expected: string): void {
   expect(printer.printFile(file).trim()).to.equal(`const a = new ${expected};`);
 }
 
-describe('AnnotationParser', () => {
+describe('ValidationConstraintParser', () => {
```
Lodin, line 27:
```suggestion
describe('ValidationConstraintProcessor', () => {
```

Lodin (changes requested):

packages/ts/generator-typescript-plugin-model/src/MetadataParser.ts
sissbruecker: Done

packages/ts/generator-typescript-plugin-model/src/ModelSchemaProcessor.ts
sissbruecker: Done

packages/ts/generator-typescript-plugin-model/src/ValidationConstraintParser.ts
sissbruecker: Done

packages/ts/generator-typescript-plugin-model/test/model/AnnotationParser.spec.ts
sissbruecker: Done

packages/ts/generator-typescript-plugin-model/src/MetadataParser.ts
Lodin: I usually use this [site](https://ts-ast-viewer.com) to understand what calls to use if I want to create a specific AST. 

For the object literal, the code should look similar to that one:

```ts
ts.factory.createPropertyAssignment(
  ts.factory.createIdentifier("attributes"),
  ts.factory.createObjectLiteralExpression(
    Object.entries(attributes).map(([name, value]) => ts.factory.createPropertyAssignment(
      ts.factory.createIdentifier(name),
      // ...create specific literal for value, e.g.:
      ts.factory.createStringLiteral(value)
    )),
    true
  )
)
```

Lodin (approved):

PR #1359, opened 2023-09-26
feat: Sort by default according to the first column

### LINKED ISSUE #1265: [autogrid] Default sorting
An `<AutoGrid>` should be sorted by default. As we currently have no info about what would be a good column to sort on, we should use the first one

Artur-:
Fixes #1265

### DIFF
--- a/packages/ts/react-grid/src/autogrid.tsx
+++ b/packages/ts/react-grid/src/autogrid.tsx
@@ -10,7 +10,7 @@ import {
 } from '@hilla/react-components/Grid.js';
 import { GridColumn } from '@hilla/react-components/GridColumn.js';
 import { GridColumnGroup } from '@hilla/react-components/GridColumnGroup.js';
-import { GridSorter } from '@hilla/react-components/GridSorter.js';
+import { GridSorterElement } from '@hilla/react-components/GridSorter.js';
 import { useCallback, useEffect, useRef, useState, type JSX } from 'react';
 import type { CrudService } from './crud';
 import { HeaderColumnContext } from './header-column-context';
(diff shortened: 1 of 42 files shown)


packages/ts/react-grid/src/autogrid.tsx
```diff
@@ -182,7 +185,9 @@ export function AutoGrid<TItem>({
   useEffect(() => {
     // Sets the data provider, should be done only once
     const grid = ref.current!;
-    grid.dataProvider = createDataProvider(grid, service, dataProviderFilter);
+    setTimeout(() => {
```
sissbruecker, line 188: Could add a comment to explain the timeout.

packages/ts/react-grid/test/autogrid.spec.tsx
```diff
@@ -38,52 +45,72 @@ async function assertColumns(result: RenderResult, ...ids: string[]) {
...
+      result.rerender(<TestAutoGrid />);
+      expect(getSortOrder(grid)).to.eql({ path: 'lastName', direction: 'desc' });
+    });
+    it('creates sortable columns', async () => {
+      const result = render(<TestAutoGrid />);
+      const grid: GridElement = result.container.querySelector('vaadin-grid')!;
+      await nextFrame();
+      await nextFrame();
+      const firstNameSorter = getHeaderCellContent(grid, 0, 0).firstElementChild as HTMLElement;
       firstNameSorter.click();
```
sissbruecker, line 85: Can reuse `sortGrid` here.

Artur-: Done

packages/ts/react-grid/src/autogrid.tsx
Artur-: Added

sissbruecker (approved):

PR #1362, opened 2023-09-27
feat: Render number columns as right aligned numbers

### LINKED ISSUE #1267: [autogrid] Format and align numbers by default
Numbers should be formatted using something like
```
new Intl.NumberFormat(getLocale(), {
    maximumFractionDigits: 0,
  }).format(value);
```

The column where numbers are shown should be 
```
  textAlign: "end",
  flexGrow: 0,
  autoWidth: true,
```
and use
```
  font-variant-numeric: tabular-nums;
```

Artur-:
Fixes #1267

### DIFF
--- a/packages/ts/react-grid/src/autogrid-column.tsx
+++ b/packages/ts/react-grid/src/autogrid-column.tsx
@@ -1,22 +1,22 @@
+import type { GridItemModel } from '@hilla/react-components/Grid.js';
+import type { GridColumnElement } from '@hilla/react-components/GridColumn.js';
 import { useContext } from 'react';
 import { ColumnContext } from './header-column-context';
-import { GridItemModel } from '@hilla/react-components/Grid.js';
-import { GridColumnElement } from '@hilla/react-components/GridColumn.js';
 
 type RendererOptions<TItem> = {
   item: TItem;
   model: GridItemModel<TItem>;
   original: GridColumnElement<TItem>;
 };
-export function AutoGridColumn<TItem>({ item, model, original }: RendererOptions<TItem>) {
+export function AutoGridColumn<TItem>({ item, model, original }: RendererOptions<TItem>): JSX.Element {
   const context = useContext(ColumnContext)!;
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
   const value = (item as any)[context.propertyInfo.name];
-  if (context?.propertyInfo.modelType === 'number') {
+  if (context.propertyInfo.modelType === 'number') {
     const formatted = new Intl.NumberFormat(undefined, {
       maximumFractionDigits: 0,
     }).format(value);
     return <span style={{ fontVariantNumeric: 'tabular-nums' }}>{formatted}</span>;
-  } else {
-    return <>{value}</>;
   }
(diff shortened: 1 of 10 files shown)


packages/ts/react-grid/src/autogrid-column.tsx
sissbruecker, line 1: Can we use a different component per column type? In the boolean renderer PR it already starts getting messy.

Let's also rename this module and the component(s) to use `renderer` rather than `column`.

Artur-: We can. Can we somehow handle column properties and rendering in the same file?

sissbruecker: I was wondering the same. Since the renderer is just another column option / prop, maybe have a factory function that returns a props object containing all options.

Artur-: One way would be to include both `gridRenderer` and `gridColumnOptions` in `PropertyInfo`

Artur-: Updated to this model, what say you?

sissbruecker: So something like:
```ts
export function getColumnProps<TItem>(propertyInfo: PropertyInfo): Partial<GridColumnProps<TItem>> {
  if (propertyInfo.modelType === 'number') {
    return {
      textAlign: 'end',
      flexGrow: 0,
      renderer: NumberRenderer,
    };
  } else if (propertyInfo.modelType === 'boolean') {
    return {
      renderer: BooleanRenderer,
    };
  }

  return {
    renderer: DefaultRenderer,
  };
}
```

Artur-: Do you think we should separate it out from PropertyInfo?

Artur-: Probably we should, let's see what I get

Artur-: There, much better

packages/ts/react-grid/src/autogrid-number-renderer.tsx
sissbruecker, line 1: IMO it would be OK to put future renderers in the same module.

packages/ts/react-grid/src/utils.ts
```diff
@@ -7,6 +7,10 @@ import {
   NumberModel,
   type ModelMetadata,
 } from '@hilla/form';
+import type { GridItemModel } from '@hilla/react-components/Grid.js';
+import type { GridColumnElement, GridColumnProps } from '@hilla/react-components/GridColumn.js';
+import type { ComponentType } from 'react';
+import { AutoGridNumberRenderer } from './autogrid-number-renderer';
```
sissbruecker, line 13: Unused imports

sissbruecker (approved):

packages/ts/react-grid/src/autogrid-number-renderer.tsx
Artur-: Let's see in the boolean PR

PR #1371, opened 2023-09-27
feat: always scan the root package of current module/project for endpoints

### LINKED ISSUE #994: Consider always scaning current project/module root package for endpoints

taefi:
## Description

This is a fixed-term solution for always including the root package of current module/project while scanning for endpoints.

Finding the root package of the current module can be done in different ways, and in this PR it is relying on the existence of a class annotated with `@SpringBootApplication`.

The better approach for scanning the endpoints could rely on spring beans instead, however, at the moment making such a breaking change is not possible, so until then, this can be considered as a partial solution.

Fixes #994

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


Artur-:
Why are there quite a few pom.xml changes here? They don't seem to be related to the actual change

taefi:
> Why are there quite a few pom.xml changes here? They don't seem to be related to the actual change

I had some random failures on CI showing symptoms of not performing prepare-frontend in dev-mode or the build-frontend in prod mode, so I added them in those modules.

Artur-:
Can we make this only the Java code change  and see if the problems have been resolved by other PRs?

taefi:
Updating this one was harder than creating a new PR. Closing in favor of #1453

PR #1373, opened 2023-09-28
chore: Run tests in development mode by default

Artur-:
Production mode can be enabled using -Pproduction


packages/java/tests/pom.xml
```diff
@@ -1,288 +1,311 @@
...
+                <plugin>
+                    <artifactId>maven-clean-plugin</artifactId>
+                    <version>3.2.0</version>
+                    <configuration>
+                        <filesets>
+                            <fileset>
+                                <directory>${project.basedir}</directory>
+                                <includes>
+                                    <include>node_modules/**</include>
+                                    <include>types.d.ts</include>
```
taefi, line 170: Is the default `types.d.ts` file generated by flow working for React apps nowadays?

packages/java/tests/pom.xml
```diff
@@ -1,288 +1,311 @@
...
+                <pluginManagement>
+                    <plugins>
+                        <plugin>
+                            <groupId>dev.hilla</groupId>
+                            <artifactId>hilla-maven-plugin</artifactId>
+                            <version>${project.version}</version>
+                            <executions>
+                                <execution>
+                                    <goals>
+                                        <goal>prepare-frontend</goal>
```
taefi, line 263:
```suggestion
```

taefi (changes requested):

packages/java/tests/pom.xml
Artur-: Done

packages/java/tests/pom.xml
Artur-: Probably not but this part is not changed in this PR

.github/workflows/validation.yml
```diff
@@ -199,6 +199,7 @@ jobs:
               -Dmaven.wagon.http.retryHandler.count=3 \
               -P\!it-modules \
               -Pit-modules-${{ matrix.it-modules }} \
+              -Pproduction \
```
taefi, line 202: Is the PR title wrong or this profile activation?

taefi (changes requested):

.github/workflows/validation.yml
Artur-: Well the intent is to run them in dev mode when you run locally, so you can develop, and in production mode in CI so it is faster

taefi (approved):

taefi (approved):

Artur-:
The validation build will not pass here as it uses the old configuration

PR #1377, opened 2023-09-28
test: Fix test for default sort order

Artur-:


sissbruecker (approved):

Artur-:
#1374 also fixes this so we can merge that one

Issue #1378, opened 2023-09-28
[autogrid] header filter support for boolean columns


Artur-:
When the column contains booleans, the header filter should be a select with the same icons as used for rendering. Additionally it should have an empty option (default) to show all

PR #1385, opened 2023-09-28
feat: boolean header filters

sissbruecker:
Adds header filters for boolean columns.

Fixes https://github.com/vaadin/hilla/issues/1378


packages/ts/react-grid/src/header-filter.tsx
```diff
@@ -28,59 +27,100 @@ export function HeaderFilter(): ReactElement {
...
+      renderer={() => (
+        <ListBox>
+          <Item value={''} {...{ label: '' }}></Item>
+          <Item value={'True'} {...{ label: 'Yes' }}>
+            Yes
+          </Item>
+          <Item value={'False'} {...{ label: 'No' }}>
+            No
+          </Item>
+        </ListBox>
```
sissbruecker, line 121: Showing icons next to the checkmark that comes with list box items looked a bit strange. Open for suggestions.

Artur-:
This renders a label for all columns with an unsupported type
image

packages/ts/react-grid/src/header-filter.tsx
Artur-: Indeed, "yes", "no" works much better

PR #1395, opened 2023-09-29
test: remove parallel switch from ts tests to reduce concurrency issues

taefi:


Artur-:
Is there any particular case this should solve?

Artur- (approved):

PR #1397, opened 2023-09-29
chore: Split out gradle tests

Artur-:


taefi:
Doesn't this PR need another dummy PR that targets the `split-gradle` branch so that the validation.yml changes are actually be tested in that dummy PR?

Artur-:
That is #1396

taefi (approved):

PR #1399, opened 2023-09-29
test: Spam less in the build logs from security tests

Artur-:


Artur-:
This is a part of #1373 split out to another PR

PR #1400, opened 2023-09-29
fix(form): errors when defaultValue is set with array items

### LINKED ISSUE #1294: Binder looses binderNode.defaultValue for nested keys inside repeat() in Hilla 2.2.0
### Description of the bug 
Binder looses binderNode.defaultValue for nested keys inside repeat() when rendering array model
The array model passed to Binder:
```

[
    {
        "organization": {
            "address": {
                "building": "",
                "city": "",
                "country": "",
                "district": "",
                "note": "",
                "region": "",
                "room": "",
                "settlement": "",
                "street": ""
            }
        }
    }
]

```
This is how binder is initialized:
```

const declarant = TDeclarantModel.createEmptyValue();
declarant.organization = TOrganizationModel.createEmptyValue();
declarant.organization.address = TAddressModel.createEmptyValue();
this.binder.value.projectDocumentsDeveloper = [declarant];

```
If BinderNode has a nested keys ('address' in example), default values of nested keys are undefined.
### Expected behavior
BinderNode.defaultValue should looks like:

```
{
    "organization": {
        "address": {
            "building": "",
            "city": "",
            "country": "",
            "district": "",
            "note": "",
            "region": "",
            "room": "",
            "settlement": "",
            "street": ""
        }
    }
} 

```
### Actual behavior
The BinderNode.defaultValus is:

{
      organization: undefined
}

However, BinderNode.value is correct ('organization' is an object as expected)

So, I get en error:
```

Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[ConclusionView.update()]' TypeError: Cannot read properties of undefined (reading 'address')
    at get defaultValue [as defaultValue] (BinderNode.ts:153:26)
    at g2.initializeValue (BinderNode.ts:422:73)
    at new g2 (BinderNode.ts:94:10)
    at b2 (Models.ts:23:26) 
```

### Versions:
    - Vaadin / Hilla version: 2.2.0
    - Node version: 18.7
    - Java version: 17
    - OS version: Windows 10
    - Browser version (if applicable): Chrome
    - Application Server (if applicable):
    - IDE/Editor (if applicable):

platosha:
In forms that have arrays, when `binder.defaultValue` contains array items (for example, when initialized using `.read()`), having deep bindings with array item object was causing an error. The reason is that the binder failed to change the user-defined default value for the array item when initializing deep objects inside the item.

This change makes the binder maintain the user-defined array-item default value applying the same approach to arrays as when initializing optional keys in objects. When it needs to initialize a deep object inside the array item, it automatically changes the value and defaultValue of the array item at the same time, so that both get the deep keys in this case.

Fixes #1294

### DIFF
--- a/packages/ts/form/src/BinderNode.ts
+++ b/packages/ts/form/src/BinderNode.ts
@@ -26,6 +26,7 @@ import {
   AbstractModel,
   ArrayModel,
   getBinderNode,
+  type ModelValue,
   ObjectModel,
 } from './Models.js';
 import type { Validator, ValueError } from './Validation.js';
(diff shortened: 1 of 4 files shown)


Artur-:
Why is this targeting 2.2 only?

platosha:
> Why is this targeting 2.2 only?

image

cromoteca (approved):
Looks more solid now, with the `value` and `defaultValue` tied together.

PR #1401, opened 2023-09-29
fix(form): inference for generic binder.model type

platosha:
Fixes TypeScript type inference for `binder.model` that was broken after #1252


cromoteca (dismissed):
LGTM, I leave you decide if you want to address the CheckStyle remarks or not. Everything seems to be related to `any` which is maybe hard to avoid.

platosha:
> LGTM, I leave you decide if you want to address the CheckStyle remarks or not. Everything seems to be related to `any` which is maybe hard to avoid.

I'll try to do something about it. `any` is clearly not correct, it should infer the data type.

cromoteca (approved):

PR #1407, opened 2023-10-02
fix: Use the same Typescript version as Flow

Artur-:
This should fix problems with multiple TypeScript versions being mixed and TypeScript 5.2 installed in node_modules/@hilla/something/node_modules

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "simple-git-hooks": "^2.9.0",
     "sync-request": "^6.1.0",
     "tsx": "^3.13.0",
-    "typescript": "^5.2.2",
+    "typescript": "5.1.6",
     "vite": "^4.4.9"
   }
 }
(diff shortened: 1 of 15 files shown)


Lodin (changes requested):
Let's have the same TSC version for `generator-typescript-*` packages then

Artur-:
Updated

package.json
```diff
@@ -58,7 +58,7 @@
     "simple-git-hooks": "^2.9.0",
     "sync-request": "^6.1.0",
     "tsx": "^3.13.0",
-    "typescript": "5.1.6",
+    "typescript": "^5.1.6",
```
Lodin, line 61: Let's not have `^` here because it will install `5.2.2` on install without `package-lock.json`.

Lodin (changes requested):

package.json
Artur-: Who put that there...

Artur-: Fixed

Lodin (approved):

PR #1409, opened 2023-10-02
chore: Fix test summary to depend on all other tasks

Artur-:


.github/workflows/validation.yml
```diff
@@ -282,6 +282,10 @@ jobs:
     if: ${{ always() }}
     needs:
       - test-it
+      - test-java
+      - test-typescript
+      - test-it
```
cromoteca, line 287: This is duplicate

Artur-: True that
True that

Artur-:
Let's check that it looks correct in #1434

Artur-:
looks correct to me 
image

manolo (approved):

PR #1442, opened 2023-10-03
feat(autogrid): add header filter for dates and times

### LINKED ISSUE #1285: [autogrid] header filter support for date columns
When the column contains dates, the header filter should be a combination of a select and a date picker
image

image

The select contains options for how to match the value (>, = or <), the text field value is used for matching

The above image is captured from the code
```jsx
          <Select
            renderer={() => (
              <ListBox>
                <Item value="GREATER_THAN" {...{ label: ">" }}>
                  &gt; Greater than
                </Item>
                <Item value="LESS_THAN" {...{ label: "<" }}>
                  &lt; Less than
                </Item>
                <Item value="EQUALS" {...{ label: "=" }}>
                  = Equals
                </Item>
              </ListBox>
            )}
            className={css.matchSelect}
            value="GREATER_THAN"
          ></Select>
          <DatePicker></DatePicker>
```

and

```css
.matchSelect {
  --vaadin-field-default-width: 2em;
  margin-right: 3px;
}
.matchSelect > vaadin-select-value-button {
  --_lumo-text-field-overflow-mask-image: none !important;
}

.matchSelect::part(toggle-button) {
  display: none;
}
```

Note that both date, time and date + time formats should be supported the same way

cromoteca:
Adds filtering to headers in date and time columns.

Closes #1285

### DIFF
--- a/packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
@@ -1,6 +1,8 @@
 package dev.hilla.crud;
 
 import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
 
 import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
@@ -93,6 +95,38 @@ public class PropertyStringFilterSpecification<T> implements Specification<T> {
             default:
                 break;
             }
+        } else if (isLocalTime(javaType)) {
+            var path = root.<LocalTime> get(filter.getPropertyId());
+            var timeValue = LocalTime.parse(value);
+            switch (filter.getMatcher()) {
+            case EQUALS:
+                return criteriaBuilder.equal(path, timeValue);
+            case CONTAINS:
+                throw new IllegalArgumentException(
+                        "A time cannot be filtered using contains");
+            case GREATER_THAN:
+                return criteriaBuilder.greaterThan(path, timeValue);
+            case LESS_THAN:
+                return criteriaBuilder.lessThan(path, timeValue);
+            default:
+                break;
+            }
+        } else if (isLocalDateTime(javaType)) {
+            var path = root.<LocalDateTime> get(filter.getPropertyId());
+            var dateTimeValue = LocalDateTime.parse(value);
+            switch (filter.getMatcher()) {
+            case EQUALS:
+                return criteriaBuilder.equal(path, dateTimeValue);
+            case CONTAINS:
+                throw new IllegalArgumentException(
+                        "A datetime cannot be filtered using contains");
+            case GREATER_THAN:
+                return criteriaBuilder.greaterThan(path, dateTimeValue);
+            case LESS_THAN:
+                return criteriaBuilder.lessThan(path, dateTimeValue);
+            default:
+                break;
+            }
         }
         throw new IllegalArgumentException("No implementation for " + javaType
                 + " using " + filter.getMatcher() + ".");
@@ -112,4 +146,11 @@ public class PropertyStringFilterSpecification<T> implements Specification<T> {
         return javaType == java.time.LocalDate.class;
     }
 
+    private boolean isLocalTime(Class<?> javaType) {
+        return javaType == java.time.LocalTime.class;
+    }
+
+    private boolean isLocalDateTime(Class<?> javaType) {
+        return javaType == java.time.LocalDateTime.class;
+    }
 }
--- a/packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
@@ -1,5 +1,7 @@
 package dev.hilla.crud;
 
+import java.time.LocalDate;
+
 import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.Expression;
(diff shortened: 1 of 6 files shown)


packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
```diff
@@ -90,4 +144,15 @@ private boolean isBoolean(Class<?> javaType) {
...
+        return javaType == java.time.LocalDate.class;
+    }
+
+    private boolean isLocalTime(Class<?> javaType) {
+        return javaType == java.time.LocalTime.class;
+    }
+
+    private boolean isLocalDateTime(Class<?> javaType) {
+        return javaType == java.time.LocalDateTime.class;
+    }
```
taefi, line 157: No support for other types that represent date and time? 
* java.sql.Date (supporting old projects that need modernization in UI)
* ZonedDateTime
* Timestamp (in many datebases this is favoured over the DATE and DATETIME db types because of being more performant in indexing, searching, storage, etc.)

packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
```diff
@@ -75,6 +79,56 @@ public Predicate toPredicate(Root<T> root, CriteriaQuery<?> query,
...
+            switch (filter.getMatcher()) {
+            case EQUALS:
+                return criteriaBuilder.equal(path, dateValue);
+            case CONTAINS:
+                throw new IllegalArgumentException(
+                        "A date cannot be filtered using contains");
+            case GREATER_THAN:
+                return criteriaBuilder.greaterThan(path, dateValue);
+            case LESS_THAN:
+                return criteriaBuilder.lessThan(path, dateValue);
```
taefi, line 94: For many real-world applications, the operator "between" or a combination of greater_than and less_than is the default for filtering date values. Also, it is vague that whether these two are inclusive or exclusive.

packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
```diff
@@ -75,6 +79,56 @@ public Predicate toPredicate(Root<T> root, CriteriaQuery<?> query,
             default:
                 break;
             }
+        } else if (isLocalDate(javaType)) {
```
taefi, line 82: Not specifically related to this PR, but, are these `if-else`s only hurting my eyes or is it an epidemic?

taefi (changes requested):

packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
Artur-: If these are needed, separate issues should be created

packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
Artur-: That could be improved, through another issue

cromoteca: We're talking about a header filter, there's no room for complex UI bits there

packages/ts/react-grid/src/header-filter.tsx
```diff
@@ -140,6 +158,53 @@ export function BooleanHeaderFilter(): ReactElement {
...
 }
+
+export function DateHeaderFilter(): ReactElement {
+  const { matcher, filterValue, updateFilter } = useFilterState(Matcher.GREATER_THAN);
+  const [invalid, setInvalid] = useState(false);
+
+  return (
+    <>
+      <ComparationSelection value={matcher} onMatcherChanged={(m) => updateFilter(m, filterValue)} />
+      <DatePicker
```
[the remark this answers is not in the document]
cromoteca: Is this supported in the grid? If so, this must be modified to match the same locale as the grid, definitely.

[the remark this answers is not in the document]
sissbruecker: Yes, dates in cells are formatted according to the current browser locale. For the date picker, we can configure custom functions for formatting and parsing dates. The formatting part is easy with the `Intl.DateTimeFormat` browser API, however there is no equivalent for parsing.

I did some prototyping yesterday to implement parsing based on the logic that exists in the Flow component. Can continue with that today.

packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
taefi: I'm not sure having support for between is a luxury just because it is header filtering. In fact, by default it should show a special control in header title not just a DatePicker (or a DateTimePicker) which by clicking on that it shows a popup containing two DatePicker (or the other variants). Then based on whether the user has chosen a value for either one or both of them, it should do the filtering by greater_than, less_than or between. This can be a future improvement of course, but it is not something extra ordinary.

taefi: But, still the answer to the other question is not clear, how the user or even the developer can indicate whether it is performing a greater_than (exclusive) or greater_than_or_equals_to (inclusive). Should it be indicated with a sign, icon, tooltip, etc.?

taefi: For the between case, I created this: https://github.com/vaadin/hilla/issues/1463

packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
taefi: https://github.com/vaadin/hilla/issues/1464

packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
taefi: It seems that is it only hurting my eyes :)

taefi (approved):

PR #1443, opened 2023-10-03
fix(form): errors when defaultValue is set with array items (#1400)

platosha:
* Create test for the issue

* fix(form): errors when defaultValue is set with array items

---------

(cherry picked from commit 6b0bfd3d6f4affc77cc036972e0fe05e76052c93)


cromoteca (approved):
I haven't tested it in a real application, but given that the test passes, it should be fine.

PR #1445, opened 2023-10-03
fix(auto-grid): remove css file import

platosha:
Non-JS imports break unpredictably depending on the user frontend build
system. One example is the Hilla docs website, which uses webpack, which
does not support CSS modules.


packages/ts/react-grid/src/header-filter.tsx
```diff
@@ -5,9 +5,24 @@ import { Select, type SelectElement } from '@hilla/react-components/Select.js';
 import { TextField, type TextFieldElement } from '@hilla/react-components/TextField.js';
 import { type ReactElement, type RefObject, useContext, useEffect, useRef, useState } from 'react';
 import { ColumnContext } from './autogrid-column-context.js';
-import css from './header-filter.module.css';
 import Matcher from './types/dev/hilla/crud/filter/PropertyStringFilter/Matcher';
 
+const autoGridFilterWithLessGreaterEqualsStyle = document.createElement('style');
```
Lodin, line 10: Let's have a `TODO` mark here to remember that we would like to replace it later.
```suggestion
// TODO: Replace with more robust solution
const autoGridFilterWithLessGreaterEqualsStyle = document.createElement('style');
```

Lodin (changes requested):

packages/ts/react-grid/src/header-filter.module.css
sissbruecker, line 1: Could also update `files` in `package.json` to not publish CSS modules.

sissbruecker (approved):

Issue #1404, opened 2023-10-01
[autogrid] Add CRUD methods to CrudService and CrudRepositoryService


tarekoraby:
### Describe your motivation

Even without an AutoCrud component in the frontend, add CRUD methods to `CrudService` and `CrudRepositoryService` because

1. It's weird to include "crud" in those class/interface names when they only do reading and filtering
2. People can still benefit from the full creation of a CRUD service/endpoint from a JPA repo. That would save some LoC from the backend.

Artur-:
Anything you are missing?

```java
public interface ListService<T> {
    /**
     * Lists objects of the given type using the paging and sorting options
     * provided in the parameter.
     *
     * @param pageable
     *            contains information about paging and sorting
     * @param filter
     *            the filter to apply or {@code null} to not filter
     * @return a list of objects or an empty list if no objects were found
     */
    List<T> list(Pageable pageable, @Nullable Filter filter);

}
```

```java
public interface CrudService<T, ID> extends ListService<T> {

    /**
     * Saves the given object and returns the (potentially) updated object.
     * <p>
     * If you store the object in a SQL database, the returned object might have
     * a new id or updated consistency version.
     *
     * @param value
     *            the object to save
     * @return the fresh object or {@code null} if no object was found to update
     */
    @Nullable
    T save(T value);

    /**
     * Deletes the object with the given id.
     *
     * @param id
     *            the id of the object to delete
     */
    void delete(ID id);
}
```

tarekoraby:
Those look good.

But thinking out loud here, I wonder if `CrudRepositoryService` could expose more of the Spring repo methods like:

- count
- findAll
- exists
- saveAll

(I'm not sure, though, about exposing the deleteAll methods by default as part of `CrudRepositoryService` )

Artur-:
Oh, there is actually one important missing... `get`

Artur-:
If we want to expose batch methods then we could expose `saveAll` and `deleteAll`. Also we could expose `count`. But these should not be mandatory in the interface.

What would you do with `exists`  that cannot be done using `get` and what would `findAll` do compared to `list`
?

tarekoraby:
> But these should not be mandatory in the interface.

Agreed. I was talking about `CrudRepositoryService` only.

> What would you do with exists that cannot be done using get and what would findAll do compared to list
?

A boolean instead of a potentially large object moving from the DB, to Hilla, to the frontend.

PR #1455, opened 2023-10-04
feat: Add get/count/exists to CrudRepositoryService

Artur-:
The get method is mandatory to implement, exists is default implemented and count is only implemented in the CrudRepositoryService

For #1404

### DIFF
--- a/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudRepositoryService.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudRepositoryService.java
@@ -38,29 +38,8 @@ public class CrudRepositoryService<T, ID, R extends CrudRepository<T, ID> & JpaS
     }
 
     @Override
-    public T get(ID id) {
-        return getRepository().findById(id).orElse(null);
-    }
-
-    @Override
-    public boolean exists(ID id) {
-        return getRepository().existsById(id);
-    }
-
-    @Override
     public void delete(ID id) {
         getRepository().deleteById(id);
-    }
-
-    /**
-     * Counts the number of entities that match the given filter.
-     *
-     * @param filter
-     *            the filter, or {@code null} to use no filter
-     * @return
-     */
-    public long count(@Nullable Filter filter) {
-        return getRepository().count(toSpec(filter));
     }
 
 }
--- a/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudRepositoryService.java
+++ b/packages/java/endpoint/src/main/java/dev/hilla/crud/CrudRepositoryService.java
@@ -2,6 +2,7 @@ package dev.hilla.crud;
 
 import dev.hilla.EndpointExposed;
 import dev.hilla.Nullable;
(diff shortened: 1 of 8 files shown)


sissbruecker (approved):

packages/java/endpoint/src/main/java/dev/hilla/crud/CrudRepositoryService.java
```diff
@@ -36,9 +37,30 @@ public CrudRepositoryService(R repository) {
...
+     * Counts the number of entities that match the given filter.
+     *
+     * @param filter
+     *            the filter, or {@code null} to use no filter
+     * @return
+     */
+    public long count(@Nullable Filter filter) {
+        return getRepository().count(toSpec(filter));
+    }
+
```
taefi, line 65: I wonder if these three methods should be defined in the `ListRepositoryService` instead.

taefi (changes requested):

packages/java/endpoint/src/main/java/dev/hilla/crud/CrudRepositoryService.java
Artur-: I can move them

Artur-: ah but it becomes a bit odd then, because we don't want to make `ListService` any more complex, as you don't need the methods for listing items in a grid

Artur-: So the methods do not implement any interface, they just exist there, even though `get` and `exists` are in `CrudService` and need to match

Artur-: What does @sissbruecker think?

Artur-: I would just leave them where they are

tarekoraby: > ah but it becomes a bit odd then, because we don't want to make ListService any more complex, as you don't need the methods for listing items in a grid

What about creating another interface or super class (not sure what it would be called) with count() and exists() and get() that both `ListRepositoryService` and `CrudRepositoryService` would implement?

taefi: Well, it is a little bit vague what are we trying to solve here. Just saving developers from writing some extra lines of code? In that case, I'd put whatever "reads" something from the data storage in the `ListRepositoryService`, and whatever manipulates data in the `CrudREpositoryService`. The interfaces `ListService` and `CrudService` still can remain as simple as possible, just having whatever necessary for their job. 

Another approach is to remain minimal, and ask who needs what? In that case who needs `count`? We can remove it. 

Are there other considerations?

Artur-: Ok, does the new version make sense?

PR #1460, opened 2023-10-04
feat: Show a header text if requested

Artur-:


packages/ts/react-grid/src/autocrud.tsx
```diff
@@ -33,7 +34,18 @@ export function ExperimentalAutoCrud<TItem>({ service, model, noDelete }: AutoCr
...
+              width: '100%',
+              justifyContent: 'space-between',
+              paddingLeft: 'var(--lumo-space-m)',
+              paddingRight: 'var(--lumo-space-m)',
+              paddingBottom: 'var(--lumo-space-s)',
+              paddingTop: 'var(--lumo-space-s)',
+              alignItems: 'center',
+            }}
+          >
+            <h2 style={{ fontSize: 'var(--lumo-font-size-l)' }}>{header}</h2>
```
sissbruecker, line 48: Ideally we wouldn't render the header element if there is no header. In order to keep the alignment of the button to the right, we could set `margin-left: auto` on the button instead of using `justifyContent: space-between` on the container.

Artur-: Do you think it is worth it?

sissbruecker: It's a potential a11y issue, and if we can avoid it with a small change, why not.

Bildschirmfoto 2023-10-05 um 13 28 19

Artur-: FIxed

sissbruecker (approved):

PR #1465, opened 2023-10-05
feat: implement locale specific date parser

sissbruecker:
## Description

Implements locale-specific parsing of dates, as a pre-requisite for being able to parse and format dates in date pickers in the same format that is used for rendering cells (https://github.com/vaadin/hilla/pull/1442). The implementation is loosely based on that of the Flow component. The basic idea is to format an example date with a specific locale, and use that to extract a Regex pattern that allows to parse a date in that format in reverse. 

This implementation is far from perfect: there is no leniency for additional or missing whitespace, it does not handle dates before 0 (mostly because Intl.DateTimeFormat doesn't), only works with arabic numerals etc. 

Using a library like `date-fns` would help with some of those issues. But it doesn't solve figuring out which date format to use based on the current browser locale, so reverse engineering the pattern from a sample date would still be necessary.

### DIFF
--- a/packages/ts/react-grid/src/date-formatter.ts
+++ b/packages/ts/react-grid/src/date-formatter.ts
@@ -1,5 +1,5 @@
 function getFormatRegex(format: Intl.DateTimeFormat) {
-  const sampleDate = new Date(1234, 5-1, 6);
+  const sampleDate = new Date(1234, 5 - 1, 6);
   const formattedSample = format.format(sampleDate);
   const pattern = formattedSample
     .replace('1234', '(?<year>\\d+)')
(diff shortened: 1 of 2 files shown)


packages/ts/react-grid/src/date-formatter.ts
```diff
@@ -0,0 +1,54 @@
...
+    .replace('6', '(?<day>\\d+)');
+
+  return new RegExp(pattern, 'u');
+}
+
+export interface DateObject {
+  year: number;
+  month: number;
+  day: number;
+}
```
sissbruecker, line 18: This is what the date picker API uses, so I built the formatter class around that rather than using actual date instances.

packages/ts/react-grid/src/date-formatter.ts
```diff
@@ -0,0 +1,54 @@
+function getFormatRegex(format: Intl.DateTimeFormat) {
+  const sampleDate = new Date(1234, 4, 6);
```
Artur-, line 2:
```suggestion
  const sampleDate = new Date(1234, 5-1, 6);
```

Artur- (approved):
This is horrible

Artur- (approved):

PR #1468, opened 2023-10-05
test: Test that data provider is initially only called once

### LINKED ISSUE #1325: [autogrid] DataProvider.list is called multiple times when an AutoGrid is rendered
### Description of the bug 
`list(...)` should be called once, it is called three times

Test in the `autogrid-call-data-provider-once` branch

Artur-:
Closes #1325

### DIFF
--- a/packages/ts/react-grid/test/autogrid.spec.tsx
+++ b/packages/ts/react-grid/test/autogrid.spec.tsx
@@ -120,6 +120,13 @@ describe('@hilla/react-grid', () => {
       expect(grid.getAttribute('model')).to.be.null;
       expect(grid.getAttribute('service')).to.be.null;
     });
+    it('calls data provider list() only once for initial data', async () => {
+      const testService = personService();
+      expect(testService.callCount).to.equal(0);
+      render(<AutoGrid service={testService} model={PersonModel} />);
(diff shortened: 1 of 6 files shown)


packages/ts/react-grid/test/autogrid.spec.tsx
```diff
@@ -120,6 +120,13 @@ describe('@hilla/react-grid', () => {
       expect(grid.getAttribute('model')).to.be.null;
       expect(grid.getAttribute('service')).to.be.null;
     });
+    it('calls data provider list() only once for initial data', async () => {
```
sissbruecker, line 123: This test also passes when removing the changes. The only scenario where it triggers less requests is when toggling header filters, though in that case you still get more than one as the sort is applied with a delay again. 

The change probably makes sense overall, but was there some specific case that you were testing with?

Artur-: It seems like the data provider is no longer being called ever with an undefined filter. Let's just add the test so it can catch potential future problems. The actual problem must have been solved after the issue was created. I just rebased an old fix for it

sissbruecker (approved):

PR #1470, opened 2023-10-05
feat: throw if an @EndpointExposed class contains AccessControl annotations

### LINKED ISSUE #1314: Exception should be thrown if an @EndpointExposed class contains AccessControl annotations
As a follow up for #1305, To make the policy more clear, an exception should be thrown while starting the application that explicitly mentions that:
```
Classes annotated with `@EndpointExposed` should not contain any of access control annotations and
the exception is for preventing the confusion. The children endpoints class level access control rules 
will be applied for the inherited methods of this class. If the access control rules for an inherited 
method  should not follow the rules of the child endpoint, that method should be overridden and annotated  
with access control annotations explicitly. 
```

taefi:
## Description

Throw a ParserException if an `@EndpointExposed` class contains AccessControl annotations.

Fixes #1314

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/Parser.java
+++ b/packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/Parser.java
@@ -3,13 +3,19 @@ package dev.hilla.parser.core;
 import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 
-import io.github.classgraph.AnnotationInfo;
 import io.github.classgraph.ClassInfo;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


packages/java/parser-jvm-core/src/main/java/dev/hilla/parser/core/Parser.java
```diff
@@ -27,6 +31,22 @@ public final class Parser {
     private static final Logger logger = LoggerFactory.getLogger(Parser.class);
     private final Config config;
 
+    private static final String ENDPOINT_EXPOSED_AND_ACL_ANNOTATIONS_ERROR_TEMPLATE = """
+            Class `%s` is annotated with `%s` and `%s` annotation. %n
+            Classes annotated with `%s` should not contain any of access control annotations and %n
+            this exception is for preventing the confusion. The class level access control rules of the child class %n
```
sissbruecker, line 37: Nit: "should not" -> "must not" would be more appropriate given that doing so doesn't work at all. 

It's also not clear what is meant by "confusion", I think it would be OK to leave the reasoning out and just state the fact.

taefi: Would something like the following be better?
```
... and this exception is for preventing the application startup with misconfiguration.
```

sissbruecker: Looks good 👍

taefi: Done.

sissbruecker (approved):

PR #1473, opened 2023-10-06
feat: Use count for scrolling size, if available

### LINKED ISSUE #1467: [autogrid] support showing the "Total Number Of Records" for the grid optionally
### Describe your motivation

When having paging support with scrolling, the number of records are not determined, and the only way to know how many records exists is to scrolling down continuously to fetch all the records. So having the "Total Number Of Records" helps with UX, while adding it manually on `<AutoGrid />` is not as easy as it should be. 

By default, it should not be enabled, as it is not performant friendly in many situations. 

In addition to that, showing the "Number of Records" when one or more filters is applied to the grid is also a super useful feature in AutoGrid.

I'd like to enable it like this:

```jsx
<AutoGrid service={...} model={...} totalCount recordCount />
```
Where `totalCount` is for the "Total Number of Records" and `recordCount` is the one referring to the number of records after filtering.

_No response_

### Additional context

_No response_

Artur-:
Also fixes #1467
Add support to show the "Total Number Of Records" for the grid optionally


krissvaa:
There is a problem with the double border on top because of the extra column group wrapper (extra hidden header).

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -88,6 +97,20 @@ interface AutoGridOwnProps<TItem> {
    * grid.
    */
   rowNumbers?: boolean;
+  /**
+   * When enabled, shows the total count of records in the grid footer.
+   */
+  totalCount?: boolean;
+  /**
+   * When enabled, shows the visible record count in the grid footer.
```
taefi, line 105: I suggest we rename the property to `filteredCount` and only show it when there is a filter applied to the grid. To be clear, as the number of filtered records may exceed the visible viewport of the grid, IMO the end user wants to see how many records (out of total) are affected by the applied filters, no matter how the UI is handling the pagination (infinite scrolling, or paging navigation buttons or etc.)
 
I'm not sure if it is doing the same thing now, as I couldn't find out whether infinite scrolling and pageSize are addressing something related to filtered count, or that part of the code is addressing something else. 
```suggestion
   * When enabled, shows the filtered record count in the grid footer.
```

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -88,6 +97,20 @@ interface AutoGridOwnProps<TItem> {
...
+   * if totalCount is also enabled, it will show both totalCount and recordCount
+   */
+  recordCount?: boolean;
```
taefi, line 108:
```suggestion
  filteredCount?: boolean;
```

packages/ts/react-crud/src/autogrid-renderers.tsx
```diff
@@ -70,3 +72,37 @@ export function AutoGridDateTimeRenderer<TItem>({ item }: RendererOptions<TItem>
 export function AutoGridRowNumberRenderer<TItem>({ model }: RendererOptions<TItem>): JSX.Element {
   return <>{model.index + 1}</>;
 }
+
+export function AutoGridFooterItemCountRenderer(
+  itemCountHolder: AutoGridItemCountHolder,
+  columns: JSX.Element[],
+  footerCountRenderer?: ComponentType<{ itemCountHolder: AutoGridItemCountHolder }>,
+): JSX.Element[] {
+  let newColumns: JSX.Element[] = columns;
```
taefi, line 81: Maybe nitpicking (feel free to ignore):
Could the `newColumns` variable be omitted completely, and the if-else be simplified as:
```tsx
if (footerCountRenderer) {
    const FooterRenderer = footerCountRenderer;
    return [
      <GridColumnGroup key="grid-footer" footerRenderer={() => <FooterRenderer itemCountHolder={itemCountHolder} />}>
        {columns}
      </GridColumnGroup>,
    ];
}
// the else logic goes here without an explicit else block
```

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -258,13 +332,24 @@ export function AutoGrid<TItem>({
     }
   };
 
+  const recordCountHolder: AutoGridItemCountHolder = {
+    totalCount,
+    recordCount,
```
taefi, line 337: The same discussion as earlier about `totalCount` and `filteredCount`

taefi (changes requested):

packages/ts/react-crud/src/autogrid.tsx
taefi, line 1: Could the conflict for this file be resolved so that it would be easier to see the changes based on the latest status of this file?

taefi (changes requested):

packages/ts/react-crud/src/autogrid-renderers.tsx
```diff
@@ -70,3 +72,36 @@ export function AutoGridDateTimeRenderer<TItem>({ item }: RendererOptions<TItem>
...
+        {columns}
+      </GridColumnGroup>,
+    ];
+  }
+  let filterCountText: string | undefined;
+  if (itemCountHolder.filteredCount && itemCountHolder.filteredItemCount >= 0) {
+    filterCountText =
+      itemCountHolder.totalCount && itemCountHolder.totalItemCount >= 0
+        ? `Showing: ${itemCountHolder.filteredItemCount} (${itemCountHolder.totalItemCount})`
+        : `Showing: ${itemCountHolder.filteredItemCount}`;
```
taefi, line 94: There is a chance that this comment is reflecting my own taste, anyway: 
I'm not sure if the word "Showing" or "Visible" is exactly the word I expect to see as it confuses me what does it exactly mean:
- Are the filtered rows are all visible now?
- Are they loaded already to the browser or I need to scroll enough to trigger the fetch?
and honestly, I don't have a better suggestion either :)

But, I would like to see the outcome of having both `totalCount` and `filteredCount` as something more explicit:
- "[Showing]|[or some other word] {filtered count} rows (**out of** {total count})"

taefi (changes requested):

taefi:
Would it be possible to fix the sonar bug and code smells before we merge?

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -156,21 +257,20 @@ function createDataProvider<TItem>(
...
-
-      const cacheSize = (grid as GridElementWithInternalAPI<TItem>)._dataProviderController.rootCache.size;
-      if (cacheSize !== undefined && size < cacheSize) {
-        // Only allow size to grow here to avoid shrinking the size when scrolled down and sorting
-        size = undefined;
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    if (!(service as any).count) {
+      if (itemCountHolder.totalCount ?? itemCountHolder.filteredCount) {
+        console.error(
```
sissbruecker, line 265: This code runs with every single data provider request, so when scrolling it keeps logging errors. We should move it out of the data provider function.

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -156,21 +257,20 @@ function createDataProvider<TItem>(
...
+          '"totalCount" and/or "filteredCount" props require the provided service to implement the CountService interface.',
+        );
       }
+      initDataProviderUsingInfiniteScrolling(items, pageSize, pageNumber, grid, callback);
```
sissbruecker, line 269: Reading the code it is confusing that there is a `initDataProviderUsingCountService` but it doesn't actually init or create anything but instead just contains some logic of the data provider request. 

Maybe it would be clearer to structure it so that there are two different data provider implementations that reuse some common parts like querying the list service and applying the `recalculateColumnWidths` timeout fix. Like, have `createInfiniteDataProvider` and `createFiniteDataProvider` functions that return a data provider with the respective implementation.

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -130,53 +180,133 @@ type GridElementWithInternalAPI<TItem = GridDefaultItem> = GridElement<TItem> &
...
+    afterCallBackExecuted();
+  };
+}
+
+function createFiniteDataProvider<TItem>(
+  service: CountService<TItem> & ListService<TItem>,
+  filter: MutableRefObject<FilterUnion | undefined>,
+  footerRef: MutableRefObject<Dispatch<SetStateAction<number>>>,
+  itemCountHolder: AutoGridItemCountHolder,
+  afterCallBackExecuted: () => void,
```
sissbruecker, line 257: Would suggest `afterPageLoaded` to give some semantic meaning.

packages/ts/react-crud/src/autogrid.tsx
sissbruecker, line 1: It seems the item count is lagging behind when filtering the grid:

https://github.com/vaadin/hilla/assets/357820/a5431d6e-4bcd-4b42-b877-8b4474fe09e1

packages/ts/react-crud/src/autogrid.tsx
taefi: Done.

sissbruecker (approved):

PR #1478, opened 2023-10-06
chore: replace Lerna with Nx (2.2) (#1366)

platosha:
(cherry picked from commit 7f47894072b256dd61cd1d4085599ad3bb1964a6)


packages/ts/react-form/src/index.ts
```diff
@@ -17,18 +18,21 @@ import {
...
+declare const __VERSION__: string;
+
+const $wnd = window as VaadinWindow;
 
 $wnd.Vaadin ??= {};
 $wnd.Vaadin.registrations ??= [];
 $wnd.Vaadin.registrations.push({
   is: '@hilla/react-form',
-  version: /* updated-by-script */ '2.2.0',
+  version: /* updated-by-script */ '2.3.0-alpha6',
```
Lodin, line 31:
```suggestion
  version: __VERSION__,
```

If there is any other obsolete version specification, we should address it as well.

packages/ts/form/package.json
```diff
@@ -67,24 +69,23 @@
...
+    "@types/mocha": "^10.0.2",
+    "@types/sinon": "^10.0.17",
+    "@types/sinon-chai": "^3.2.10",
+    "@types/validator": "^13.11.2",
     "chai-dom": "^1.11.0",
-    "sinon": "^15.2.0",
+    "sinon": "^16.0.0",
     "sinon-chai": "^3.7.0",
-    "typescript": "^5.1.6"
+    "typescript": "^5.2.2"
```
Lodin, line 89: Artur reverted that update since Flow is using `5.1.6` for now. So I suggest using the same TS version here (without `^` btw).

Lodin (changes requested):

packages/ts/form/package.json
platosha: According to https://github.com/vaadin/flow/blob/24.1/flow-server/src/main/resources/com/vaadin/flow/server/frontend/dependencies/default/package.json it's 5.0.4 for Flow 24.1 at the moment, so I used that.

Lodin (approved):

PR #1487, opened 2023-10-09
fix: render nested properties in auto-grid

sissbruecker:
Fixes rendering of nested properties when using auto-grid specific renderers.

### DIFF
--- a/packages/ts/react-grid/test/autogrid.spec.tsx
+++ b/packages/ts/react-grid/test/autogrid.spec.tsx
@@ -52,6 +52,7 @@ describe('@hilla/react-grid', () => {
   function TestAutoGridNoHeaderFilters(customProps: Partial<AutoGridProps<Person>>) {
     return <AutoGrid service={personService()} model={PersonModel} noHeaderFilters {...customProps}></AutoGrid>;
   }
+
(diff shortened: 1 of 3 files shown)


packages/ts/react-grid/test/autogrid.spec.tsx
```diff
@@ -467,5 +468,30 @@ describe('@hilla/react-grid', () => {
...
+    });
+
+    it('renders nested booleans as icons', () => {
+      const vip = getBodyCellContent(grid, 0, 9).querySelector('vaadin-icon')!;
+      expect(vip.icon).to.equal('lumo:checkmark');
+      const notVip = getBodyCellContent(grid, 1, 9).querySelector('vaadin-icon')!;
+      expect(notVip.icon).to.equal('lumo:minus');
+    });
+
+    it('renders java.util.Date as right aligned', () => {
```
taefi, line 491: It is a minor comment, but we support LocalDate and not `java.util.Date` at the moment.

taefi (approved):

packages/ts/react-grid/test/autogrid.spec.tsx
sissbruecker: Discussed this, and `java.util.Date` is currently lacking support for filtering. Will add this in a separate PR.

PR #1489, opened 2023-10-09
reactor(react-grid): rename filter prop to experimentalFilter

platosha:


packages/ts/react-grid/test/autogrid.spec.tsx
sissbruecker, line 1: There's also two integration tests using the property:
- `ReadOnlyGridOrFilter.tsx`
- `ReadOnlyGridSinglePropertyFilter.tsx`

sissbruecker (approved):

Issue #1509, opened 2023-10-10
[autogrid][dx] Enable multi-column sorting


taefi:
### Describe your motivation

Having the multi-column sorting enabled by default is a nice feature to have.
Then, there could be a way to disable sorting completely, or fall back to single column sorting based on developer discretion.

tarekoraby:
I think @sissbruecker's [argument against enabling multi-sorting by default ](https://github.com/vaadin/hilla/pull/1510#issue-1936237905) makes sense

> I did not enable multi-sort by default, as I found it quite bothersome that you first have to remove sort from one column (two clicks) if you just want to sort a different column (third click). 

Also, as discussed in one of the dailies, let's aim to be consistent with the default Grid behavior whenever possible (where multi-sorting is not the default). 

We should, however, provide an option to enable it as an opt-in.

PR #1510, opened 2023-10-10
feat: add multi-sort support to auto-grid

sissbruecker:
Changes the auto-grid's internal sort state to support multi-sort. 

For now I did not enable multi-sort by default, as I found it quite bothersome that you first have to remove sort from one column (two clicks) if you just want to sort a different column (third click). Let's discuss this separately, maybe we can use the new multi-sort on shift click mode as a compromise.

Closes https://github.com/vaadin/hilla/issues/1509

### DIFF
--- a/packages/ts/react-crud/test/autogrid.spec.tsx
+++ b/packages/ts/react-crud/test/autogrid.spec.tsx
@@ -6,11 +6,14 @@ import type { TextFieldElement } from '@hilla/react-components/TextField.js';
 import { render, type RenderResult } from '@testing-library/react';
 
 import sinonChai from 'sinon-chai';
+import type { CrudService } from '../crud';
 import { AutoGrid, type AutoGridProps } from '../src/autogrid.js';
 import { _generateHeader } from '../src/property-info.js';
 import type AndFilter from '../src/types/dev/hilla/crud/filter/AndFilter.js';
 import Matcher from '../src/types/dev/hilla/crud/filter/PropertyStringFilter/Matcher.js';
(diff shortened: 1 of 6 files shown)


packages/ts/react-crud/test/autogrid.spec.tsx
```diff
@@ -141,6 +151,66 @@ describe('@hilla/react-crud', () => {
...
+            { property: 'lastName', direction: Direction.DESC, ignoreCase: false },
+          ],
+        });
+        expect(getSortOrder(grid)).to.eql([
+          { property: 'firstName', direction: Direction.ASC },
+          { property: 'lastName', direction: Direction.DESC },
+        ]);
+      });
+    });
+
```
taefi, line 213: Could we have a test to make sure when triggering the sort for `lastName` for the third time, it is excluded from the list of sorter columns?

taefi (changes requested):

packages/ts/react-crud/test/autogrid.spec.tsx
sissbruecker: The sort function actually just sets the direction programmatically, it doesn't use clicks:
https://github.com/vaadin/hilla/blob/07cf6f3bf17bae2b03579ee15e59dbda0eb28842/packages/ts/react-crud/test/grid-test-helpers.ts#L66-L72

sissbruecker: But still, removing the direction is a code path that should be tested, good point. Will add a test.

sissbruecker: Done

taefi (approved):

Issue #1496, opened 2023-10-10
Auto Grid: enable configuring columns by referencing models


platosha:
Currently the only way to refer to columns in Auto Grid API is using string names.

Expected: the Auto Grid API should support referencing model instances as a primary way of customizing columns (visibility, order, rendering options, etc).

PR #1512, opened 2023-10-11
feat(react-grid): disallow unknown column names

### LINKED ISSUE #1495: Auto Grid: type checking for string property references
Some properties of Auto Grid require referencing model properties by string, namely `visibleColumns`, and `columnOptions`.

Currently the TypeScript definitions for those are just `string`, no static type checking is done on the string name.

Expected: TypeScript error for unknown property names that are not keys in the given model.

platosha:
Fixes #1495

### DIFF
--- a/packages/ts/react-crud/src/autogrid.tsx
+++ b/packages/ts/react-crud/src/autogrid.tsx
@@ -12,7 +12,7 @@ import { GridColumn } from '@hilla/react-components/GridColumn.js';
 import { GridColumnGroup } from '@hilla/react-components/GridColumnGroup.js';
 import { useEffect, useRef, useState, type JSX } from 'react';
 import { ColumnContext, type SortState } from './autogrid-column-context.js';
-import { getColumnOptions, type ColumnOptions } from './autogrid-columns.js';
+import { getColumnProps, type ColumnProps } from './autogrid-columns.js';
 import { AutoGridRowNumberRenderer } from './autogrid-renderers.js';
 import type { ListService } from './crud';
 import { HeaderSorter } from './header-sorter';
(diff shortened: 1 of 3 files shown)


packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -23,16 +23,84 @@ import type PropertyStringFilter from './types/dev/hilla/crud/filter/PropertyStr
 import type Sort from './types/dev/hilla/mappedtypes/Sort';
 import Direction from './types/org/springframework/data/domain/Sort/Direction';
 
+type AllPropertyNames<TItem> = string & keyof TItem;
```
sissbruecker, line 26: The new types slow down type-checking a lot (testing in IntelliJ). It's several times slower than without, regardless what you touch. 

Also affects auto-complete suggestions. The new types allow providing suggestions for which columns you can use, but it takes several seconds to calculate them. If you just force IntelliJ to auto-complete without waiting for several seconds after typing, it just shows some random stuff.


https://github.com/vaadin/hilla/assets/357820/6553d845-e7c9-4d5d-b309-90449e28d082

platosha: I have limited inheritance to 3 levels deep, did it help?

sissbruecker: Performance is a lot better now, the auto-complete as well. Now it starts suggesting sensible options right after typing the first letter of a name.

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -23,16 +23,63 @@ import type PropertyStringFilter from './types/dev/hilla/crud/filter/PropertyStr
...
+
+type UnsupportedColumnDataTypes = unknown[];
```
sissbruecker, line 28: The property filtering doesn't quite match what is implemented in code. It suggests nested properties from `@ManyToOne` relations which are not included by this logic here: 
https://github.com/vaadin/hilla/blob/ecc473cfabc068803a9cada7b75760c4365c6b3d/packages/ts/react-crud/src/property-info.ts#L66-L91

The same would apply for any other object model not annotated with `@OneToOne`.

If we want to make this work, we would have to:
- Include nested properties from all object models
- But do not show them by default

sissbruecker: I was using `GridWithEntityReferences.tsx` for testing.

platosha:
As noted above, this idea is not compatible with the way nested objects work right now.  Let me close this and focus on solving #1496 instead.

PR #1523, opened 2023-10-14
feat: Simple dev tools plugin with support for opening H2 console

Artur-:


Artur-:
This should be merged after #1485 as it depends on the project setup for `hilla-dev-mode`

packages/java/engine-runtime/src/main/java/dev/hilla/internal/TaskGenerateEndpointImpl.java
```diff
@@ -70,5 +78,27 @@ public void execute() throws ExecutionFailedException {
...
+                        }
+                        """;
+                FileUtils.writeStringToFile(vaadinTs, current,
+                        StandardCharsets.UTF_8);
+            }
+        } catch (IOException e) {
+            throw new ExecutionFailedException("Failed to update vaadin.ts", e);
+
+        }
+
```
taefi, line 102: Couldn't an `if (productionMode) { return; }` prevent opening and reading `vaadinTs` when it is building for production?

packages/java/hilla-dev-mode/src/main/resources/META-INF/resources/frontend/dev-tools-database.ts
```diff
@@ -0,0 +1,68 @@
...
+    const h2Window = window.open(this.h2path)!;
+    await waitForConsole(h2Window);
+    loginUsingUrl(h2Window, this.h2jdbcUrl);
+  }
+}
+
+const plugin: DevToolsPlugin = {
+  init(devToolsInterface: DevToolsInterface): void {
+    devTools = devToolsInterface;
+    devTools.addTab('Database', 'devtools-database');
```
taefi, line 63:
```suggestion
    devTools.addTab('H2 Database', 'devtools-database');
```

taefi (changes requested):
It seems that there are two unrelated features being added in a single PR, right? 
1. Hilla HotSwap Plugin
2. A dev tools plugin for opening H2 console

packages/java/hilla-dev-mode/src/main/resources/META-INF/resources/frontend/dev-tools-database.ts
Artur-: I would keep it as "database" as the intent is to add more functionality there, like example data generation. This is just a first step

packages/java/engine-runtime/src/main/java/dev/hilla/internal/TaskGenerateEndpointImpl.java
Artur-: Is that better?

Artur-: The current version does not read vaadinTs in production mode either

taefi (approved):

packages/java/engine-runtime/src/main/java/dev/hilla/internal/TaskGenerateEndpointImpl.java
Artur-: Extracted to a method to make it cleaner

taefi (approved):

Issue #1524, opened 2023-10-16
[auto-form] Editing numbers


sissbruecker:
Number properties are currently using a `TextField`. It should be either an `IntegerField` or a `NumberField`, depending on the Java type in the model metadata.

PR #1533, opened 2023-10-16
feat: edit numbers with auto form

sissbruecker:
## Description

Adds auto form fields for number properties.

Fixes https://github.com/vaadin/hilla/issues/1524


packages/ts/react-crud/src/autogrid-columns.tsx
```diff
@@ -22,7 +22,8 @@ export type ColumnOptions = Omit<GridColumnProps<any>, 'dangerouslySetInnerHTML'
 function getTypeColumnOptions(propertyInfo: PropertyInfo): ColumnOptions {
   // eslint-disable-next-line default-case
   switch (propertyInfo.type) {
-    case 'number':
+    case 'integer':
+    case 'decimal':
```
sissbruecker, line 26: For auto-grid both types render as integers for now, same as before. I didn't want to bloat up the PR any further. Let's address that in a separate PR.

packages/ts/react-crud/test/TextFieldController.ts
```diff
@@ -19,8 +19,8 @@ export default class TextFieldController {
     child: HTMLElement,
     user: ReturnType<(typeof userEvent)['setup']>,
   ): Promise<TextFieldController> {
-    const instance = await waitFor(() => child.closest('vaadin-text-field')!);
-    return new TextFieldController(instance, user);
+    const instance = await waitFor(() => child.parentElement!);
+    return new TextFieldController(instance as TextFieldElement, user);
```
sissbruecker, line 23: The new test abstractions are rather difficult to deal with. For now I just made `TextFieldController` work with any field, we can consider refactoring this in a separate PR to make it more generic / flexible.

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -39,7 +40,7 @@ describe('@hilla/react-crud', () => {
     function getExpectedValues(person: Person) {
       return (Object.entries(person) as ReadonlyArray<[keyof Person, Person[keyof Person]]>)
         .filter(([key]) => KEYS.includes(key))
-        .map(([, value]) => value);
+        .map(([, value]) => value.toString());
```
sissbruecker, line 43: This should return something that can be compared to field values, which will always be strings. So the numbers in the expected person object need to be converted into strings as well. I think this was working previously because binder was setting the value of `vaadin-text-field` to a `0` (the number, not the string), and the component doesn't have logic to sanitize that into a string. `vaadin-integer-field` and `vaadin-number-field` OTOH do have that sanitization.

packages/java/tests/spring/react-grid-test/package.json
```diff
@@ -15,11 +15,11 @@
     "@hilla/generator-typescript-plugin-subtypes": "file:../../../../ts/generator-typescript-plugin-subtypes",
     "@hilla/generator-typescript-utils": "file:../../../../ts/generator-typescript-utils",
     "@hilla/react-auth": "file:../../../../ts/react-auth",
-    "@hilla/react-components": "2.2.0-beta2",
+    "@hilla/react-components": "2.2.0-rc1",
     "@hilla/react-crud": "file:../../../../ts/react-crud",
     "@hilla/react-form": "file:../../../../ts/react-form",
     "@polymer/polymer": "3.5.1",
-    "@vaadin/bundles": "24.2.0-beta2",
+    "@vaadin/bundles": "24.2.0-rc1",
```
cromoteca, line 22: Do you really need the new versions here? Otherwise I'd leave `package.json` out of the PR. Or, if this is needed, also add the lock file.

cromoteca (changes requested):
Everything's fine for me, I just have that doubt about `package.json`.

packages/java/tests/spring/react-grid-test/package.json
sissbruecker: Not really needed, it just updates automatically when starting the IT app. Reverted the change.

PR #1535, opened 2023-10-17
chore: upgrade jackson.version to 2.15.3

ZheSun88:
this will also resolve one found vulnerability https://nvd.nist.gov/vuln/detail/CVE-2023-35116


cromoteca:
Swagger Core still uses Jackson 2.15.2. If this update is important, we have to add exclusions, otherwise let's wait for a new Swagger version.

ZheSun88:
well.. it is a disputed report.. in this case, let us wait for the next swagger.

PR #1536, opened 2023-10-17
chore: fix Gradle tests failure

Lodin:


packages/java/tests/gradle/single-module-tests/package.json
```diff
@@ -15,21 +15,14 @@
...
     "@hilla/react-auth": "file:../../../../ts/react-auth",
-    "@hilla/react-components": "2.1.0-alpha9",
-    "@hilla/react-form": "file:../../../../ts/react-form",
+    "@hilla/react-components": "2.1.11",
     "@hilla/react-crud": "file:../../../../ts/react-crud",
+    "@hilla/react-form": "file:../../../../ts/react-form",
     "@polymer/polymer": "3.5.1",
     "@vaadin-component-factory/vcf-nav": "1.1.0",
-    "@vaadin/bundles": "24.1.0-alpha9",
+    "@vaadin/bundles": "24.1.11",
```
taefi, line 23: Isn't there a newer version of `bundles` e.g. `24.2.0`?

packages/java/tests/gradle/single-module-tests/package.json
```diff
@@ -72,68 +65,55 @@
...
+      "@hilla/generator-typescript-cli": "2.2.2",
+      "@hilla/generator-typescript-core": "2.2.2",
+      "@hilla/generator-typescript-plugin-backbone": "2.2.2",
+      "@hilla/generator-typescript-plugin-barrel": "2.2.2",
+      "@hilla/generator-typescript-plugin-client": "2.2.2",
+      "@hilla/generator-typescript-plugin-model": "2.2.2",
+      "@hilla/generator-typescript-plugin-push": "2.2.2",
+      "@hilla/generator-typescript-utils": "2.2.2",
+      "@hilla/react-components": "2.1.11",
+      "@hilla/react-form": "2.2.2",
```
taefi, line 91: Why not using `2.3.1` already?

taefi (changes requested):
ATM, I think `single-module-tests` can be just a bare-bone Hilla project with no code in the `src/main` folder (and possibly no heavy package.json content, then no `npm install` and not extra waits for it). Its existence purpose is just for running the `ProductionBuildFunctionalIT` tests from outside of the `single-module` project. There shouldn't be any sort of builds happening for the `single-module-tests` project.

platosha:
Numbers are outdated, but it's OK to update them in another PR.

taefi (approved):

PR #1538, opened 2023-10-17
chore: fix building hilla gradle plugin

sissbruecker:
The gradle plugin should build in CI same as every other module. Otherwise bumping a snapshot version results in the plugin missing from the build, causing it to fail.


Lodin (approved):
LGTM

PR #1557, opened 2023-10-19
feat: add discard button to auto form

### LINKED ISSUE #1433: [autoform] Discard button when editing
When editing an item (not when creating a new item) there should be a "discard" button to indicate that the changes you made are not relevant and should be discarded. The form is reverted to the original values

sissbruecker:
Shows a discard button in auto form if the form is dirty. Clicking the button resets the form.

Fixes #1433

### DIFF
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -76,8 +76,14 @@ export function ExperimentalAutoForm<TItem>({
           <AutoFormField key={propertyInfo.name} propertyInfo={propertyInfo} form={form} disabled={disabled} />
         ))}
       {formError ? <div style={{ color: 'var(--lumo-error-color)' }}>{formError}</div> : <></>}
-      <HorizontalLayout style={{ marginTop: 'var(--lumo-space-m)' }}>
+      <HorizontalLayout theme="spacing" style={{ marginTop: 'var(--lumo-space-m)', alignSelf: 'flex-end' }}>
+        {form.dirty ? (
+          <Button theme="tertiary" onClick={() => form.reset()}>
+            Discard
+          </Button>
+        ) : null}
         <Button
+          theme="primary"
           disabled={disabled}
           // eslint-disable-next-line @typescript-eslint/no-misused-promises
           onClick={submitButtonClicked}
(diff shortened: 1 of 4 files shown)


cromoteca (approved):

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -76,8 +76,14 @@ export function ExperimentalAutoForm<TItem>({
           <AutoFormField key={propertyInfo.name} propertyInfo={propertyInfo} form={form} disabled={disabled} />
         ))}
       {formError ? <div style={{ color: 'var(--lumo-error-color)' }}>{formError}</div> : <></>}
-      <HorizontalLayout style={{ marginTop: 'var(--lumo-space-m)' }}>
+      <HorizontalLayout theme="spacing" style={{ marginTop: 'var(--lumo-space-m)', alignSelf: 'flex-end' }}>
+        {form.dirty ? (
```
taefi, line 80: I'm not a UX designer or anything, but I was thinking that the primary button (in this case the submit button) should appear first to make it more accessible for the end user.

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -76,8 +76,14 @@ export function ExperimentalAutoForm<TItem>({
...
+          <Button theme="tertiary" onClick={() => form.reset()}>
+            Discard
+          </Button>
+        ) : null}
```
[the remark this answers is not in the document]
sissbruecker: I think the current behavior makes sense as long as it's only a discard button. One thing that I was thinking about it whether it should rather be a "cancel" button that would actually deselect the item and close the form (which is what the CRUD component has). IMO this is a UX issue, and as mentioned in the Daily yesterday we should consult with an expert on what the UX of the auto crud should be instead of trying to figure this out ourselves. For now I'd suggest to just add the feature and clean up the UX in a second pass.

packages/ts/react-crud/src/autoform.tsx
sissbruecker: Actually very typical to put the primary action on the far right side. The CRUD component does the same. But again, a UX expert should decide this 🙂.

taefi (approved):

taefi (approved):

PR #1559, opened 2023-10-19
feat(autoform): add fields for dates, times, booleans, and enums

### LINKED ISSUE #1525: [auto-form] Editing booleans
For boolean properties auto form should add a checkbox.

### LINKED ISSUE #1526: [auto-form] Editing dates and times
For date and time properties auto-form should add a `DatePicker`, `TimePicker` or `DateTimePicker`. By default, the fields should use the date format for the browser's locale, same as the auto grid column.

### LINKED ISSUE #1528: [auto-form] Editing enums
For enum properties auto-form should render a single selection field. The field lists all possible enum values and has a useful string representation of each value. A `Select` seems like a reasonable default field to use.

cromoteca:
Closes #1525.
Closes #1526.
Closes #1528.

DateTime is not implemented, as https://github.com/vaadin/react-components/pull/152 is not merged yet. I'll create a new issue for that feature.


packages/ts/react-crud/src/autoform.tsx
```diff
@@ -17,22 +17,33 @@ type SubmitEvent<TItem> = {
...
   service: CrudService<TItem>;
   model: DetachedModelConstructor<AbstractModel<TItem>>;
   item?: TItem | typeof emptyItem;
   disabled?: boolean;
+  fieldRenderer?: ComponentType<AutoFormFieldRendererProps>;
   onSubmitError?({ error }: SubmitErrorEvent): void;
   afterSubmit?({ item }: SubmitEvent<TItem>): void;
 }>;
 
+function IdentityRenderer({ field: Field, ...props }: AutoFormFieldRendererProps) {
```
Lodin, line 35: Oops, that should be removed. It's a leftover from one of discarded attempts

cromoteca:
There are some `codecov` errors due to unused code related to the DateTimeField, which is missing due to https://github.com/vaadin/react-components/pull/151. I could remove that code, although I hope we'll soon be able to use it.

Lodin (approved):

taefi (approved):

taefi (approved):

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -43,6 +67,14 @@ describe('@hilla/react-crud', () => {
         .map(([, value]) => value.toString());
     }
 
+    function TestFieldRenderer({ field: Field, propertyInfo, ...other }: AutoFormFieldRendererProps) {
+      return <Field data-testid={propertyInfo.name} propertyInfo={propertyInfo} {...other} />;
```
sissbruecker, line 71: Why do we need test ids? Every field should have a label that we can query for. That would also implicitly test that labels are set correctly, which is missing currently.

sissbruecker (changes requested):

packages/ts/react-crud/test/autoform.spec.tsx
sissbruecker: It also seems wrong to add this field renderer abstraction to the public API of the component, just so that we can test it.

packages/ts/react-crud/src/autoform-field.tsx
```diff
@@ -1,36 +1,103 @@
...
+  // @ts-expect-error: TODO: fix the dangerouslySetInnerHTML typing issue
+  return <DateTimePicker {...other} {...form.field(model)} label={propertyInfo.humanReadableName} />;
+}
+
+type AutoFormEnumFieldProps = SelectProps & SharedFieldProps;
+
+function AutoFormEnumField({ propertyInfo, form, ...other }: AutoFormEnumFieldProps) {
+  const model = getPropertyModel(form, propertyInfo) as EnumModel;
+  const options = Object.keys(model[_enum]).map((value) => ({
+    label: value,
```
sissbruecker, line 77: Let's render a more user-friendly label here as mentioned in the [ticket](https://github.com/vaadin/hilla/issues/1528). With the current example model I would expect the following options:
- `Male`
- `Female`
- `Non binary` or `Non Binary`

packages/ts/react-crud/src/autoform-field.tsx
```diff
@@ -1,36 +1,103 @@
...
+  // @ts-expect-error: TODO: fix the dangerouslySetInnerHTML typing issue
+  return <NumberField {...other} {...form.field(model)} label={propertyInfo.humanReadableName} />;
+}
+
+type AutoFormDateFieldProps = DatePickerProps & SharedFieldProps;
+
+function AutoFormDateField({ propertyInfo, form, ...other }: AutoFormDateFieldProps) {
+  const model = getPropertyModel(form, propertyInfo);
+  // @ts-expect-error: TODO: fix the dangerouslySetInnerHTML typing issue
+  return <DatePicker {...other} {...form.field(model)} label={propertyInfo.humanReadableName} />;
```
sissbruecker, line 53: When testing with the `auto crud` view in `react-grid-test`, the `Joined date` field is not properly populated.

Bildschirmfoto 2023-10-26 um 12 16 51

sissbruecker, line 53: Date pickers should use the same date format as the one in the header filter.

Bildschirmfoto 2023-10-26 um 12 21 39

packages/ts/react-crud/test/autoform.spec.tsx
cromoteca: I removed test ids from the package. They are also used in `react-form`, but that's outside of the scope of this PR.

packages/ts/react-crud/src/autoform-field.tsx
cromoteca: Should be OK now

cromoteca: I propose to address this separately after discussing https://github.com/vaadin/hilla/issues/1615.

packages/ts/react-crud/src/autoform-field.tsx
cromoteca: I propose to implement that separately by addressing https://github.com/vaadin/hilla/issues/1617.

packages/ts/react-crud/src/property-info.ts
```diff
@@ -48,6 +59,12 @@ function determinePropertyType(model: AbstractModel): PropertyType {
   } else if (constructor === BooleanModel) {
     return 'boolean';
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if ((model as EnumModel)[_enum]) {
+    return 'enum';
+  }
```
sissbruecker, line 66: We can use `instanceof` checks here, as well as the checks above:
```ts
  // Otherwise detect by model instance
  if (model instanceof StringModel) {
    return 'string';
  } else if (model instanceof NumberModel) {
    return 'decimal';
  } else if (model instanceof BooleanModel) {
    return 'boolean';
  } else if (model instanceof EnumModel) {
    return 'enum';
  }
```

packages/ts/react-crud/src/autoform-field.tsx
```diff
@@ -1,44 +1,112 @@
...
+  DatePickerProps &
+  IntegerFieldProps &
+  NumberFieldProps &
+  SelectProps &
+  SharedFieldProps &
+  TextFieldProps &
+  TimePickerProps;
+
 export function AutoFormField(props: AutoFormFieldProps): JSX.Element | null {
+  const i18n = useDatePickerI18n();
```
sissbruecker, line 93: Would suggest to move this into `AutoFormDateField`.

sissbruecker (approved):

PR #1560, opened 2023-10-19
fix: reset form when creating new item

sissbruecker:
Clears the form and removes the selection from grid when creating a new item. Also simplifies the grid selection logic a bit.


cromoteca (approved):
Nice and clean

PR #1561, opened 2023-10-19
test: enforce en-US as default locale

sissbruecker:
Use `en-US` as default locale in auto grid tests, so that they pass if Karma starts browsers with a different locale.

### DIFF
--- a/packages/ts/react-crud/src/i18n.ts
+++ b/packages/ts/react-crud/src/i18n.ts
@@ -1,4 +1,9 @@
 // The default locale to use for renderers, filter inputs, etc.
 // If undefined, the browser's locale will be used.
 // Allows to modify the locale for testing purposes.
-export const defaultLocale: string | undefined = undefined;
+// eslint-disable-next-line import/no-mutable-exports
+export let defaultLocale: string | undefined;
+
+export function setDefaultLocale(locale: string | undefined): void {
+  defaultLocale = locale;
+}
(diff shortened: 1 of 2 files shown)


Lodin:
Actually, I went a bit different way. I want `locale` to be a property of `AutoGrid`: it is more React-ish and more familiar for users

sissbruecker:
@Lodin Seems a bit much to introduce a new feature to fix a test setup. Besides we might want to have this as part of a unified I18N API, which we don't know yet how it will look.

Lodin:
I understand that. However, I doubt that this small change could actually solve the issue. The problem is that the top-level `Intl.DateTimePicker` created with the `defaultLocale` won't change even if we use `setDefaultLocale` because they are created way earlier.

sissbruecker:
I've tested it and it works. If at some point a test is added that relies on the date format created from importing the module, then we can change it to be created lazily. There's no need to fix stuff that isn't broken.

Lodin (changes requested):
Just checked. Yes, as I thought, it does not work for [header-filter](https://github.com/vaadin/hilla/blob/main/packages/ts/react-crud/src/header-filter.tsx#L33) `DateFormatter`. It gets me `en-GB` in Headless Chrome and `en-US` in regular Chrome even if I set it to another locale. It allows all the tests passing but still is a bug that requires a fix.

So we need at least to create this `DateFormatter` instance on request, not at the top-module level.

PR #1562, opened 2023-10-19
refactor(react-crud): allow user to set locale for AutoGrid through optional context

Lodin:
The alternative solution for #1561.

I see the following advantages and disadvantages of this approach.

Advantages:
- The most dynamic support for the i18n dates & numbers inside of the `AutoGrid`.
- React-ish way to provide i18n solution so it is more familiar to developers. 

Disadvantages:
- New feature to present.
- Encapsulated within the React component, so can be inconsistent with the future global i18n approach.


packages/ts/react-crud/src/autogrid-column-context.tsx
```diff
@@ -10,12 +11,18 @@ export interface SorterState {
...
   setPropertyFilter(propertyFilter: PropertyStringFilter): void;
   sortState: SortState;
   setSortState: Dispatch<SetStateAction<SortState>>;
   customColumnOptions?: ColumnOptions;
-};
+}>;
 
 export const ColumnContext = createContext<ColumnContext | null>(null);
+
+export type GeneralColumnContext = Readonly<{
```
sissbruecker, line 24: How about using a context that can be wrapped around the whole grid instead? The default context value can have a formatter with an `undefined` locale, which uses the browser locale by default. For the tests, we can explicitly wrap the grid into a context using a fixed locale.

This would also avoid adding public API at this point, at least in a sense that the context is not as discoverable as a prop.

packages/ts/react-crud/src/locale-formatter.ts
```diff
@@ -0,0 +1,105 @@
...
+
+function tryFormatDate(formatter: Intl.DateTimeFormat, value?: string): string {
+  try {
+    return value ? formatter.format(new Date(value)) : '';
+  } catch (e) {
+    return '';
+  }
+}
+
+export class LocaleFormatter {
```
sissbruecker, line 30: `LocaleFormatter` seems like a worse name, since we're formattings dates, not locales.

packages/ts/react-crud/src/autogrid-column-context.tsx
Lodin: Yeah, actually, this context wraps the whole grid. Maybe, the naming is not good enough...

packages/ts/react-crud/src/locale-formatter.ts
Lodin: I have added a `NumberFormat` here as well because it depends on locale, so the `Date` part sounds misleading. However, I'm open to any naming suggestions 🙂

packages/ts/react-crud/src/autogrid-column-context.tsx
sissbruecker: Sorry, what I meant was: Remove the `locale` prop from auto grid, and instead wrap a context around the auto grid to provide either a formatter or a locale.

Lodin: Hmm... I was sure that React will throw an error if there is no `Provider` for the context but after you mentioned it I decided to check. And it looks like the `Provider` is optional 🤔. Let me try that option then. Sounds pretty interesting.

sissbruecker: If you don't use an explicit provider, then accessing the context will give you the default value that you need to specify when creating the context.

packages/ts/react-crud/src/locale-formatter.ts
sissbruecker: Ah ok, I missed the number formatting. Let's keep it as is then.

packages/ts/react-crud/src/autogrid-column-context.tsx
Lodin: @sissbruecker thanks, works like a charm!

packages/ts/react-crud/src/locale.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (dateInstance.getFullYear() !== year || dateInstance.getMonth() !== month || dateInstance.getDate() !== day) {
+      return undefined;
+    }
+
+    return { year, month, day };
+  }
+}
+
+export function useLocaleFormatter(): LocaleFormatter {
+  return new LocaleFormatter(useContext(LocaleContext));
```
sissbruecker, line 105: May be a minor thing, but can we wrap this into a `useMemo`?

packages/ts/react-crud/test/locale-formatter.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { expect } from '@esm-bundle/chai';
-import { type DateObject, DateFormatter } from '../src/date-formatter';
+import type { DatePickerDate } from '@hilla/react-components/DatePicker.js';
+import { LocaleFormatter } from '../src/locale.js';
 
 describe('DateFormatter', () => {
```
sissbruecker, line 5:
```suggestion
describe('LocaleFormatter, () => {
```

packages/ts/react-crud/src/header-filter.tsx
```diff
@@ -179,7 +172,15 @@ export function DateHeaderFilter(): ReactElement {
       <DatePicker
         value={filterValue}
         placeholder="Filter..."
-        i18n={datePickerI18n}
+        i18n={{
```
sissbruecker, line 175: Could also be memoized, assuming the formatter is memoized too.

Lodin: I decided that memoizing formatter using `useMemo` does not worth it since it is just a simple instance of the class with couple of objects within and does not do any complex calculations. So it is just created from the provided locale each time.

packages/ts/react-crud/src/locale.ts
Lodin: Oops, missed this comment. 

I decided that memoizing formatter does not worth it since it is just a simple instance of the class with couple of objects within and does not do any complex calculations. What do you think?

sissbruecker: It seems like an easy optimization and avoids:
- Doing the regex format extraction on every render
- Changing the `i18n` property of date pickers on every render, which could have unexpected side-effects. Couldn't observe anything from some quick testing, but the component does quite a lot of things when `i18n` changes, so it might be safer to avoid that.

Lodin: Ok, got it. Gonna add it then

Lodin: Done!

sissbruecker (approved):

PR #1582, opened 2023-10-23
feat: formatting of decimals in auto grid

sissbruecker:
Auto grid currently renders every number as integer, removing all fraction digits. This change introduces a decimal formatter so that decimals are rendered with two fraction digits by default.


packages/ts/react-crud/src/locale.ts
```diff
@@ -47,10 +48,15 @@ export class LocaleFormatter {
...
 
-    this.#number = new Intl.NumberFormat(locale, {
+    this.#integer = new Intl.NumberFormat(locale, {
       maximumFractionDigits: 0,
     });
 
+    this.#decimal = new Intl.NumberFormat(locale, {
+      maximumFractionDigits: 2,
+      minimumFractionDigits: 2,
+    });
```
taefi, line 58: Could this be customizable?

packages/ts/react-crud/test/autogrid.spec.tsx
```diff
@@ -531,70 +531,76 @@ describe('@hilla/react-crud', () => {
...
+      it('renders integers as right aligned numbers', () => {
         expect(grid.getBodyCellContent(0, 1)).to.have.style('text-align', 'end');
         expect(grid.getBodyCellContent(0, 1)).to.have.rendered.text('123,456');
         expect(grid.getBodyCellContent(1, 1)).to.have.rendered.text('-12');
       });
 
-      it('renders booleans as icons', () => {
-        expect(grid.getBodyCellContent(0, 2).querySelector('vaadin-icon')).to.have.attribute('icon', 'lumo:checkmark');
-        expect(grid.getBodyCellContent(1, 2).querySelector('vaadin-icon')).to.have.attribute('icon', 'lumo:minus');
+      it('renders decimals as right aligned numbers', () => {
```
taefi, line 540: Could we have a couple of more tests what is happening in the following scenarios:

* formatting 123.784
* formatting 123.7 (would there be zero padding?)
* formatting 0.1 (would there be zero padding?)

taefi (changes requested):

packages/ts/react-crud/test/autogrid.spec.tsx
sissbruecker: Sure, added two more cases for zero padding.

packages/ts/react-crud/src/locale.ts
sissbruecker: It could, but this PR is really just meant to provide a reasonable default formatting.

taefi (approved):

Issue #1575, opened 2023-10-22
`react-crud` in Hilla 2.3 depends on RC versions of React Components and Lumo


cromoteca:
### Describe the bug

This causes `Tried to define vaadin-lumo-styles version 24.2.0 when version 24.2.0-rc1 is already in use. Something will probably break.`

### Reproduction

See https://github.com/vaadin/hilla/blob/a7eb92e82d895ae1ff0e203cd032922b62d30ff0/packages/ts/react-crud/package.json#L51-L53

### System Info

Hilla 2.3.1

cromoteca:
This is also visible in the dev tools:
![image](https://github.com/vaadin/hilla/assets/498086/55bee484-00d2-4fd2-9b9b-6a1004c4754f)

Artur-:
Fixed by #1585

PR #1576, opened 2023-10-22
chore: remove RC versions

cromoteca:
Remove some RC versions found in the project

PR #1585, opened 2023-10-23
fix: bump react-crud dependencies

sissbruecker:
Fixes https://github.com/vaadin/hilla/issues/1575

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2626,67 +2626,67 @@
       "link": true
     },
     "node_modules/@hilla/react-components": {
-      "version": "2.2.0-rc1",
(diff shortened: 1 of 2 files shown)


cromoteca:
I made a similar PR for `main`: https://github.com/vaadin/hilla/pull/1576, but I have the same validation error

sissbruecker:
The problem seems to be that it now creates a separate `node_modules` into `packages/ts/react-crud`. With that it ends up with duplicates of `@hilla/react-components`. 

I guess I did something wrong with bumping the versions, it should only install dependencies into the root `node_modules` but not into the individual packages.

PR #1586, opened 2023-10-23
Add support for properties in parent classes

### LINKED ISSUE #1579: [autogrid] show properties from parent classes
### Describe your motivation

When using AutoGrid in the Pet Clinic applications, veterinarians don't have first and last name since they come from the `Person` parent class. They should be available in the grid by default.

See https://github.com/cromoteca/hilla-petclinic/blob/10fc9164cbb29629b35bebe8e2cdf52ab86bdd4d/src/main/java/org/springframework/samples/petclinic/vet/Vet.java#L46

Hilla 2.3.1

cromoteca:
Fixes #1579

### DIFF
--- a/packages/ts/react-crud/src/property-info.ts
+++ b/packages/ts/react-crud/src/property-info.ts
@@ -65,7 +65,7 @@ export const getPropertyIds = (model: DetachedModelConstructor<AbstractModel>): 
 
   for (let proto = model; proto !== ObjectModel; proto = Object.getPrototypeOf(proto)) {
     // the choice of putting parent properties at the beginning is deliberate, `push` could be used instead
-    propertyIds.unshift(...Object.keys(Object.getOwnPropertyDescriptors(proto.prototype)));
+    propertyIds.unshift(...Object.keys(Object.getOwnPropertyDescriptors(proto.prototype)).filter((p) => p !== 'new'));
   }
 
   return propertyIds;
--- a/packages/ts/react-crud/src/property-info.ts
+++ b/packages/ts/react-crud/src/property-info.ts
@@ -7,6 +7,7 @@ import {
   type AbstractModel,
   type DetachedModelConstructor,
   type ModelMetadata,
+  ObjectModel,
 } from '@hilla/form';
 
 export type PropertyType = 'boolean' | 'date' | 'datetime' | 'number' | 'string' | 'time' | undefined;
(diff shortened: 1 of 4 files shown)


packages/ts/react-crud/src/property-info.ts
```diff
@@ -59,34 +60,47 @@ export function _generateHeader(path: string): string {
...
-      type,
-      meta,
-    };
-  });
+  return (
+    propertyIds
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      .map((name) => [name, modelInstance[name]])
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      .filter(([, m]) => m?.[_meta])
```
sissbruecker, line 83: Is the filter here necessary? Metadata is in general optional and there are specific conditions that determine whether it is generated at all.

packages/ts/react-crud/src/property-info.ts
```diff
@@ -59,34 +60,47 @@ export function _generateHeader(path: string): string {
     .replace(/^./u, (match) => match.toUpperCase());
 }
 
-export const getPropertyIds = (model: DetachedModelConstructor<AbstractModel>): string[] =>
-  Object.keys(Object.getOwnPropertyDescriptors(model.prototype)).filter((p) => p !== 'constructor');
+export const getPropertyIds = (model: DetachedModelConstructor<AbstractModel>): string[] => {
+  const propertyIds: string[] = [];
+
+  for (let proto = model; proto !== ObjectModel; proto = Object.getPrototypeOf(proto)) {
+    // the choice of putting parent properties at the beginning is deliberate, `push` could be used instead
```
sissbruecker, line 67: Explaining what the alternatives could be is unnecessary and just adds cognitive load. Better to just explain what it does.

packages/ts/react-crud/src/property-info.ts
cromoteca: I'm filtering out those who don't have metadata at all

packages/ts/react-crud/src/property-info.ts
cromoteca:
```suggestion
    // parent properties are added at the beginning
```

sissbruecker: Looks good 👍

packages/ts/react-crud/src/property-info.ts
cromoteca: The original code (before this PR) assumes that `meta` is there. It maybe is, but TypeScript complains if I remove this check.

sissbruecker: I just tried to remove it, I can not observe any errors.

sissbruecker (approved):

sissbruecker (approved):

Issue #1425, opened 2023-10-03
[autocrud] Mobile layout support


Artur-:
On mobile, you should not see a grid and form side by side but it should always be a popup style form that covers the full screen

sissbruecker:
Has been resolved with #1605

PR #1605, opened 2023-10-25
feat: implement mobile layout

sissbruecker:
Implements a mobile layout for auto crud that shows the form in a dialog overlay. This also rearranges the general layout a bit:
- Remove the option for a custom header, which is easy enough to add on your own, and with proper header semantics instead of a hard-coded `h2`
- Move the new button into a toolbar below the grid, similar to `vaadin-crud`
- The auto form within the auto-crud or its mobile overlay now gets a scrollable area for the fields, as well as a matching toolbar

Some things I want to address in follow-up PRs:
- Make dialog title dynamic between "Edit item" and "Create item"
- Add a fixed column with edit buttons
- Maybe only show the form if something is being edited

Part of https://github.com/vaadin/hilla/issues/1425

Basic layout:

Bildschirmfoto 2023-10-25 um 11 45 32

Mobile overlay:

Bildschirmfoto 2023-10-25 um 11 46 03


packages/ts/react-crud/src/autocrud.tsx
```diff
@@ -1,55 +1,61 @@
...
   model: DetachedModelConstructor<AbstractModel<TItem>>;
   noDelete?: boolean;
-  header?: string;
 }>;
 
-export function ExperimentalAutoCrud<TItem>({ service, model, noDelete, header }: AutoCrudProps<TItem>): JSX.Element {
+export function ExperimentalAutoCrud<TItem>({ service, model, noDelete }: AutoCrudProps<TItem>): JSX.Element {
   const [item, setItem] = useState<TItem | typeof emptyItem | undefined>(undefined);
   const [refreshTrigger, setRefreshTrigger] = useState(0);
+  const fullScreen = useMediaQuery('(max-width: 600px), (max-height: 600px)');
```
sissbruecker, line 23: This is the same query that `vaadin-crud` uses:
https://github.com/vaadin/web-components/blob/735f5d90df1618c700bfc7be0d9c8ad0cd8436dc/packages/crud/src/vaadin-crud.js#L627-L629

packages/ts/react-crud/src/styles.css
sissbruecker, line 1: Extracted the styles into a CSS file as suggested by Vlad.

The CRUD layout styles are mostly adapted from the web component.

packages/ts/react-crud/test/autocrud.spec.tsx
```diff
@@ -14,15 +17,31 @@ use(chaiDom);
...
+    before(() => {
+      matchMediaStub = sinon.stub(window, 'matchMedia');
+      matchMediaStub.returns({
+        matches: false,
+        addListener: () => {},
+        removeListener: () => {},
+        addEventListener: () => {},
+        removeEventListener: () => {},
+      });
+    });
```
sissbruecker, line 31: Since we run tests headless by default we need to stub `matchMedia` for all tests, otherwise `matches` always returns `true`.

packages/java/tests/spring/react-grid-test/src/test/java/dev/hilla/test/reactgrid/AutoCrudViewIT.java
sissbruecker, line 1: With the custom header option removed, there are no ITs remaining for auto crud 😳

packages/ts/react-crud/src/index.ts
```diff
@@ -1,3 +1,6 @@
+// eslint-disable-next-line import/no-unassigned-import
+import './styles.css';
```
Lodin, line 2: Hmm, what about making `.module.css` here? We could rely on Vite to bundle separate files and import `style.css` in `index.ts`. So we will be free to use any kind of class transformation.

packages/ts/react-crud/src/header-filter.tsx
```diff
@@ -110,7 +93,7 @@ function ComparationSelection({ onMatcherChanged, value }: ComparationSelectionP
           </Item>
         </ListBox>
       )}
-      className={autoGridFilterWithLessGreaterEqualsStyle.id}
+      className="autoGridFilterWithLessGreaterEquals"
```
Lodin, line 96: Can this class be in kebab-case as well as all others?

packages/ts/react-crud/src/styles.css
```diff
@@ -0,0 +1,83 @@
...
+  align-items: stretch;
+}
+
+.auto-crud .auto-form {
+  width: 40%;
+  border: solid 1px var(--lumo-contrast-20pct);
+  border-left-width: 0;
+}
+
+.auto-crud .auto-form-fields, .auto-crud-dialog .auto-form-fields {
```
Lodin, line 35: Could we use more BEM approach here? There are too many nested classes here IMO.

packages/ts/react-crud/src/index.ts
sissbruecker: As discussed, will leave as is for now.

packages/ts/react-crud/src/header-filter.tsx
sissbruecker: Done

packages/ts/react-crud/src/styles.css
sissbruecker: Simplified selectors where class names were redundant.

packages/ts/react-crud/src/media-query.ts
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react';
+
+export function useMediaQuery(query: string): boolean {
+  const [matches, setMatches] = useState(() => window.matchMedia(query).matches);
```
Lodin, line 4: What about using [`useMedia`](https://github.com/streamich/react-use/blob/master/docs/useMedia.md) from `react-use` here? That would reduce the amount of code we have to support.

packages/ts/react-crud/test/FormController.ts
```diff
@@ -8,20 +8,21 @@ export type FormElement = HTMLElement & {
...
-  readonly #result: RenderResult;
+  readonly #result: FormQueries;
   readonly #user: ReturnType<(typeof userEvent)['setup']>;
 
-  static async init(result: RenderResult, user: ReturnType<(typeof userEvent)['setup']>): Promise<FormController> {
-    const form = (await waitFor(
-      () => result.container.querySelector('vertical-layout[theme="padding"]')!,
-    )) as HTMLElement;
+  static async init(result: FormQueries, user: ReturnType<(typeof userEvent)['setup']>): Promise<FormController> {
+    const form = await waitFor(async () => result.findByTestId('auto-form'));
```
Lodin, line 19: @taefi used search by `vaadin-form-layout`. What do you think about using it instead of `data-testid`?

Lodin (changes requested):

packages/ts/react-crud/src/media-query.ts
sissbruecker: I'd prefer not to add a new dependency for a single hook. The argument is not really sound, as a dependency is something you have to maintain.

packages/ts/react-crud/test/FormController.ts
sissbruecker: There is no `vaadin-form-layout` at this point. From a testing best practices POV it doesn't make a difference as the user interacting with the app doesn't care about what kind of element is rendered. I'd suggest to leave as is, I'm fine with it being changed in a subsequent PR.

packages/ts/react-crud/src/media-query.ts
Lodin: Ok, let's have this implementation for now

Lodin (approved):

Issue #1617, opened 2023-10-26
[autocrud] use a better textual representation of enum entries


cromoteca:
Currently, the uppercase value of enum entries is shown in AutoGrid and AutoForm. Let's use title case or sentence case.

PR #1626, opened 2023-10-27
feat: improve enum support

sissbruecker:
Adds following improvements to enum support:
- Format labels in auto enum form fields in title case
- Format enum values in auto grid cells in title case
- Add enum filter support for auto grid

Fixes https://github.com/vaadin/hilla/issues/1617


packages/java/endpoint/src/main/java/dev/hilla/crud/PropertyStringFilterSpecification.java
```diff
@@ -148,6 +148,24 @@ public Predicate toPredicate(Root<T> root, CriteriaQuery<?> query,
             default:
                 break;
             }
+        } else if (javaType.isEnum()) {
+            Enum<?> enumValue = parseEnumValue(javaType, value);
```
cromoteca, line 152:
```suggestion
            var enumValue = Enum.valueOf(javaType.asSubclass(Enum.class), value);
```
And then you can remove `parseEnumValue`.

cromoteca (changes requested):

cromoteca (approved):

cromoteca:
Ideally, the tests should cover an unsupported type, although this is not in the scope of this PR. But I think that enums played that role before, by accident.

PR #1627, opened 2023-10-27
chore: add CSS import system

Lodin:
This PR sets up a system for importing CSS in React packages. Each CSS file is transformed into a JS that construct a `CSSStyleSheet` with the minified CSS code.

The system works for production building and for testing.

Also, PR depends on #1605 because it implements some CSS stuff already.


scripts/build.ts
```diff
@@ -1,29 +1,65 @@
...
 const packageJson: PackageJson = JSON.parse(packageJsonFile);
 
+const cssTransformer = postcss([cssnanoPlugin()]);
+const cssPath = /['"](.+)\.obj\.css['"]/gmu;
+const cssConstructPlugin: Plugin = {
+  name: 'construct-css',
+  setup(_build) {
+    // Here CSS imports like `import css from 'autocrud.obj.css';` are transformed to
+    // JS imports: `import css from 'autocrud.obj.js'`.
+    _build.onLoad({ filter: /\.tsx?/u }, async ({ path }) => {
```
Artur-, line 26: This looks like duplicating the CSS import feature from Vite?

Lodin: Not really. It's an artificial import of `CSSStyleSheet`. I wish I could use import attributes but they are not correctly implemented yet.

Artur-: So why does Hilla need this when Vite already implements support for css imports?

Lodin: Because it's hard to control CSS imports in libraries. Usually, React either uses a CSS-in-JS approach (e.g. [material-ui](https://mui.com)) or leaves importing pure CSS files to the user (e.g. [rmwc](https://rmwc.io). Both these approaches have the cost.

1) CSS-in-JS usually has significant performance impact even when pre-compiled. Also, I can see that it looses popularity since one of it's killer features was the context-depending styling which is gone with CSS properties spreading.

2) Importing CSS files by user is better in performance but it also requires additional actions from the user. That's ok only if you don't have an alternative.

We could probably try to import CSS in `index.ts` and leave handling all the building stuff to the user but along with additional actions it also could break the CSS ordering, so the user would have to use `!important` everywhere.

So I chose constructing a `CSSStyleSheet` instance. It is simple, supported by all the modern browsers, doesn't require anything from the user and could handle the ordering by simply put the imported style to the beginning of the style sheets list.

platosha (approved):

PR #1628, opened 2023-10-27
feat: add actions column to auto crud

sissbruecker:
Adds an actions column to auto crud that has buttons for editing and deleting an item. This improves UX for the mobile layout where there is currently no indication that items are editable.

Bildschirmfoto 2023-10-26 um 14 25 54


cromoteca:
Ideally, the tests should cover an unsupported type, although this is not in the scope of this PR. But I think that enums played that role before, by accident.

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -198,9 +200,11 @@ export function ExperimentalAutoForm<M extends AbstractModel>({
   }
 
   return (
-    <section className="flex flex-col p-m gap-m">
-      {layout}
-      {formError ? <div style={{ color: 'var(--lumo-error-color)' }}>{formError}</div> : <></>}
+    <VerticalLayout className="auto-form" theme="spacing">
```
sissbruecker, line 203: Some opinionated changes here:
- Let's not use utility classes in components. There is no guarantee that those are loaded in an application which silently breaks things. Let's only assume that Lumo CSS properties are there.
- Don't add any sort of margin or padding to the outer element of a component. That is easy enough to add on the application-side, plus it requires customization to remove it if you want to do anything practical, for example show a heading and the form with the same horizontal alignment.
- Using `section` doesn't provide any benefit here as it's lacking a heading.

Also restored the classes that are required for properly styling the component when used in auto crud.

packages/ts/react-crud/src/styles.css
sissbruecker, line 1: Removed this as it's no longer used since https://github.com/vaadin/hilla/pull/1627

packages/ts/react-crud/src/autocrud.obj.css
```diff
@@ -8,6 +8,22 @@
...
+  right: 0;
+  padding: 0 var(--lumo-space-xs);
+  display: flex;
+  gap: var(--lumo-space-xs);
+  align-items: center;
+}
+
+.auto-crud-actions vaadin-button {
+  margin: 0;
+}
```
sissbruecker, line 25: Elaborate styles to minimize the padding, and thus the width, of the column which has a tricky to customize padding otherwise.

cromoteca:
Code is OK for me, but by trying it in the browser (also with phone emulation), I cannot see any difference between clicking on the row or on the corresponding edit icon.

sissbruecker:
> Code is OK for me, but by trying it in the browser (also with phone emulation), I cannot see any difference between clicking on the row or on the corresponding edit icon.

@cromoteca I did not add any different behavior there, you can tap the row on mobile to open the editor overlay, same as on desktop. The intention with the column was more to give an indication that there is something editable in the first place, otherwise it would just look like a regular grid on mobile. Do you think the behavior should be different somehow?

cromoteca (approved):
As discussed, this PR is an improvement compared to the current implementation and we can further improve it later after getting some feedback.

PR #1637, opened 2023-11-01
chore(test,autogrid): remove hardcoded test column numbers

cromoteca:


packages/ts/react-crud/test/autogrid.spec.tsx
```diff
@@ -158,10 +158,12 @@ describe('@hilla/react-crud', () => {
       it('data provider provides data', async () => {
         const grid = await GridController.init(render(<TestAutoGridNoHeaderFilters />), user);
         expect(grid.getVisibleRowCount()).to.equal(2);
-        expect(grid.getBodyCellContent(0, 0)).to.have.rendered.text('Jane');
-        expect(grid.getBodyCellContent(0, 1)).to.have.rendered.text('Love');
-        expect(grid.getBodyCellContent(1, 0)).to.have.rendered.text('John');
-        expect(grid.getBodyCellContent(1, 1)).to.have.rendered.text('Dove');
+        const firstNameRowIndex = await grid.findColumnIndexByHeaderText('First name');
```
taefi, line 161: I think all of the local variables named as `xxxRowIndex` should probably be called `xxxColIndex`:
```suggestion
        const firstNameColIndex = await grid.findColumnIndexByHeaderText('First name');
```

taefi (changes requested):

packages/ts/react-crud/test/autogrid.spec.tsx
cromoteca: Definitely

taefi (approved):

PR #1644, opened 2023-11-02
feat(autoform): support custom label via field options

### LINKED ISSUE #1568: [autoform] support custom labels for auto created fields
### Describe your motivation

Developers should be able to define custom labels for form fields. 

This can be addressed similar to what `<AutoGrid />` does with `columnOptions`.

_No response_

### Additional context

_No response_

taefi:
## Description

This allows setting custom labels for AutoForm fields through `fieldOptions` props. 
This also adopts the possibility of overriding the default field renderer through `fieldOptions` props and removes the so called `customFields` property.

Fixes #1568

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


packages/ts/react-crud/src/autoform-field.tsx
```diff
@@ -93,6 +98,9 @@ export type AutoFormFieldProps = CheckboxProps &
   TimePickerProps;
 
 export function AutoFormField(props: AutoFormFieldProps): JSX.Element | null {
+  if (props.options?.label) {
+    props.propertyInfo.humanReadableName = props.options.label;
+  }
```
cromoteca, line 103: I think that this is controversial, especially given that we already know that `options` will also replace `customFields`. In fact, doing both in the same PR would make it easier to think about the best approach for merging default values with the ones coming from options.

sissbruecker (approved):

packages/ts/react-crud/src/autoform-field.tsx
```diff
@@ -25,62 +31,62 @@ function getPropertyModel(form: UseFormResult<any>, propertyInfo: PropertyInfo)
 
 type AutoFormTextFieldProps = SharedFieldProps & TextFieldProps;
 
-function AutoFormTextField({ propertyInfo, form, ...other }: AutoFormTextFieldProps) {
+function AutoFormTextField({ propertyInfo, form, options, label, ...other }: AutoFormTextFieldProps) {
```
platosha, line 34: `options` is not mentioned below, was it intentional?

platosha (changes requested):

packages/ts/react-crud/src/autoform-field.tsx
taefi: Yes, at this point we are not using any of the options in those set of functions, so adding it would result in `unused` ES Lint errors.

platosha (approved):

PR #1661, opened 2023-11-06
feat: add delete button to auto form

sissbruecker:
Based on the UX feedback by Rolf, we want to remove the actions column from auto crud as it takes up horizontal space and creates a lot of visual noise. This change moves the delete button to the auto form and removes the actions column. Showing the delete button is optional and disabled by default. The delete button is only visible when editing an existing item.


packages/ts/react-crud/src/autoform.tsx
```diff
@@ -40,8 +49,11 @@ export type AutoFormProps<M extends AbstractModel = AbstractModel> = Readonly<{
   disabled?: boolean;
   customLayoutRenderer?: AutoFormLayoutProps | ComponentType<AutoFormLayoutRendererProps<M>>;
   fieldOptions?: Record<string, FieldOptions>;
+  withDeleteButton?: boolean;
```
sissbruecker, line 52: Naming suggestions?

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -203,26 +253,44 @@ export function ExperimentalAutoForm<M extends AbstractModel>({
...
         {layout}
         {formError ? <div style={{ color: 'var(--lumo-error-color)' }}>{formError}</div> : <></>}
       </VerticalLayout>
-      <HorizontalLayout className="auto-form-toolbar" theme="spacing">
-        {form.dirty ? (
-          <Button theme="tertiary" onClick={() => form.reset()}>
-            Discard
-          </Button>
-        ) : null}
+      <div className="auto-form-toolbar">
```
sissbruecker, line 261: The buttons here are rendered in the reverse order than they will be displayed. This has the benefit that when using keyboard navigation, the first button that gets focused after the fields is the save button, then the discard button, and finally the delete button. Visually the delete button is aligned to the left, and the save and discard buttons on the right. This basically reproduces how `vaadin-crud` renders and aligns the buttons.

packages/ts/react-crud/src/autoform.tsx
cromoteca: I'd use `deleteButtonVisible`: https://vaadin.com/docs/latest/components/confirm-dialog#buttons

sissbruecker: Thanks, done.

cromoteca (approved):
I reviewed the code and tested the feature locally, LGTM

PR #1675, opened 2023-11-08
feat: make auto crud panels resizable

sissbruecker:
Makes the panels in auto crud resizable by wrapping them in a split layout. The split layout is only used for the desktop layout.

### DIFF
--- a/packages/ts/react-crud/src/autocrud.obj.css
+++ b/packages/ts/react-crud/src/autocrud.obj.css
@@ -1,28 +1,23 @@
 .auto-crud {
   display: flex;
   overflow: hidden;
+  border: solid 1px var(--lumo-contrast-20pct);
+}
+
+.auto-crud vaadin-split-layout::part(splitter) {
+  border-top: solid 1px var(--lumo-contrast-20pct);
+  border-left: solid 1px var(--lumo-contrast-20pct);
 }
 
 .auto-crud-main {
   flex: 1 1 100%;
+  min-width: 200px;
   display: flex;
   flex-direction: column;
 }
 
-.auto-crud-actions {
-  position: absolute;
-  top: 0;
-  bottom: 0;
-  left: 0;
-  right: 0;
-  padding: 0 var(--lumo-space-xs);
-  display: flex;
-  gap: var(--lumo-space-xs);
-  align-items: center;
-}
-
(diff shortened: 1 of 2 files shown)


packages/ts/react-crud/src/autocrud.obj.css
```diff
@@ -1,28 +1,23 @@
...
+}
+
+.auto-crud vaadin-split-layout::part(splitter) {
+  border-top: solid 1px var(--lumo-contrast-20pct);
+  border-left: solid 1px var(--lumo-contrast-20pct);
 }
 
 .auto-crud-main {
   flex: 1 1 100%;
+  min-width: 200px;
```
sissbruecker, line 14: Using a min-width for the main panel containing the grid so that you can't collapse it completely.

packages/ts/react-crud/src/autocrud.obj.css
```diff
@@ -1,28 +1,23 @@
...
   display: flex;
   flex-direction: column;
 }
 
-.auto-crud-actions {
```
sissbruecker, line 12: Removed leftover styles from the removed actions column.

packages/ts/react-crud/src/autocrud.obj.css
```diff
@@ -46,10 +40,19 @@
 
 .auto-crud .auto-form {
   width: 40%;
-  border: solid 1px var(--lumo-contrast-20pct);
-  border-left-width: 0;
-  box-shadow: var(--lumo-box-shadow-s);
+  min-width: 300px;
```
sissbruecker, line 43: Also use a min-width for the form panel. The current value ensures the form's toolbar doesn't overflow when all three buttons are visible.

packages/ts/react-crud/src/autocrud.obj.css
```diff
@@ -46,10 +40,19 @@
...
+}
+
+/* Move box shadow and required z-index modification into pseudo-element
+   as it otherwise messes with the drag handle of the split layout */
+.auto-crud .auto-form::before {
```
sissbruecker, line 48: Adding a z-index to the form caused the splitter handle to move to the left. This seems to work as an alternative for applying the box shadow.

cromoteca (approved):

PR #1680, opened 2023-11-08
feat(autoform): simplify basic form layout settings

### LINKED ISSUE #1678: [autoform] simplify basic form layout settings
It seems that the multi-dimensional template API is not something that we want to support for a long run. At the same time, exposing the renderer for the form layout does not cover simple cases well, such as:

- Defining which fields are visible
- Defining the order of the fields
- Changing the number of columns

Let us replace the form layout template API and our custom layout mechanism with the default `<FormLayout>` and these two properties in `<AutoForm>`:

- `visibleFields: string[]` that defines the list of visible fields and the order of them in the default `<FormLayout>`
- `formLayoutProps` that gets passed down to the default `<FormLayout>` that renders the field list.

Aside from this, let us keep the `fieldOptions` property for passing the props to the individual field components. Make sure that `colspan` could be defined using `fieldOptions`.

In addition, rename `customLayoutRenderder` to `layoutRenderer`. The only supported value format in this property should be the callback / component.

cromoteca:
The way `AutoForm` is customized changes:

- `visibleFields` is used to list fields that are visible in the form (and their order)
- `formLayoutProps` is used to pass `responsiveSteps` and `ComponentStyleProps` properties to the default `FormLayout`
- `colspan` is added to `fieldOptions` and passed to the default `FormLayout`
- `customLayoutRenderder` is renamed to `layoutRenderer`

Closes #1678

### DIFF
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -4,7 +4,7 @@ import { ConfirmDialog } from '@hilla/react-components/ConfirmDialog';
 import { FormLayout } from '@hilla/react-components/FormLayout';
 import { VerticalLayout } from '@hilla/react-components/VerticalLayout.js';
 import { useForm, type UseFormResult } from '@hilla/react-form';
-import { type ComponentType, type JSX, type ReactElement, useEffect, useState, cloneElement } from 'react';
+import { type ComponentType, type JSX, type ReactElement, useEffect, useState } from 'react';
 import { AutoFormField, type AutoFormFieldProps, type FieldOptions } from './autoform-field.js';
 import css from './autoform.obj.css';
 import type { CrudService } from './crud.js';
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -33,16 +33,6 @@ export type AutoFormLayoutRendererProps<M extends AbstractModel> = Readonly<{
   children: ReadonlyArray<ReactElement<AutoFormFieldProps>>;
 }>;
 
-type FieldColSpan = Readonly<{
-  property: string;
-  colSpan: number;
-}>;
-
-export type AutoFormLayoutProps = Readonly<{
-  template: FieldColSpan[][] | string[][];
-  responsiveSteps?: Array<{ minWidth: string; columns: number }>;
-}>;
-
 export type AutoFormProps<M extends AbstractModel = AbstractModel> = ComponentStyleProps &
   Readonly<{
     /**
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -4,12 +4,12 @@ import { ConfirmDialog } from '@hilla/react-components/ConfirmDialog';
 import { FormLayout } from '@hilla/react-components/FormLayout';
 import { VerticalLayout } from '@hilla/react-components/VerticalLayout.js';
 import { useForm, type UseFormResult } from '@hilla/react-form';
-import React, { type ComponentType, type JSX, type ReactElement, useEffect, useState } from 'react';
+import { type ComponentType, type JSX, type ReactElement, useEffect, useState, cloneElement } from 'react';
 import { AutoFormField, type AutoFormFieldProps, type FieldOptions } from './autoform-field.js';
 import css from './autoform.obj.css';
 import type { CrudService } from './crud.js';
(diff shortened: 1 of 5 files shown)


packages/ts/react-crud/src/autoform.tsx
```diff
@@ -313,29 +234,37 @@ export function ExperimentalAutoForm<M extends AbstractModel>({
...
-    }
-  }
+  const defaultProperties = getProperties(model);
+
+  const visibleProperties = visibleFields
+    ? visibleFields.reduce<PropertyInfo[]>((foundProperties, propertyName) => {
+        const maybeProperty = defaultProperties.find((p) => p.name === propertyName);
+        return maybeProperty ? [...foundProperties, maybeProperty] : foundProperties;
+      }, [])
+    : defaultProperties.filter(includeProperty);
```
sissbruecker, line 259: It looks like this would allow you to specify properties that `AutoFormField` doesn't know how to render. In comparison, in auto grid we first filter the default properties on whether they may be included, and only from that derive the visible columns.

OTOH you can provide a custom renderer with `fieldOptions`, so still a valid use case I guess?

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -408,194 +401,43 @@ describe('@hilla/react-crud', () => {
...
-      expect(form.formLayout).to.not.exist;
-      const layout = await waitFor(() => form.renderResult.querySelector('vaadin-vertical-layout')!);
-      expect(layout).to.exist;
+    describe('visibleFields', () => {
+      it('renders the form according to visibleFields if defined', async () => {
+        const form = await populatePersonForm(1, { visibleFields: ['firstName', 'lastName', 'id', 'dummy'] });
+        const fields = await form.getFields('First name', 'Last name', 'Id');
+        const tagNames = fields.map((field) => field.localName);
+        expect(tagNames).to.eql(['vaadin-text-field', 'vaadin-text-field', 'vaadin-number-field']);
+        expect(form.queryField('Dummy')).to.be.undefined;
```
sissbruecker, line 439: Could also verify that some of the other properties are not rendered. Currently it just asserts that:
- There are fields for the three existing properties specified in `visibleFields`
- There is no field for the non-existing property

But it does not verify that no fields have been generated for the other existing properties that have not been specified in `visibleFields`.

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -408,194 +401,43 @@ describe('@hilla/react-crud', () => {
...
-        'screen-1024-768',
-        true,
-      );
-
-      expect(await form.getField('First name')).to.have.attribute('disabled');
-    });
+    describe('layoutRenderer', () => {
+      it('uses default two-column form layout if layoutRenderer is not defined', async () => {
+        const form = await populatePersonForm(1);
+        expect(form.formLayout.responsiveSteps).to.have.length(3);
```
sissbruecker, line 407: Nit: I think this is superfluous together with the deep equal check below.

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -4,12 +4,12 @@ import { ConfirmDialog } from '@hilla/react-components/ConfirmDialog';
 import { FormLayout } from '@hilla/react-components/FormLayout';
 import { VerticalLayout } from '@hilla/react-components/VerticalLayout.js';
 import { useForm, type UseFormResult } from '@hilla/react-form';
-import React, { type ComponentType, type JSX, type ReactElement, useEffect, useState } from 'react';
+import { type ComponentType, type JSX, type ReactElement, useEffect, useState, cloneElement } from 'react';
```
sissbruecker, line 7:
```suggestion
import { type ComponentType, type JSX, type ReactElement, useEffect, useState } from 'react';
```

packages/ts/react-crud/src/autoform.tsx
cromoteca: Yes, `AutoForm` will try to include all fields listed in `visibleFields`. Anyway, if there's no custom renderer, the `AutoFormField` will be undefined for unhandled properties, so the filtering you mentioned still happens.

packages/ts/react-crud/test/autoform.spec.tsx
cromoteca: Good point

packages/ts/react-crud/src/autoform.tsx
sissbruecker: Right, should be fine then.

sissbruecker (approved):

PR #1693, opened 2023-11-10
chore: add lock files update script

platosha:


cromoteca:
It's really good news to finally have this!
I have a question though: I see that this build has been run 12 times yesterday, 6 of them on main. Looks like it will create 6 PRs, right? Unless we're fast enough to merge the first one.
One possible solution would be:

use a branch name which only depends on the original branch (i.e. without the date.now)
if the branch doesn't exist, create it and the PR
if the branch exists, force-push on it

platosha:
> One possible solution would be:
> 
> * use a branch name which only depends on the original branch (i.e. without the date.now)
> * if the branch doesn't exist, create it and the PR
> * if the branch exists, force-push on it

@cromoteca thanks, good suggestion. Now it can force-push like so: https://github.com/vaadin/hilla/pull/1716#event-10941756380

cromoteca (approved):
Code looks fine to me, although now there's no way to test it immediately.

PR #1696, opened 2023-11-10
refactor: expose style modules for importing

sissbruecker:
Exposes the style modules in `react-crud`, so that they can be imported from the `@hilla/react-crud` package.

Also fixes the CSS priority by inserting inline stylesheets into the document's head instead of using adopted stylesheets. This allows application styles from a theme's `styles.css` to override the default styles.


packages/ts/react-crud/src/util.ts
```diff
@@ -12,8 +18,12 @@ export function convertToTitleCase(inputString: string): string {
...
-  className?: string;
-}>;
+export function registerStylesheet(stylesheet: CSSStyleSheet): void {
+  const css = Array.from(stylesheet.cssRules)
+    .map((rule) => rule.cssText)
+    .join('\n');
+
+  const styleTag = document.createElement('style');
+  styleTag.textContent = css;
+  document.head.insertAdjacentElement('afterbegin', styleTag);
```
Lodin, line 28: Would the `document.head.prepend` work here? It's shorter.

sissbruecker: Yep, that works as well. Changed.

PR #1715, opened 2023-11-13
refactor: expose auto form event types

sissbruecker:
Expose event types for auto form so they can be imported by the application.

PR #1717, opened 2023-11-13
fix(frontend+autoform): use proper error types

cromoteca:
Fixes `frontend` error message logic.
Proposes a way to handle error types in Auto Form.

It works better if endpoints throw `EndpointException`s, as in that case it's able to get back the message of the original exception, but also works reasonably for other exceptions.

If this can be a good solution, I'll work on the tests.

Conflicts with #1715, we should make a choice.

### DIFF
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -223,9 +223,14 @@ export function AutoForm<M extends AbstractModel>({
         afterDelete({ item: deletedItem });
       }
     } catch (error) {
-      if (onDeleteError && error instanceof EndpointError) {
-        onDeleteError({ error });
+      if (error instanceof EndpointError) {
+        if (onDeleteError) {
+          onDeleteError({ error });
+        } else {
+          setFormError(error.message);
+        }
       }
+      throw error;
     } finally {
       setShowDeleteDialog(false);
     }
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -1,4 +1,5 @@
 import { type AbstractModel, type DetachedModelConstructor, ValidationError, type Value } from '@hilla/form';
+import { EndpointError } from '@hilla/frontend';
 import { Button } from '@hilla/react-components/Button.js';
 import { ConfirmDialog } from '@hilla/react-components/ConfirmDialog';
 import { FormLayout } from '@hilla/react-components/FormLayout';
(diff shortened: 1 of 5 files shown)


packages/ts/react-crud/src/autoform.tsx
```diff
@@ -185,7 +186,7 @@ export function AutoForm<M extends AbstractModel>({
       const newItem = await form.submit();
       if (newItem === undefined) {
         // If update returns an empty object, then no update was performed
-        throw new Error('generic error');
+        throw new EndpointError('No update performed');
       } else if (afterSubmit) {
         afterSubmit({ item: newItem });
```
sissbruecker, line 191: This can still result in an error, and would now cause the `Something went wrong, please check all your values` text to show up in the form. That doesn't seem right. I think we could:
- Either wrap `afterSubmit` into a try/catch and swallow the error
- Or do error handling only for `EndpointError`s, and rethrow the error otherwise. Something like:
```ts
if (error instanceof ValidationError) {
  // Handled automatically
  return;
}
if (error instanceof EndpointError) {
  if (onSubmitError) {
    onSubmitError({ error });
  } else {
    setFormError('Something went wrong, please check all your values');
  }
}
// Rethrow
throw error;
```

The second option seems preferrable IMO. It doesn't swallow errors from application code in `afterSubmit`, and all of the error handling in the component becomes explicit about endpoint errors.

cromoteca: I agree and committed it. I just changed it so that the error message is used instead of that default.

sissbruecker (approved):

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -219,12 +220,14 @@ export function AutoForm<M extends AbstractModel>({
...
-      } else {
-        setFormError(genericError);
+      if (error instanceof EndpointError) {
+        if (onSubmitError) {
+          onSubmitError({ error });
+        } else {
+          setFormError(error.message);
+        }
       }
+      throw error;
```
taefi, line 230: Having this `throw` statement that re-throws anyway, makes me think of:
- What should the user be doing to catch it? Suppose that the error is type of `EndpointError` which is feed to the `onSubmitError` callback if exists. If the `onSubmitError` exists, the user still gets an uncaught error in the browser.
- What happens in the AutoCrud? 

Or maybe I'm missing a point?

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -245,9 +248,14 @@ export function AutoForm<M extends AbstractModel>({
...
-      if (onDeleteError) {
-        onDeleteError({ error });
+      if (error instanceof EndpointError) {
+        if (onDeleteError) {
+          onDeleteError({ error });
+        } else {
+          setFormError(error.message);
+        }
       }
+      throw error;
```
taefi, line 258: Same comment as for `onSubmitError`.

taefi (changes requested):

PR #1718, opened 2023-11-13
refactor: clarify auto form renderer props

sissbruecker:
Following feedback from the DX tests:
- Fixes the field renderer docs to remove the label prop that was removed
- Expands the form layout renderer docs and adds examples

### DIFF
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -68,8 +68,30 @@ export type AutoFormProps<M extends AbstractModel = AbstractModel> = ComponentSt
      */
     disabled?: boolean;
     /**
-     * Allows to customize the layout of the form by providing a custom renderer.
-     * Check the component documentation for details.
+     * Allows to customize the layout of the form by providing a custom
+     * renderer. The renderer receives the form instance and the pre-rendered
+     * fields as props. The renderer can either reuse the pre-rendered fields in
+     * the custom layout, or render custom fields and connect them to the form
+     * manually.
+     *
+     * Check the component documentation for details and examples.
+     *
+     * Example using pre-rendered fields:
+     * ```tsx
+     * <AutoForm layoutRenderer={({ children }) =>
+     *   <VerticalLayout>{children}</VerticalLayout>
+     * } />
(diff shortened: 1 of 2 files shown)


packages/ts/react-crud/src/autoform.tsx
```diff
@@ -68,8 +68,30 @@ export type AutoFormProps<M extends AbstractModel = AbstractModel> = ComponentSt
...
+     * renderer. The renderer receives the form instance and the pre-rendered
+     * fields as props. The renderer can either reuse the pre-rendered fields in
+     * the custom layout, or render custom fields and connect them to the form
+     * manually.
+     *
+     * Check the component documentation for details and examples.
+     *
+     * Example using pre-rendered fields:
+     * ```tsx
+     * <AutoForm layoutRenderer={({ children }) =>
```
cromoteca, line 81: `layoutRenderer={VerticalLayout} should work as well in this case

sissbruecker: Sure, but I was mostly trying to show how to make use of the prop in case you want to have a more complex layout.

cromoteca: What about two examples then? As we're trying to minimize code, I see value in suggesting the shortest one too.

sissbruecker: Well, I don't think that just rendering the fields in a vertical layout instead of a form layout is a practical use-case. So maybe just add another element somewhere in there?

sissbruecker: So how about this:
```tsx
<AutoForm layoutRenderer={({ children }) =>
  <VerticalLayout>
    {children}
    <p>All data is collected anonymously.</p>
  </VerticalLayout>
} />
```

cromoteca: > Well, I don't think that just rendering the fields in a vertical layout instead of a form layout is a practical use-case.
I still see some value in showing that you can use a layout directly. And I like your last example which makes explicit declaration necessary. I let you decide.

cromoteca (approved):

PR #1719, opened 2023-11-13
chore(2.3) update CRUD dependencies

cromoteca:
Some versions are still lagging behind in branch 2.3. A new patch release is needed to avoid the `Tried to define vaadin-lumo-styles version 24.2.0 when version 24.2.1 is already in use. Something will probably break.` error.


Lodin (approved):
Won't it trigger the bot to override the dependencies you've set?

PR #1739, opened 2023-11-15
feat: enable setting custom column order by visibleColumns

### LINKED ISSUE #1448: [autogrid] Make it possible to define where custom columns are added
Right now the custom columns are always added at the end. This is quite limiting for any other case than action buttons. It should be possible to define the order of all columns in the grid, including the custom ones

taefi:
## Description

If the user define a custom column and gives it a unique key, the same key can be used in `visibleColumns` props to set its order of appearance among the columns:

```tsx
<AutoGrid 
    service={...} 
    model={...} 
    visibleColumns={['fullName', 'gender', 'email', 'vip', 'birthDate', 'shiftStart']}
    customColumns={[
        <GridColumn key="fullName" autoWidth renderer={FullNameRenderer}></GridColumn>,
    ]}
/>,
```

In case the `visibleColumns` property is absent, the custom columns would be added at the end.

In the presence of `visibleColumns`, only those custom columns would appear in the grid that their key is listed in the `visibleColumns`. 

Fixes #1448

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


cromoteca (dismissed):
I like the idea. I'm not sure that custom columns without a key are retained. They should still find a place at the end of the list.

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -197,7 +197,23 @@ function useColumns(
...
-    columns = [...columns, ...options.customColumns];
+    const columnMap = new Map<string, JSX.Element>();
+    columns.forEach((column) => {
+      const { key } = column;
+      if (key) {
+        columnMap.set(key, column);
+      }
+    });
+    options.customColumns.forEach((customColumn) => {
+      const { key } = customColumn;
```
sissbruecker, line 208: I wonder if we should use `path` instead of `key`. Semantically, `key` is a React thing that is only needed for rendering an array of components. Ideally developers wouldn't have to specify a `key` at all, and it would be set by us. `path` on the other hand is actually about specifying what should be rendered in the column, so there is a semantic relation at least. The `path` values also align with how we specify properties in `visibleColumns`.

cromoteca: I agree with this, I also actually had this concern, but I've seen that `key` is already used in existing columns, that's probably why Soroosh chose it. In case we want to use `path`, maybe we should do that everywhere.

taefi:
> I like the idea. I'm not sure that custom columns without a key are retained. They should still find a place at the end of the list.

With 2.4.0.beta2, custom columns without a key are rendered at the end, however, react is complaining about the elements without a key. Unless, we come up with some automatic wrapping of each `GridColumn` provided by the user in an imaginary component such as `AutoGridColumn` that provides the key, the user needs to provide the key. This is one of the reasons I based my solution on the keys to minimize the effort for now.

packages/ts/react-crud/src/autogrid.tsx
sissbruecker: It sounds like we should narrow the type then. It should not be possible to render `div`s into the grid.

sissbruecker: Well, assuming there is a type for that. Otherwise we could consider a runtime check.

sissbruecker: Still, I don't quite understand the original concern. What's wrong with getting the path from the props? It should be accessible using something like `column.props.path`.

taefi: Well, there's a chance that I don't know how to do it, but when I watch `column.props` there is no `path` set for it, in other word, `column.props.path` undefined at runtime. The closest way to get the same value as path is to read the value of `column.props.value.propertyInfo.name`. 

My concern around both `path` and `column.props.value.propertyInfo.name` is that both can be undefined when creating a computed `GridColumn`. For example the following is working fine without setting any `path`:

```tsx
<AutoGrid
      customColumns={[
           <GridColumn renderer={FullNameRenderer}></GridColumn>,
      ]}
/>
```
What we need is to set "some property" required when defining a custom column so that the developer set it on the `GridColumn` and use it in the `visibleColumns`. Maybe path is a good candidate, but I chose `key` as it is holding the same value (`propertyInfo.name`) for the auto rendered columns, and React checks its uniqueness automatically, and we don't need to worry about accidental overlaps and confusions. But, in general I understand your concern about the `key`. I don't have enough insight about how annoying or abnormal is that to ask developers to set the `key` property for this purpose.

sissbruecker: Having to set a non-existing path for columns that don't relate to a property is indeed strange. Another argument for using `key` is column groups. In order to deal with the fact that header filters insert a top level column group, you could also pass a column group to custom columns, so the header title would align with the other ones. However column groups don't have the `path` property. For example:
```tsx
    <AutoGrid
      service={PersonListOnlyService}
      model={PersonModel}
      customColumns={[
        <GridColumnGroup key="custom" header="Custom">
          <GridColumn header="" renderer={CustomRenderer} />,
        </GridColumnGroup>
      ]}
    />
```

So I think I would be fine with using `key` for now.

packages/ts/react-crud/test/autogrid.spec.tsx
```diff
@@ -505,16 +505,52 @@ describe('@hilla/react-crud', () => {
...
+        const FullNameRenderer = ({ item }: { item: Person }): JSX.Element => (
+          <span>
+            {item.firstName} {item.lastName}
+          </span>
+        );
+        const FullNameHyphenRenderer = ({ item }: { item: Person }): JSX.Element => (
+          <span>
+            {item.firstName}-{item.lastName}
+          </span>
+        );
```
sissbruecker, line 546: It seems these could be extracted into the upper `describe` block and then reused between the examples.

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -190,15 +190,41 @@ function useColumns(
...
+      options.customColumns.forEach((customColumn) => {
+        const { key } = customColumn;
+        if (key) {
+          columnMap.set(key, customColumn);
+        }
+      });
+      columns = Array(columns.length + options.customColumns.length).fill(null);
+      effectiveColumns.forEach((key, index) => {
+        columns[index] = columnMap.get(key)!;
+      });
```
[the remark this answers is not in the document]
taefi: The reason behind choosing the Map was for not scanning either of those two arrays (columns and customColumns) for each effective column. Although, probably the JS Map structure is not be similar to a HashMap in java (and I think it scans and underlying array for the `get` operation), it might not be that big of an issue to choose either of the above suggested codes. I went for the Map version just with a hope that JS Map might be a better choice here semantically.

sissbruecker (approved):

PR #1746, opened 2023-11-16
feat(autoform): allow customizing error text from error handler

### LINKED ISSUE #1730: [autoform] Allow customizing error text from error handler
When saving an item results in an `EndpointException`, the form does show the exception's message in a paragraph below the form fields:

Bildschirmfoto 2023-11-15 um 11 04 41

Alternatively, developers can provide a submit error handler that will be called when an exception occurs, however in that case the paragraph below the form fields is not shown.

When using a custom error handler, instead of showing feedback to the user in some other location, it would be convenient if you could show a custom text in the same paragraph at the bottom of the form. For example, the component could pass a function to the error handler that allows setting the error message. Reusing the existing error element might also make it easier to make a custom error message more accessible, assuming the default error element is properly configured.

cromoteca:
Closes #1730


packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -361,6 +361,30 @@ describe('@hilla/react-crud', () => {
       expect(errorSpy).to.have.been.calledWith(sinon.match.hasNested('error.message', 'foobar'));
     });
 
+    it('allows to show a custom error message if the endpoint call fails', async () => {
+      const service: CrudService<Person> & HasTestInfo = createService<Person>(personData);
+      // eslint-disable-next-line @typescript-eslint/require-await
+      service.save = async (_item: Person): Promise<Person | undefined> => {
+        throw new EndpointError('foobar');
+      };
```
sissbruecker, line 369:
```suggestion
      sinon.stub(service, 'save').rejects(new EndpointError('foobar'));
```

packages/ts/react-crud/test/autoform.spec.tsx
sissbruecker, line 1: Let's also add a test for `onDeleteError`.

sissbruecker (approved):

PR #1751, opened 2023-11-20
feat: clear form after submitting new item

### LINKED ISSUE #1669: [autoform] Allow clearing the form after creating a new item
In order to create a new item, you can pass `null` as the edited item, or leave the `item` prop undefined. However after submitting the auto form, the fields stay filled with the previous values, and you can not reset the form to an empty state because:
- You don't have a different value to pass to the `item` prop in order for the auto form to detect that it should reset
- The is no imperative API for clearing the form

There should be some way to clear the form. Alternatively the form should reset automatically after submitting if the `item` prop does not contain an existing item.

A workaround could be to pass a new empty object each time you want to reset the form, though you'd have to deal with required properties in Typescript. Also not sure if there would be a problem with missing properties in Binder.

sissbruecker:
Automatically clears the auto form after submitting a new item. This seems a reasonable default behavior and developers can easily keep editing the newly created item by passing it to the `item` prop in a subsequent render. This is also consistent with the contract of the `item` prop - as long as you don't pass an existing item the form should keep creating new items.

As an alternative, we could pass a callback via the submit success event to let the developer clear the form manually, but I currently can't see a case where doing this automatically doesn't make sense or is inconvenient.

Fixes #1669


cromoteca:
Is there a test that confirms that the form is cleared after submission?

sissbruecker:
Yes, it's this one:
https://github.com/vaadin/hilla/blob/a4fa1553d5218fcf871ae59e497f020cabdcd674/packages/ts/react-crud/test/autoform.spec.tsx#L244-L265

Above that are tests that verify that the form is not cleared when editing an existing item.

cromoteca (approved):

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -220,6 +230,13 @@ export function AutoForm<M extends AbstractModel>({
       } else if (onSubmitSuccess) {
         onSubmitSuccess({ item: newItem });
       }
+      // Automatically clear the form after submitting a new item.
+      // Otherwise, there would be no way for the developer to clear it, as the
+      // there is no new value to set for the item prop to trigger the above
+      // effect in case the prop is already null, undefined or the empty item.
+      if (!item || item === emptyItem) {
+        form.clear();
+      }
```
taefi, line 239: I just wonder if there should be a switch to disable this?

taefi (changes requested):

packages/ts/react-crud/src/autoform.tsx
sissbruecker: What would be the use case? The only two I see is creating a new item, or editing an existing item. For the first one you want to clear the form, for the second one you assign the item prop.

taefi: If there is no pressing need for having a switch off for this behavior, you can ignore this and merge.

sissbruecker: @taefi I think you need to remove your request for changes, or approve, otherwise merge is not available.

taefi (approved):

PR #1757, opened 2023-11-21
feat: improve header filter sizing

### LINKED ISSUE #1668: [autogrid] Improve header filter sizing
Header filters are currently using the default Vaadin field widths. While the columns are all set to auto-size, effectively the column size is usually determined by the width of the filter field, as it is the largest element in the column. See the following example where the column has nearly double the size that it needs because the filter is so large:

Bildschirmfoto 2023-11-07 um 11 55 58

The header filters should have a more reasonable / smaller size, ideally based on the actual content, so that more columns can be displayed at once.

sissbruecker:
Reduces default header filter sizes so that more columns are visible at the same time. Also fixes the sizing of the filters so that they stretch if there is long form column content.

One challenge here is sizing the filters so that they would fully fit all possible filter values, which are effectively the contents rendered into the column. The filters have additional margins, paddings and controls that effectively reduce the amount of text you can fit in there. So even if a filter has the same width as a column cell, it can not fit the same amount of text. There also isn't a way to make the grid auto size mechanism compensate for that.

Instead, I changed the filters to use the small theme variant, which uses a smaller text size, allowing to fit more text. Then, based on manual testing, I defined some minimum widths for the filters that should allow to fit most values. That will not work perfectly in every case, but should be a good default and better than the status quo. Also noting that the grid auto size mechanism only considers visible rows, so there is always a chance that after some scrolling there will be a long text somewhere in there that will fit neither in the column nor the filter. Developers can still increase the column width, or modify the minimum widths using CSS, if they want to have a larger filter.

I also combined the operator select and filter input visually. That saves a few pixels, also looks cleaner IMO. Someone could probably do a better job of styling this, but it seems like a good start.

**Column sizing - Before**

Bildschirmfoto 2023-11-21 um 11 11 56

**Column sizing - After**

Bildschirmfoto 2023-11-21 um 11 11 14

**Header stretch - Before**

Bildschirmfoto 2023-11-21 um 11 14 01

**Header stretch - After**

Bildschirmfoto 2023-11-21 um 11 14 20

Fixes #1668


taefi (dismissed):

taefi (approved):

cromoteca (approved):
I'm not the best person to approve CSS changes, but I tested it locally and it is much better than before, so approving.

PR #1763, opened 2023-11-21
feat(autgrid): Add custom placeholder, debounce and minimum length to filters

### LINKED ISSUE #1492: [autogrid] add a way to customize the minimum number of typed characters in header filters before the filtering is triggered
### Describe your motivation

At the moment, the header filtering is triggered with every typed character in the text field for string typed column. While this might be OK in some projects, there should be a way to customize the threshold for the minimum number of typed characters in header filters before the filtering is triggered. 

On top of this, there should be a debounce to prevent the search being triggered while the user is still typing and not expecting a search result yet.

_No response_

_No response_

### Additional context

_No response_

krissvaa:
Add custom placeholder, debounce, and minimum length properties to column options for grid filters.

Fixes #1492

### DIFF
--- a/packages/ts/react-crud/src/autogrid-columns.tsx
+++ b/packages/ts/react-crud/src/autogrid-columns.tsx
@@ -16,10 +16,11 @@ import {
   NumberHeaderFilter,
   StringHeaderFilter,
   TimeHeaderFilter,
+  type HeaderFilterProps,
 } from './header-filter';
 import type { PropertyInfo } from './model-info';
 
-export type ColumnOptions = Omit<GridColumnProps<any>, 'dangerouslySetInnerHTML'>;
+export type ColumnOptions = HeaderFilterProps & Omit<GridColumnProps<any>, 'dangerouslySetInnerHTML'>;
 
 // eslint-disable-next-line consistent-return
 function getTypeColumnOptions(propertyInfo: PropertyInfo): ColumnOptions {
--- a/packages/ts/react-crud/src/header-filter.tsx
+++ b/packages/ts/react-crud/src/header-filter.tsx
@@ -16,6 +16,7 @@ import { convertToTitleCase } from './util';
 export type HeaderFilterProps = Readonly<{
   /**
    * Placeholder text for the filter input.
+   * Only applies to string, number and date/time value filters.
    */
   filterPlaceholder?: string;
   /**
--- a/packages/ts/react-crud/src/header-filter.tsx
+++ b/packages/ts/react-crud/src/header-filter.tsx
@@ -51,6 +51,12 @@ type ComparationSelectionProps = {
   onMatcherChanged(matcher: Matcher): void;
 };
 
+export type HeaderFilterProps = Readonly<{
+  filterPlaceholder?: string;
+  filterDebounceTime?: number;
+  filterMinLength?: number;
+}>;
+
 function ComparationSelection({ onMatcherChanged, value }: ComparationSelectionProps): ReactElement {
   const select = useRef<SelectElement>(null);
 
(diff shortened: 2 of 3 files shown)


packages/ts/react-crud/src/autogrid-columns.tsx
```diff
@@ -16,10 +16,11 @@ import {
   NumberHeaderFilter,
   StringHeaderFilter,
   TimeHeaderFilter,
+  type HeaderFilterProps,
 } from './header-filter';
 import type { PropertyInfo } from './property-info';
 
-export type ColumnOptions = Omit<GridColumnProps<any>, 'dangerouslySetInnerHTML'>;
+export type ColumnOptions = HeaderFilterProps & Omit<GridColumnProps<any>, 'dangerouslySetInnerHTML'>;
```
sissbruecker, line 23: It's inconvenient that you need to configure a debounce individually for every single column. It seems less likely that you want a debounce for individual columns instead of for all of them. Same for the minimum length maybe.

So maybe it would make more sense to start with a version that allows to configure this for the whole grid?

packages/ts/react-crud/src/header-filter.tsx
```diff
@@ -82,33 +88,67 @@ function ComparationSelection({ onMatcherChanged, value }: ComparationSelectionP
 }
 
 export function StringHeaderFilter(): ReactElement {
+  const context = useContext(ColumnContext)!;
+  const { filterPlaceholder, filterDebounceTime, filterMinLength } = context.customColumnOptions ?? {};
   const { updateFilter } = useFilterState(Matcher.CONTAINS);
+  const [inputValue, setInputValue] = useState('');
+
+  useEffect(() => {
+    if (filterMinLength && inputValue && inputValue.length < filterMinLength) {
```
sissbruecker, line 97: The behavior here could be confusing: It only starts filtering after entering the min. number of characters. However after removing a single character, it does not remove the filter and still shows the filter result. Maybe it should reset the filter immediately if the filter value has less than the minimum number of characters? Not quite sure which is better.

packages/ts/react-crud/src/header-filter.tsx
```diff
@@ -51,6 +51,12 @@ type ComparationSelectionProps = {
   onMatcherChanged(matcher: Matcher): void;
 };
 
+export type HeaderFilterProps = Readonly<{
```
sissbruecker, line 54: Whenever we add something to the public API we should add JS doc.

packages/ts/react-crud/src/autogrid-columns.tsx
krissvaa: I see three levels we could allow users to configure: 
1. whole grid -  one property would apply to all of the filters
2. type specific - apply debounce only for string filters or change placeholder only for number type filters
3. column specific - current solution, configure each column separately. 

For now, I implemented 3rd option as it was possible without any significant changes.
Grid level is also possible without any significant changes. 
The type level could require some bigger changes.

I don't see a debounce option for types other than number and string; maybe I am wrong.
This also could be a good start, or I can add props to both grid and column levels. 

I was also deciding if we want to have to debounce by default but I decided against it as it makes default filtering feel a bit more sluggish. 

As for minimum length - this is a quite specific requirement - usually for performance reasons. 
I would leave it as column-specific.

krissvaa: After internal talk, I will leave column-specific filter properties. We can return to this in the future.

Also will implement the default value for debounce (200ms) as in webcomponents: 
https://github.com/vaadin/web-components/blob/6f4bc1e642db3cc2781146af3a7e131c682e8c0b/packages/grid/src/vaadin-grid-filter-element-mixin.js#L98C101-L98C101

packages/ts/react-crud/src/header-filter.tsx
krissvaa: Agreed, and made changes that the filter is cleared with empty input value if input value length is less than min length

Issue #1432, opened 2023-10-03
[autoform] Disable submit button when editing and there are no changes


Artur-:
When editing an item (not when creating a new item) the submit button should be disabled unless the user has made some changes to form values. When making changes to a value, the submit button becomes enabled. Note that the enabled state must change on input and not on field blur so you can type into a field and click "save" without wondering why the button is disabled

PR #1764, opened 2023-11-21
feat(autoform): submit form on Enter key press

### LINKED ISSUE #1728: [autoform] Add keyboard shortcut for submitting the form
A common way to submit a form is to press <kbd>Enter</kbd> or <kbd>Control/Command</kbd> + <kbd>Enter</kbd> in any field of the form. This makes the form more effective for power users, and is also how native HTML forms behave.

We should consider adding such a shortcut as default, or via an option, either as opt-in or opt-out.

krissvaa:
Fixes #1728

### DIFF
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -346,7 +346,7 @@ export function AutoForm<M extends AbstractModel>({
   }
 
   const handleKeyDown = (event: KeyboardEvent): void => {
-    if (event.key === 'Enter') {
+    if (event.key === 'Enter' && (disabled === false || (isEditMode && form.dirty))) {
       // eslint-disable-next-line no-void
       void handleSubmit();
     }
@@ -387,7 +387,6 @@ export function AutoForm<M extends AbstractModel>({
       </VerticalLayout>
       <div className="auto-form-toolbar">
         <Button
-          id="form-submit"
           theme="primary"
           disabled={!!disabled || (isEditMode && !form.dirty)}
           // eslint-disable-next-line @typescript-eslint/no-misused-promises
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -5,7 +5,7 @@ import { ConfirmDialog } from '@hilla/react-components/ConfirmDialog';
 import { FormLayout } from '@hilla/react-components/FormLayout';
 import { VerticalLayout } from '@hilla/react-components/VerticalLayout.js';
 import { useForm, type UseFormResult } from '@hilla/react-form';
-import { type ComponentType, type JSX, type ReactElement, useEffect, useMemo, useState } from 'react';
+import { type ComponentType, type JSX,  type KeyboardEvent, type ReactElement, useEffect, useMemo, useState } from 'react';
 import { AutoFormField, type AutoFormFieldProps, type FieldOptions } from './autoform-field.js';
 import css from './autoform.obj.css';
 import type { CrudService } from './crud.js';
@@ -345,6 +345,13 @@ export function AutoForm<M extends AbstractModel>({
     setShowDeleteDialog(false);
   }
 
+  const handleKeyDown = (event: KeyboardEvent): void => {
+    if (event.key === 'Enter') {
+      // eslint-disable-next-line no-void
+      void handleSubmit();
+    }
+  };
+
   function createAutoFormField(propertyInfo: PropertyInfo): JSX.Element {
     const fieldOptionsForProperty = fieldOptions?.[propertyInfo.name];
     const colspanValue = fieldOptionsForProperty?.colspan;
@@ -374,12 +381,13 @@ export function AutoForm<M extends AbstractModel>({
 
   return (
     <div className={`auto-form ${className ?? ''}`} id={id} style={style} data-testid="auto-form">
-      <VerticalLayout className="auto-form-fields">
+      <VerticalLayout className="auto-form-fields" onKeyDown={handleKeyDown}>
         {layout}
         {formError ? <div style={{ color: 'var(--lumo-error-color)' }}>{formError}</div> : <></>}
       </VerticalLayout>
       <div className="auto-form-toolbar">
         <Button
+          id="form-submit"
           theme="primary"
           disabled={!!disabled || (isEditMode && !form.dirty)}
           // eslint-disable-next-line @typescript-eslint/no-misused-promises
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -5,7 +5,15 @@ import { ConfirmDialog } from '@hilla/react-components/ConfirmDialog';
 import { FormLayout } from '@hilla/react-components/FormLayout';
 import { VerticalLayout } from '@hilla/react-components/VerticalLayout.js';
 import { useForm, type UseFormResult } from '@hilla/react-form';
-import { type ComponentType, type JSX,  type KeyboardEvent, type ReactElement, useEffect, useMemo, useState } from 'react';
+import {
+  type ComponentType,
+  type JSX,
+  type KeyboardEvent,
+  type ReactElement,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 import { AutoFormField, type AutoFormFieldProps, type FieldOptions } from './autoform-field.js';
 import css from './autoform.obj.css';
 import type { CrudService } from './crud.js';
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -268,6 +268,7 @@ export function AutoForm<M extends AbstractModel>({
 
   const isEditMode = item !== undefined && item !== null && item !== emptyItem;
   const showDeleteButton = deleteButtonVisible && isEditMode && modelInfo.idProperty;
+  const isSubmitDisabled = !!disabled || (isEditMode && !form.dirty);
 
   useEffect(() => {
     if (item !== emptyItem) {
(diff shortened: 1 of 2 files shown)


krissvaa:
I am just wondering if the key-down should be on wrapping VerticalLayout or the 1 level down -  on FormLayout.

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -325,12 +332,13 @@ export function AutoForm<M extends AbstractModel>({
...
   return (
     <div className={`auto-form ${className ?? ''}`} id={id} style={style} data-testid="auto-form">
-      <VerticalLayout className="auto-form-fields">
+      <VerticalLayout className="auto-form-fields" onKeyDown={handleKeyDown}>
         {layout}
         {formError ? <div style={{ color: 'var(--lumo-error-color)' }}>{formError}</div> : <></>}
       </VerticalLayout>
       <div className="auto-form-toolbar">
         <Button
+          id="form-submit"
```
sissbruecker, line 341: In general we should avoid adding any IDs to our components, as it can result in duplicate IDs if you use multiple of them. Does this have a specific purpose?

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -344,7 +352,7 @@ export function AutoForm<M extends AbstractModel>({
           </Button>
         ) : null}
         {deleteButtonVisible && isEditMode && (
-          <Button className="auto-form-delete-button" theme="tertiary error" onClick={deleteItem}>
+          <Button className="auto-form-delete-button" theme="tertiary error" onClick={deleteItem} typeof="submit">
```
sissbruecker, line 355: What does `typeof` do here? I can't find any reference to such an attribute existing. Also why is the delete button type of "submit"?

packages/ts/react-crud/src/autoform.tsx
sissbruecker, line 1: Let's add tests for this new functionality. For example:
- Pressing enter in a field submits the form
- Pressing enter in a field submits the form when using a custom layout
- Pressing enter on the discard button does not submit
- Pressing enter on the delete button does not submit

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -482,6 +484,84 @@ describe('@hilla/react-crud', () => {
       expect(submitButton.disabled).to.be.true;
     });
 
+    describe('form submit', () => {
```
sissbruecker, line 487:
```suggestion
    describe('keyboard shortcuts', () => {
```

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -482,6 +484,84 @@ describe('@hilla/react-crud', () => {
...
+        const service = personService();
+        const form = await populatePersonForm(1, undefined, undefined, undefined, service);
+
+        const submitButton = await form.findButton('Submit');
+        const submitSpy = sinon.spy(service, 'save');
+
+        await form.typeInField('First name', 'J');
+        expect(submitButton.disabled).to.be.false;
+
+        fireEvent.keyDown(form.formLayout, { key: 'Enter', code: 'Enter', charCode: 13 });
```
sissbruecker, line 498: Would suggest to fire from a field, as that is what would usually have focus.

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -482,6 +484,84 @@ describe('@hilla/react-crud', () => {
...
+
+        const submitSpy = sinon.spy(service, 'save');
+
+        const submitButton = await form.findButton('Submit');
+        await form.typeInField('First name', 'J');
+        expect(submitButton.disabled).to.be.false;
+
+        const element = form.instance.getElementsByTagName('vaadin-vertical-layout')[0];
+
+        fireEvent.keyDown(element, { key: 'Enter', code: 'Enter', charCode: 13 });
```
sissbruecker, line 558: Would likewise suggest to press enter from a field.

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -289,6 +289,13 @@ export function AutoForm<M extends AbstractModel>({
     setShowDeleteDialog(false);
   }
 
+  const handleKeyDown = (event: KeyboardEvent): void => {
+    if (event.key === 'Enter' && (disabled === false || (isEditMode && form.dirty))) {
```
sissbruecker, line 293: `isEditMode` reflects whether an existing item is being edited or not (unfortunate name maybe). So currently it's not possible to submit by enter when creating a new item. I don't think there is any reason to disable this shortcut other than when the form is disabled, or when it's not dirty.

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -482,6 +484,84 @@ describe('@hilla/react-crud', () => {
       expect(submitButton.disabled).to.be.true;
     });
 
+    describe('form submit', () => {
+      it('submit when enter key is pressed', async () => {
```
sissbruecker, line 488:
```suggestion
      it('submits when enter key is pressed', async () => {
```

Nit: Prefer to keep test cases names consistent with the `it` prefix. Ideally it reads like: "It submits then the enter key is pressed", or "It should submit when the enter key is pressed".

packages/ts/react-crud/src/autoform.tsx
taefi: `isEditMode` is needed for disabling the Submit button according to the criteria mentioned in https://github.com/vaadin/hilla/issues/1432. 

If we want to keep the `disabled` status of the Submit button consistent with the availability of the form submission via keyboard shortcut, the condition should be something like:
```tsx
event.key === 'Enter' && (disabled === false && (!isEditMode || form.dirty))
```
So probably it makes more sense to define variable for it and use it in both places.

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -524,6 +525,84 @@ describe('@hilla/react-crud', () => {
...
+        const submitSpy = sinon.spy(service, 'save');
+
+        await form.typeInField('First name', 'J');
+        expect(submitButton.disabled).to.be.false;
+
+        const firstNameField = await within(form.instance).findByLabelText('First name');
+        fireEvent.keyDown(firstNameField, { key: 'Enter', code: 'Enter', charCode: 13 });
+        await nextFrame();
+
+        expect(submitSpy).to.have.been.calledOnce;
```
sissbruecker, line 543:
```suggestion
        const service = personService();
        const form = await populatePersonForm(1, undefined, undefined, undefined, service);
        const submitSpy = sinon.spy(service, 'save');

        await form.typeInField('First name', 'J{enter}');
        expect(submitSpy).to.have.been.calledOnce;
```

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -524,6 +525,84 @@ describe('@hilla/react-crud', () => {
...
+
+        const submitSpy = sinon.spy(service, 'save');
+        const submitButton = await form.findButton('Submit');
+        expect(submitButton.disabled).to.be.true;
+
+        const firstNameField = await within(form.instance).findByLabelText('First name');
+        fireEvent.keyDown(firstNameField, { key: 'Enter', code: 'Enter', charCode: 13 });
+        await nextFrame();
+
+        expect(submitSpy).to.have.not.been.called;
```
sissbruecker, line 558:
```suggestion
        const service = personService();
        const form = await populatePersonForm(1, undefined, undefined, undefined, service);
        const submitSpy = sinon.spy(service, 'save');

        await form.typeInField('First name', '{enter}');
        expect(submitSpy).to.have.not.been.called;
```

packages/ts/react-crud/test/autoform.spec.tsx
```diff
@@ -524,6 +525,84 @@ describe('@hilla/react-crud', () => {
...
+
+        const submitButton = await form.findButton('Submit');
+        await form.typeInField('First name', 'J');
+        expect(submitButton.disabled).to.be.false;
+
+        const firstNameField = await within(form.instance).findByLabelText('First name');
+        fireEvent.keyDown(firstNameField, { key: 'Enter', code: 'Enter', charCode: 13 });
+        await nextFrame();
+
+        expect(submitSpy).to.have.been.calledOnce;
```
sissbruecker, line 602:
```suggestion
        const service = personService();
        function MyLayoutRenderer({ children }: AutoFormLayoutRendererProps<PersonModel>) {
          return <VerticalLayout>{children}</VerticalLayout>;
        }
        const form = await populatePersonForm(1, { layoutRenderer: MyLayoutRenderer }, undefined, undefined, service);
        const submitSpy = sinon.spy(service, 'save');

        await form.typeInField('First name', 'J{enter}');
        expect(submitSpy).to.have.been.calledOnce;
```

sissbruecker (commented):
Some suggestions for simplifying tests

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -345,6 +354,13 @@ export function AutoForm<M extends AbstractModel>({
     setShowDeleteDialog(false);
   }
 
+  const handleKeyDown = (event: KeyboardEvent): void => {
+    if (event.key === 'Enter' && !isSubmitDisabled) {
+      // eslint-disable-next-line no-void
+      void handleSubmit();
```
taefi, line 360: Question: if we change the return type of `handleSubmit` from `Promise<void>` to `Promise<never>`, would it enable us to call it here just as `handleSubmit();` (without the `void` and the hint for the eslint)?

taefi: I guess not, since we have the following in the catch block:
```tsx
if (error instanceof ValidationError) {
    // Handled automatically
    return;
}
```

taefi (approved):

Issue #1760, opened 2023-11-21
Tests can be unintentionally skipped


Lodin:
According to #1751 and #1759, the repository configuration lacks a configuration that forbids unintentional test skipping in PRs.

To address it, I would suggest addressing it in four different ways:

[] Set up Codecov as a condition to merge the PR.
[] Make Codecov more friendly to the developer (reduce the amount of alerts it shows so the coverage amount is simpler to address)
[] Find and add an Eslint rule to forbid committing the `it.only` and similar instructions.
[] Investigate if Mocha can skip `it.only` and similar instructions during CI

PR #1765, opened 2023-11-21
chore: forbid unintentional exclusive tests

Lodin:
Addresses two points of #1760.

- Add an Eslint rule to forbid exclusive tests.
- Add a Mocha configuration that fail tests during CI if exclusive test is found.


.eslintrc
```diff
@@ -16,6 +16,7 @@
     "import/prefer-default-export": "off",
     "max-params": "off",
     "max-classes-per-file": "off",
+    "mocha/no-exclusive-tests": "error",
```
sissbruecker, line 19: I can still commit files that have been modified to use `it.only`. I think our lint-staged config is broken somehow. Though it seems to notice the issue in CI, which is good enough for me.

.eslintrc
sissbruecker, line 1: This uncovered another usage of `it.only` in `packages/ts/generator-typescript-plugin-backbone/test/hierarchy-ref/HierarchyRef.spec.ts` 🙈. Could you please fix it?

.eslintrc
Lodin: Fixed that as well

sissbruecker (approved):
lint-staged works now 👍

PR #1769, opened 2023-11-22
feat: provide refresh to be called from outside

taefi:
This adds the possibility of refreshing the 
AutoGrid from outside by calling the `refresh`
function, instead of passing down a
refreshTrigger state from the parent component.


packages/ts/react-crud/test/autogrid.spec.tsx
```diff
@@ -873,25 +875,42 @@ describe('@hilla/react-crud', () => {
...
         await nextFrame();
         expect(listSpy).to.have.been.calledOnce;
 
-        // Does not refresh if refreshTrigger is not changed
-        result.rerender(<AutoGrid service={service} model={PersonModel} refreshTrigger={0} />);
+        // Does not refresh if refresh is not called
+        result.rerender(<AutoGridRefreshTestWrapper service={service} />);
         await nextFrame();
         await nextFrame();
         expect(listSpy).to.have.been.calledOnce;
```
sissbruecker, line 907: I think we can remove these.

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -317,16 +323,36 @@ export function AutoGrid<TItem>({
...
 
   return (
-    <Grid itemIdPath={modelInfo.idProperty?.name} {...gridProps} ref={ref}>
+    <Grid itemIdPath={modelInfo.idProperty?.name} {...gridProps} ref={gridRef}>
       {children}
     </Grid>
   );
 }
+
+type AutoGridSignature = <TItem>(
```
sissbruecker, line 340: We could name the type `AutoGrid`, being a type only it would not conflict with the constant below. While potentially confusing for us, it could be beneficial for users if the type they are dealing with is called `AutoGrid` rather than `AutoGridSignature`.

packages/ts/react-crud/test/autogrid.spec.tsx
sissbruecker, line 907: I think we can remove these.

taefi: Probably, that's not necessary to re-test rendering and re-rendering effect on the call to the service here.

taefi: Done.

packages/ts/react-crud/src/autogrid.tsx
taefi: Done.

sissbruecker (approved):

PR #1779, opened 2023-11-22
feat: allow configuring columns and fields for any nested property

sissbruecker:
Currently `AutoGrid` and `AutoForm` only allow using nested properties from classes that are annotated with `@OneToOne`. That is because the detection of available properties is currently intermixed with logic which determines which columns / fields to show by default.

This change separates that logic by introducing a `ModelInfo` class which can give information about properties of models in general, as well as a `getDefaultProperties` which returns the columns / fields to show by default. That allows to specify any (nested) property in `visibleColumns` or `visibleFields`.


taefi (approved):
LGTM.

platosha (approved):

PR #1791, opened 2023-11-24
chore: remake the IT installation system to avoid symlink issues

Lodin:


Lodin:
Closing because this change is no longer necessary

Issue #1799, opened 2023-11-27
[autocrud] Disable delete if no ID property is found


sissbruecker:
Deleting an item through a `CrudService` requires `AutoForm` / `AutoCrud` to detect an ID property. If such a property can not be found then deleting just fails.

The delete button should not be shown if no ID property is detected.

Issue #1798, opened 2023-11-27
[autocrud] Enhance detection of ID property and allow configuring it


sissbruecker:
Currently `AutoGrid`, `AutoForm` and `AutoCrud` detect the ID property by looking for a property that is annotated with `@Id`. The ID property is currently used to:
- Delete an item through a `CrudService`
- Keep a grid item selected after a reload

When using items without such a property, for example with DTOs, this functionality just stops working.

For those cases it should be possible to configure a custom ID property. Furthermore the ID detection logic could be improved, for example by using a property named `id` as the ID property.

PR #1807, opened 2023-11-27
feat: customize item ID property

sissbruecker:
Allows to customize the item ID property when not using JPA entities. Also enhances the ID property detection to use a property with the name `id` as fallback if no JPA annotation is found. Lastly this also hides the delete button if no ID property is detected, as delete doesn't work in that case.

Fixes https://github.com/vaadin/hilla/issues/1798
Fixes https://github.com/vaadin/hilla/issues/1799


taefi (approved):

taefi:
@sissbruecker I disabled the Auto Merge as it is blocked by the sonar error anyway. As resolving the sonar lint error regarding the duplicity does not make much sense, you can ignore and merge IMO, once other checks are passed.

PR #1825, opened 2023-11-29
feat(autogrid): add options to disable column filtering and sorting

cromoteca:


packages/ts/react-crud/src/autogrid-columns.tsx
```diff
@@ -19,7 +19,18 @@ import {
 } from './header-filter';
 import type { PropertyInfo } from './model-info';
 
-export type ColumnOptions = Omit<GridColumnProps<any>, 'dangerouslySetInnerHTML'>;
+export type ColumnOptions = Omit<GridColumnProps<any>, 'dangerouslySetInnerHTML'> & {
+  /**
+   * If true, the column can be sorted. This is useful to disable sorting on
```
sissbruecker, line 24:
```suggestion
   * If true, the column can be sorted. This is useful to disable sorting for
   * properties that are not sortable in the backend, or that require excessive processing.
```
Would be good to mention that this really depends on the specific property displayed by the column, and maybe also hint about that the sorting capability depends on the backend.

Same for the filter docs below.

sissbruecker (approved):

PR #1831, opened 2023-11-29
feat(autoform): Add props to provide custom client-side validators

### LINKED ISSUE #1797: [autoform] Allow configuring client-side validators
When rendering an `AutoForm`, there is no straight-forward way to configure client-side validators. This would be useful for cases where developers either don't want to, or can not use Jakarta annotations on entities or DTOs.

There should be a way to configure client-side validators for individual fields as well as the whole form.

As a workaround, it might be possible to define a custom layout renderer, which gets access to the form instance.

krissvaa:
Fixes #1797

Add functionality that users can set custom client-side validators to form by passing it to `fieldOptions`

### DIFF
--- a/packages/ts/react-crud/src/autoform-field.tsx
+++ b/packages/ts/react-crud/src/autoform-field.tsx
@@ -1,4 +1,4 @@
-import { _enum, type EnumModel } from '@hilla/form';
+import { _enum, type EnumModel, type Validator } from '@hilla/form';
 import { Checkbox, type CheckboxProps } from '@hilla/react-components/Checkbox.js';
 import { DatePicker, type DatePickerProps } from '@hilla/react-components/DatePicker.js';
 import { DateTimePicker, type DateTimePickerProps } from '@hilla/react-components/DateTimePicker.js';
@@ -8,7 +8,9 @@ import { Select, type SelectProps } from '@hilla/react-components/Select.js';
 import { TextField, type TextFieldProps } from '@hilla/react-components/TextField.js';
 import { TimePicker, type TimePickerProps } from '@hilla/react-components/TimePicker.js';
 import type { FieldDirectiveResult, UseFormResult } from '@hilla/react-form';
+import { useFormPart } from '@hilla/react-form';
 import type { JSX } from 'react';
+import { useEffect } from 'react';
 import { useDatePickerI18n, useDateTimePickerI18n } from './locale.js';
 import type { PropertyInfo } from './model-info.js';
 import { convertToTitleCase } from './util.js';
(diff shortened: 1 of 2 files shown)


packages/ts/react-crud/src/autoform-field.tsx
```diff
@@ -133,6 +140,16 @@ export type AutoFormFieldProps = CheckboxProps &
...
   const label = options?.label ?? propertyInfo.humanReadableName;
+
+  const formPart = useFormPart(getPropertyModel(form, propertyInfo));
+  useEffect(() => {
+    if (options?.validators) {
+      options.validators.forEach((validator) => {
+        formPart.addValidator(validator);
+      });
+    }
+  }, [formPart, options]);
```
sissbruecker, line 151: This will add new validators with every single render. The only way to make this work is if developers would memoize (`useMemo`) the validators before passing them into the options, which no one will think of doing. I have a hard time seeing how the props approach could work here.

sissbruecker (changes requested):

packages/ts/react-crud/src/autoform-field.tsx
sissbruecker: An alternative might be to expose the `form` instance (from `useForm`) through a ref. Though I'm not sure if you can add validators to individual fields through that.

packages/ts/react-crud/src/autoform-field.tsx
```diff
@@ -132,8 +139,16 @@ export type AutoFormFieldProps = CheckboxProps &
...
-  const label = options?.label ?? propertyInfo.humanReadableName;
-  if (options?.renderer) {
+  const label = options.label ?? propertyInfo.humanReadableName;
+
+  const formPart = useFormPart(getPropertyModel(form, propertyInfo));
+  const defaultValidators = useMemo(() => formPart.validators, []);
+  const { validators } = options;
+  useEffect(() => {
+    formPart.setValidators([...defaultValidators, ...(validators ?? [])]);
+  }, [validators]);
```
krissvaa, line 151: It seems that this code doesn't have any side effects. 
For now, I think there are no other simple solutions for users to add custom validators. 

An already-ready option is for users to use a custom layout where the form object is provided to set any validators they want. 
Other options are to provide some callbacks that interact with formPart/binder directly for the model property.

krissvaa: @sissbruecker  was concerned that setValidators would cause a change event in Hilla Binder instance, which in turn updates some state in useForm.

platosha: I have tried it, and it seems to not have any circular recursion.

platosha (dismissed):

platosha (approved):

PR #1870, opened 2023-12-08
feat: add a helper to remap properties in filters and pageables

### LINKED ISSUE #1865: [autocrud] allow developers to easily adapt `Pageable` and `Filter` when DTO field names don't correspond to entity
### Describe your motivation

When using DTOs, it happens that some fields have different names or paths than in the original entity. For example `name` can become `personName`, or `department.name` can become `departmentName`.

In that case, when requesting items using `ListService` or `CrudService`, property names match the DTO, while filtering and sorting happen at JPA level or according to the backend.

The developer then needs to modify those before using them.

A tool can be added to Hilla that, given some name mappings, can modify paging and filtering accordingly.

An additional hook can be provided to allow for any other kind of manual manipulation of each filter.

_No response_

### Additional context

_No response_

cromoteca:
This class deals with the simplest cases of properties that change name or path in a DTO. It also supports a function to be applied each time a `PropertyStringFilter` is encountered, to allow for more sophisticated changes.

Closes #1865


packages/java/endpoint/src/main/java/dev/hilla/crud/filter/Remapper.java
```diff
@@ -0,0 +1,115 @@
...
+     */
+    public Remapper withMapping(String from, String to) {
+        mappings.put(from, to);
+        return this;
+    }
+
+    /**
+     * Declares a function that will be applied to all
+     * {@link PropertyStringFilter} instances. This can be used to modify the
+     * filter value in a more complex way than a simple mapping.
```
taefi, line 38:
```suggestion
     * Declares a function that will be applied to all
     * {@link PropertyStringFilter} instances. This can be used to modify the
     * filter value in a more complex way than a simple mapping.
     * <p>
     * Note: The passed in {@code filterTransformation} function is applied
     * after all other mappings are applied.
```

taefi (changes requested):

packages/java/endpoint/src/main/java/dev/hilla/crud/filter/Remapper.java
taefi: Another comment that comes to mind is that the `filterTransformation` function being applied on top of the `dto -> entity` mappings may not be very intuitive. Providing support a custom way of transformation for corner cases may be a good idea, but, doing it on top of simple mappings makes it complicated IMO. Developer need to keep it in mind that though this is a filter transformation function, they may need to rely on property names from entity depend on what mappings applied beforehand. I'd suggest simplify it so that it only support simple property name/path mappings from DTO to ENTITY.

taefi: Maybe we can provide factory methods to instantiate a `Remapper` instance either with simple mappings or based on a passed in transformation function.

packages/java/endpoint/src/main/java/com/vaadin/hilla/crud/filter/FilterTransformer.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort;
+
+/**
+ * Remaps property names in filters and pageable objects.
```
taefi, line 12:
```suggestion
 * Utility class for transforming property names in filters and pageable objects.
```

packages/java/endpoint/src/main/java/com/vaadin/hilla/crud/filter/FilterTransformer.java
```diff
@@ -0,0 +1,118 @@
...
+ */
+public class FilterTransformer {
+
+    private final Map<String, String> mappings = new HashMap<>();
+    private Function<PropertyStringFilter, PropertyStringFilter> filterTransformation;
+
+    /**
+     * Declares a mapping from one property name to another. If a filter or
+     * pageable is remapped, all occurrences of the property name will be
```
taefi, line 21:
```suggestion
     * pageable is transformed, all occurrences of the property name will be
```

packages/java/endpoint/src/main/java/com/vaadin/hilla/crud/filter/FilterTransformer.java
```diff
@@ -0,0 +1,118 @@
...
+     * @return This instance.
+     */
+    public FilterTransformer withMapping(String from, String to) {
+        mappings.put(from, to);
+        return this;
+    }
+
+    /**
+     * Declares a function that will be applied to all
+     * {@link PropertyStringFilter} instances. This can be used to modify the
```
taefi, line 37: This description of `function that will be applied to all {@link PropertyStringFilter} instances.` is a bit vague to me. If it means that this function takes the filter instance as a parameter and can effect all the filter properties and and their values, could we clarify the javadoc to reflect this?

packages/java/endpoint/src/main/java/com/vaadin/hilla/crud/filter/FilterTransformer.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Remaps a filter, replacing all property names with their mapped values.
+     *
+     * @param filter
+     *            The filter to remap.
+     * @return The remapped filter.
+     */
+    public Filter remap(Filter filter) {
```
taefi, line 60:
```suggestion
    /**
     * Applies registered property mappings and transformation function on the provided filter instance.
     *
     * @param filter
     *            The filter instance to transform.
     * @return The transformed filter.
     */
    public Filter apply(Filter filter) {
```

packages/java/endpoint/src/main/java/com/vaadin/hilla/crud/filter/FilterTransformer.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Remaps a pageable, replacing all property names with their mapped values.
+     *
+     * @param pageable
+     *            The pageable to remap.
+     * @return The remapped pageable.
+     */
+    public Pageable remap(Pageable pageable) {
```
taefi, line 104:
```suggestion
    /**
     * Applies registered property mappings on the provided pageable instance.
     * <p>
     * Note: The passed in {@code filterTransformation} function is not applied on pageables.
     *
     * @param pageable
     *            The pageable instance to transform.
     * @return The transformed pageable.
     */
    public Pageable apply(Pageable pageable) {
```

taefi (changes requested):

taefi (approved):

Issue #1732, opened 2023-11-15
[autoform] Allow customizing more properties via `FieldOptions`


sissbruecker:
Currently you can only customize a limited number of field properties via the `fieldOptions` prop (`label` and `colSpan` at this point). To customize any other property you need to implement a custom renderer and set the property on the manually rendered component. This is inconvenient if you just want to quickly set a property, but then have to figure out how to implement a custom renderer.

We should look into supporting more properties that can be configured via field options. The challenges here seem to be:
- Not every property might be supported by every type of field component that we render automatically
- Typing this properly with Typescript so that you can ideally on set sensible / supported properties
  - Or alternatively mostly disable typing: allow declaring any property, but still try to provide auto-complete for known properties. Let developers deal with warnings / errors that might result from setting an unsupported prop.

PR #1876, opened 2023-12-11
feat: add more common props to field options

sissbruecker:
Adds more common field properties to `fieldOptions`:
- `id`
- `className`
- `style`
- `readonly`
- `disabled`
- `placeholder`
- `helperText`

`placeholder` and `helperText` are not supported by component types, for example checkbox doesn't have either, and a select with a preselected value can never show a placeholder. Still think they are useful to have by default.

As for more specific properties that only apply to some components I think we are at a dead-end with this approach - there isn't any way how we could properly type the respective field options object so that it accepts props for the respective component. 

One idea I had was to allow passing a prerendered JSX element to `renderer` which we would then enhance with the other options and the form binding. That would be easier than writing a custom renderer function and still gives you component-specific type information:
```tsx
<AutoForm 
  fieldOptions: {
    firstName: {
      renderer: <TextField clearButtonVisible />
    }
  }
/>
```

Part of #1732

### DIFF
--- a/packages/ts/react-crud/src/autoform-field.tsx
+++ b/packages/ts/react-crud/src/autoform-field.tsx
@@ -1,36 +1,74 @@
-import { _enum, type EnumModel, type Validator } from '@hilla/form';
-import { Checkbox, type CheckboxProps } from '@hilla/react-components/Checkbox.js';
-import { DatePicker, type DatePickerProps } from '@hilla/react-components/DatePicker.js';
-import { DateTimePicker, type DateTimePickerProps } from '@hilla/react-components/DateTimePicker.js';
-import { IntegerField, type IntegerFieldProps } from '@hilla/react-components/IntegerField.js';
-import { NumberField, type NumberFieldProps } from '@hilla/react-components/NumberField.js';
-import { Select, type SelectProps } from '@hilla/react-components/Select.js';
-import { TextArea, type TextAreaProps } from '@hilla/react-components/TextArea.js';
-import { TextField, type TextFieldProps } from '@hilla/react-components/TextField.js';
-import { TimePicker, type TimePickerProps } from '@hilla/react-components/TimePicker.js';
+import { _enum, type AbstractModel, type EnumModel, type Validator } from '@hilla/form';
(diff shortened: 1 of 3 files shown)


packages/ts/react-crud/src/autoform-field.tsx
```diff
@@ -62,127 +94,117 @@ export type FieldOptions = Readonly<{
...
   return pathParts.reduce<any>((model, property) => (model ? model[property] : undefined), form.model);
 }
 
-type AutoFormTextFieldProps = SharedFieldProps & TextFieldProps;
-
-function AutoFormTextField({ propertyInfo, form, options, label, ...other }: AutoFormTextFieldProps) {
-  const model = getPropertyModel(form, propertyInfo);
-  return <TextField {...other} {...form.field(model)} label={label} />;
+function AutoFormTextField({ field, fieldProps }: FieldRendererProps) {
+  return <TextField {...field} {...fieldProps} />;
```
sissbruecker, line 115: Felt a bit wrong to just assign the `AutoFormFieldProps` to the Vaadin component, which includes properties like `options`, which the Vaadin component doesn't support. Likewise any other prop added to `AutoFormFieldProps` would end up being assigned to the component as well, which could lead to a bug at some point.

So I refactored the fields a bit so that they receive individual props for the model of the property, the `field` directive result and then the common `fieldProps` which come from `fieldOptions`.

packages/ts/react-crud/src/autoform.tsx
```diff
@@ -363,8 +363,6 @@ export function AutoForm<M extends AbstractModel>({
 
   function createAutoFormField(propertyInfo: PropertyInfo): JSX.Element {
     const fieldOptionsForProperty = fieldOptions?.[propertyInfo.name] ?? {};
-    const colspanValue = fieldOptionsForProperty.colspan;
-    const colspan = colspanValue ? { colspan: colspanValue } : {};
```
sissbruecker, line 367: `colspan` is now assigned in the `AutoFormField` itself.

cromoteca (approved):

PR #1879, opened 2023-12-11
chore: update @vaadin/common-frontend

platosha:


Lodin (approved):

platosha:
Let's make sure that https://github.com/vaadin/flow/pull/18273 is merged at the same time.

PR #1887, opened 2023-12-12
chore: bump versions to 24.4

platosha:


packages/java/endpoint/src/main/java/dev/hilla/EndpointController.java
```diff
@@ -76,7 +76,7 @@
  */
 @RestController
 @Import({ EndpointControllerConfiguration.class, EndpointProperties.class })
-@NpmPackage(value = "@hilla/frontend", version = "2.5.0-beta1")
+@NpmPackage(value = "@hilla/frontend", version = "3.0.0-alpha0")
```
cromoteca, line 79: Version upgrades like this one seem to be problematic, looks like they should still point to the latest existing version. Is it OK if I revert those or there is a reason to use them as such?

cromoteca (dismissed):

taefi (approved):

taefi (approved):

PR #1891, opened 2023-12-13
fix: Avoid using decorators

Artur-:
field decorators are currently problematic when projects can use useDefineForClassFields true or false

### DIFF
--- a/packages/java/hilla-dev-mode/src/main/resources/META-INF/resources/frontend/dev-tools-database.ts
+++ b/packages/java/hilla-dev-mode/src/main/resources/META-INF/resources/frontend/dev-tools-database.ts
@@ -1,5 +1,5 @@
 import { LitElement, type TemplateResult, html } from 'lit';
-import { customElement, state } from 'lit/decorators.js';
+import { customElement } from 'lit/decorators.js';
 import { loginUsingUrl, waitForConsole } from './h2-util';
 import type {
   DevToolsInterface,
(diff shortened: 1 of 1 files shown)


packages/java/hilla-dev-mode/src/main/resources/META-INF/resources/frontend/dev-tools-database.ts
```diff
@@ -16,12 +16,12 @@ type H2Data = {
 };
 @customElement('devtools-database')
```
Lodin, line 17: I guess, we should get rid of decorators completely here

Lodin (changes requested):

packages/java/hilla-dev-mode/src/main/resources/META-INF/resources/frontend/dev-tools-database.ts
Artur-: Should we? Class level decorators have no issues as far as I know

Lodin: Yes but it looks strange if you use class decorators but not the property ones. And we could also avoid any possible issues with the decorators at all

Artur-: Ok, now there are no decorators

Lodin (approved):

PR #1896, opened 2023-12-14
refactor: separate data provider logic from auto grid

sissbruecker:
Separates the data provider implementation from `AutoGrid` specific logic:
- Moves the DP implementation into a separate module
- Refactored the data providers to be class based, with specific implementations for infinite and fixed-size data providers.
- Moves the `recalculateColumnWidths` workaround back to the `AutoGrid`
- Removes the item count holder and footer ref, and instead provides a callback that returns the total and filtered item counts
- Adds test coverage for the DP implementation itself

Also refactors the footer item count logic a bit:
- The footer ref has been removed. Instead the `AutoGrid` now has a state for the item counts, which is passed down into the footer column context to trigger a re-render when the item counts change.
- Changed the signature of the custom footer renderer to only receive the item counts, and not the options which counts should be displayed. The renderer can decide on its own whether to render a count or not.


taefi (approved):
LGTM

PR #1900, opened 2023-12-18
feat(react-crud): support additional columns / fields from options

### LINKED ISSUE #1743: [autoform] Adding custom fields by providing the renderer function to fieldOptions is not working
### Describe your motivation

As there is no such a property as `customFields` in AutoForm, `fieldOptions` was supposed to taking care of both replacing a field as well as rendering additional fields to the form. Without this, for adding one additional field, the user need to customize the whole form layout. Also, this can be helpful when dealing with DTOs and nested objects.

Suggestion:
If the `property` for a fieldOption doesn't match any of the existing data properties, it should be rendered as an extra field on the page.

_No response_

### Additional context

_No response_

platosha:
Fixes #1743

This allows to go from this:

... to this:

by specifying the column renderer like so:

```tsx
function DepartmentRenderer({item}: {item: Employee}) {
  return <>{item.department.name}</>;
}

export function GridWithEntityReferences(): JSX.Element {
  return <AutoGrid service={EmployeeService} model={EmployeeModel} noHeaderFilters
   columnOptions={{
     "department.name": {
       renderer: DepartmentRenderer,
       header: "Department"
     },
   }}/>;
}
```

### DIFF
--- a/packages/ts/react-crud/src/model-info.ts
+++ b/packages/ts/react-crud/src/model-info.ts
@@ -1,14 +1,15 @@
 import {
+  AbstractModel,
   BooleanModel,
   NumberModel,
   StringModel,
   _meta,
   createDetachedModel,
-  type AbstractModel,
   type DetachedModelConstructor,
   type ModelMetadata,
   EnumModel,
   ObjectModel,
+  _parent,
 } from '@hilla/form';
 
 export type PropertyType =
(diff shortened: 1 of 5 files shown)


packages/ts/react-crud/src/model-info.ts
sissbruecker, line 1: I think there might be some confusion about the intention of the linked issue and what is implemented here in this PR.

From what I understand, this PR adds a column/field for an existing property to the default columns / fields when you configure a renderer for a property.

As for the issue, I think the intention was to be able to add fields that are not based on a property, similar to how you can configure a custom column that is not based on a property. At least for `AutoGrid` the "custom columns" terminology basically refers to "this column/field does not have an underlying property", instead of "this column/field is not shown by default". If you want to show a property that is not visible by default, you can already use `visibleColumns` / `visibleFields`.

So I'm not quite convinced of the value here. It could also be confusing that the additional column / field only shows up if you configure a renderer. Shouldn't it show up if you configure a header, label, or anything else as well?

PR #1910, opened 2023-12-20
feat: useDataProvider hook

sissbruecker:
Adds a `useDataProvider` hook that can be used to connect a regular `Grid` to a `ListService`. The hook returns a `gridRef` that needs to be passed to the grid in order to apply the data provider to it:
```tsx
const { gridRef } = useDataProvider(MyListService);

<Grid ref={gridRef}>
  ...
</Grid>
```

The hook allows passing a filter, and returns a function for refreshing the grid's data:
```ts
const { gridRef, refresh } = useDataProvider(MyListService, myFilter);
```

It also works with `GridSortColumn` / `GridSorter` for sorting the grid as long as the `path` matches a property on the model class. It does not support `GridFilterColumn` at this point, developers would have to implement an external filter UI or custom header filters to produce a `FilterUnion` that can be passed to the hook.

Passing the ref to the grid may be a little odd, but the data provider implementation requires the grid instance for a couple of reasons:
- It automatically assigns the data provider function
- If the service does not implement `CountService`, an infinite data provider is used. It's implementation needs to access the grid's cache.
- Updating the filter requires to clear the grid's cache, which the data provider can then do automatically when the filter changes.
- The hook also exposes a `refresh` function to reload data, which requires the grid instance.

For comparison, if we would only have a utility for creating a data provider function, the code would look something like this while lacking support for services that do not implement `CountService`:
```tsx
const gridRef = useRef<GridElement>();
const dataProvider = useMemo(() => createDataProvider(MyListService, filter), [filter]);

function refresh() {
  gridRef.clearCache();
}

<Grid ref={gridRef} dataProvider={dataProvider}>
  ...
</Grid>
```


taefi:
It looks great! 
As mentioned, the fact that users are obtaining the `gridRef` from the hook, and not passing it to the hook is a bit strange for me. But, the way this hook is simplifying the usage by minimizing the number of inputs to just the service and the filter instance is very appealing. Not sure, if there could be a better API design considering how easy to use this hook is.

Not sure, if *not* passing any `dataProvider` to the Grid is confusing the users or not. I guess, a good documentation can prevent confusions, if any.

If we consider providing utilities instead, could having two separate utility functions be of any help to overcome the limitation for the List services (that don't implement count)?

For example, What if we provide `createDataProvider` that accepts `ListService` and an additional function parameter for providing the size (e.g. from the grid's cache), and `createSizedDataProvider`  that accepts only the implementations of `ListService & CountService`?

taefi:
After an internal discussion, decided to close this in favor of #1934

PR #1912, opened 2023-12-20
fix: add name property to Validator interface

### LINKED ISSUE #1840: JSR380 validation error message interpolation does not work in production mode
### Describe the bug

https://hilla.dev/docs/lit/guides/forms/binder-validation#validation-message-interpolation
suggests to use validator.constructor.name to construct the key for localization message lookup.

In production mode, class names are minified (or what causes this), 'NotBlank' changes to e.g. 'Yg' and the lookup fails.
I.e. the key is 'validationError.Yg' and the translation file contains 'validationError.NotBlank'.

JSR380 validation messages are always shown in English.
The documentation is misleading - it appears that there is a clean way to do this, but it does not work in production mode. 

### Expected-behavior

Hilla should provide a clean and working way to interpolate JSR380 messages. Documentation and sample code should be fixed.

    Idea 1:
        Exempt the jakarta.validation.constraints.* from minification (or what causes this). How?
        It will work with the Hilla sample interpolation logic.

    Workaround1:
        Define an explicit message for every validation annotation usage. This works, but requires
        a lot of unnecessary and duplicate code in annotations. E.g.
        @Size(max = 15, message = "jsr380Error.Size")
        Then move all the translations in LANG_CODE.json from validationError to jsr380Error subtree.
        Use the following interpolation logic:
        Binder.interpolateMessageCallback = (message, validator, binderNode) => {
            if(message.startsWith('customValidationError.') || message.startsWith('jsr380Error.')) {
                if (translateConfig.lookup(message, translateConfig)) {
                    return get(message, validator as any);
                }
            }
            // Fall back to original message if no translations are found
            return message;
        };

    Workaround2:
        Detect the validator name from its message in localization.ts. This is bad.
        It will work with the Hilla sample interpolation logic (to do: check DecimalMin, DecimalMax).
        let validatorName : String = validator.constructor.name;
        if(message === 'must not be blank') {
            validatorName = 'NotBlank';
        } else if(message === 'must be true') {
            validatorName = 'AssertTrue';
        } else if(/size must be between [0-9]+ and [0-9]+/.test(message)) {
            validatorName = 'Size';
        }
        key = `validationError.${validatorName}`;

### Reproduction

Follow the https://hilla.dev/docs/lit/guides/forms/binder-validation#validation-message-interpolation and run the app in production mode.

### System Info

Hilla 2.4.0 (it is very probable that it applies also to previous versions)

taefi:
## Description
This adds the possibility of defining a name
for all the validators implementing this
interface, which helps to preserve validator's
name in production mode, which is needed when
interpolating messages for client-side validators
(including those added to the binder based on the
existence of JSR380 jakarta validation annotations 
on the endpoints transfer types).

Fixes #1840

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist

### DIFF
--- a/packages/ts/form/src/Validators.ts
+++ b/packages/ts/form/src/Validators.ts
@@ -59,7 +59,7 @@ export class Required<T> extends AbstractValidator<T> {
     return value !== undefined;
   }
 
-  name = 'Required';
+  readonly name: string = 'Required';
 }
 
 function _asValidatorAttributes(attrs: PatternAttributes | RegExp | ValueNumberAttributes | number | string) {
--- a/packages/ts/form/src/Validators.ts
+++ b/packages/ts/form/src/Validators.ts
@@ -43,6 +43,7 @@ abstract class AbstractValidator<T> implements Validator<T> {
   }
 
   abstract validate(value: T): Promise<boolean> | boolean;
+  abstract get name(): string;
 }
 
 export class Required<T> extends AbstractValidator<T> {
@@ -57,6 +58,8 @@ export class Required<T> extends AbstractValidator<T> {
     }
     return value !== undefined;
   }
+
+  name = 'Required';
 }
 
 function _asValidatorAttributes(attrs: PatternAttributes | RegExp | ValueNumberAttributes | number | string) {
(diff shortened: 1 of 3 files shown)


packages/ts/form/src/Validators.ts
```diff
@@ -57,6 +58,8 @@ export class Required<T> extends AbstractValidator<T> {
     }
     return value !== undefined;
   }
+
+  name = 'Required';
```
sissbruecker, line 62:
```suggestion
  readonly name: string = 'Required';
```

sissbruecker (approved):

PR #1930, opened 2023-12-22
feat: add customizable form header to auto crud

sissbruecker:
Adds a customizable header above the form in `AutoCrud`. By default it shows `New item` and `Create item` depending on whether a new item is created or an existing one edited. It's also possible to customize the header by providing a function that renders a custom JSX element based on the edited item and the disabled state of the form. The header is part of the `AutoCrud` for now instead of the `AutoForm`. Otherwise it would have been a challenge to move it into the header of the dialog that shows up on smaller screens.

Disabled:

Bildschirmfoto 2023-12-22 um 09 33 18

Editing:

Bildschirmfoto 2023-12-22 um 09 33 25

Creating:

Bildschirmfoto 2023-12-22 um 09 33 32

### DIFF
--- a/packages/ts/react-crud/src/autocrud.tsx
+++ b/packages/ts/react-crud/src/autocrud.tsx
@@ -12,10 +12,27 @@ import { type ComponentStyleProps, registerStylesheet } from './util.js';
 
 registerStylesheet(css);
 
+type AutoCrudFormHeaderRenderer<TItem> = (
+  editedItem: TItem | null,
+  disabled: boolean,
+) => JSX.Element | null | undefined;
+
 export type AutoCrudFormProps<TModel extends AbstractModel> = Omit<
   Partial<AutoFormProps<TModel>>,
   'disabled' | 'item' | 'model' | 'onDeleteSuccess' | 'onSubmitSuccess' | 'service'
->;
+> &
+  Readonly<{
+    /**
+     * A custom renderer function to create the header for the form. The
+     * function receives the edited item as the first parameter, and a boolean
+     * indicating whether the form is disabled as the second parameter. The
+     * edited item is `null` when creating a new item.
+     *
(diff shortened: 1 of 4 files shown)


packages/ts/react-crud/src/autocrud.tsx
```diff
@@ -69,6 +86,11 @@ export type AutoCrudProps<TModel extends AbstractModel = AbstractModel> = Compon
     gridProps?: AutoCrudGridProps<Value<TModel>>;
   }>;
 
+function defaultFormHeaderRenderer<TItem>(editedItem: TItem | null, disabled: boolean): JSX.Element | null | undefined {
+  const style = { color: disabled ? 'var(--lumo-disabled-text-color)' : 'var(--lumo-text-color)' };
+  return editedItem ? <h3 style={style}>Edit item</h3> : <h3 style={style}>New item</h3>;
+}
```
taefi, line 92: Isn't it better to create the default title based on the name of the Model name e.g. Person, Address, etc., and save the `customFormHeaderRenderer` for more complex titles? IMO, something like `New Person` has a better chance of remaining as it is over `New Item`, so a higher chance of less lines of codes.

taefi (changes requested):

packages/ts/react-crud/src/autocrud.tsx
sissbruecker: That's a good idea, but I don't know how we would solve that currently. The only way to get the name of the model is to use the class name, which probably has the same issue as the validators in that it gets minified in production. Unless we add the name as a property somewhere this doesn't seem feasible. 

I would suggest to use the PR as is, and create an improvement ticket for using the model name, together with a hint that the generator needs to be improved to make this possible.

taefi (approved):

PR #1934, opened 2023-12-22
feat: add useDataProvider hook

sissbruecker:
As an alternative to #1910, this implements a hook that returns a data provider function that can be passed to the grid:
```tsx
const { dataProvider } = useDataProvider(MyListService);

<Grid dataProvider={dataProvider} />
```

It supports the same filter + refresh functionality as the implementation in #1910. In order to make this work, the hook returns a new data provider function reference when the filter changes or refresh is called, which in turn causes grid to reload.

This also requires making some changes to the data provider implementation itself, as it currently requires the grid element for a number of features. In order to remove the grid reference the following changes were necessary:
- The infinite data provider now uses its cached filtered count instead of the grid's cache size for detecting whether to shrink the size or not. Those values should be equivalent, as long as `reset` is called on the data provider whenever the grid should be reloaded or the filter changes.
- Removed the refresh functionality from the data provider itself. Instead, whatever uses the data provider needs to both reset the data provider and trigger a reload of the grid.
- Removed the automatic reload of the grid when the filter changes in the data provider. Instead, whatever uses the data provider needs to both update the filter in the data provider and then trigger a reload of the grid.

The last two points don't concern developers really, as these automatisms are provided by `AutoGrid` and `useDataProvider`.


taefi:
Thanks @sissbruecker for this alternative implementation. IMO, both the usage and the implementation are more conventional this way. The refactoring done for removing the dependency of the DataProvider on the Grid makes it more clear and less coupled. I'm in favor of this implementation the one suggested in #1910.

cromoteca (approved):

cromoteca (changes requested):
It's weird not to see `useDataProvider` in a TS test

Lodin (approved):
LGTM

cromoteca (approved):

PR #1938, opened 2023-12-28
refactor: rewrite preparation script in TS to avoid cross-platform issues

Lodin:


ZheSun88:
summarize :
- log some more info about which and where the generated files are
- remove the `JavaVersion: {{version}}` line from hilla-versions.json

- (?) maybe change the `branch` value to use platform branch instead of release tag so we dont need to update that version as hilla release is following platform

ZheSun88 (approved):
look good to me

ZheSun88 (approved):

PR #1939, opened 2024-01-02
feat: add form level validation error from entity level validator

### LINKED ISSUE #1674: [autoform] Show type validation errors in the form
### Describe your motivation

Field errors are shown with validation for each field, but the whole type(class, record...) validations do not appear in the form.

Validation works, but the user should see something on the page as well. 
Unexcepted errors are already shown on the form bottom. 

```java
@Constraint(validatedBy = AdressValidator.class)
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
public @interface ValidAddress {
    String message() default "Please verify that zip code is correct in the provided country";
    Class<?>[] groups() default { };
    Class<? extends Payload>[] payload() default { };
}

...

@ValidAddress
public record Adress(...  @NotNull String country, @NotNull String zipCode, ...) { }
```

Validation error message should be shown in the same place as unexpected errors.

image

krissvaa:
Add form-level validation errors from validators with no property.

Fixes #1674

### DIFF
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -289,7 +289,7 @@ export function AutoForm<M extends AbstractModel>({
   function handleSubmitError(error: unknown) {
     if (error instanceof ValidationError) {
       const nonPropertyError = error.errors.find((validationError) => !validationError.property);
-      if (nonPropertyError) {
+      if (nonPropertyError?.validatorMessage) {
         setFormError(nonPropertyError.validatorMessage);
       }
     } else if (error instanceof EndpointError) {
--- a/packages/ts/react-crud/src/autoform.tsx
+++ b/packages/ts/react-crud/src/autoform.tsx
@@ -286,6 +286,23 @@ export function AutoForm<M extends AbstractModel>({
     }
   }, [item]);
 
+  function handleSubmitError(error: unknown) {
+    if (error instanceof ValidationError) {
+      const nonPropertyError = error.errors.find((validationError) => !validationError.property);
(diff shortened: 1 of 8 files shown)


packages/ts/react-crud/src/autoform.tsx
```diff
@@ -286,6 +286,23 @@ export function AutoForm<M extends AbstractModel>({
     }
   }, [item]);
 
+  function handleSubmitError(error: unknown) {
+    if (error instanceof ValidationError) {
+      const nonPropertyError = error.errors.find((validationError) => !validationError.property);
```
taefi, line 291: What happens if more than one record level validation fails? We only show one of the errors sent back from the endpoint?

taefi (commented):
The way individual property validators and record level validators are mixed (prior to this PR) is a bit messy. I wish we could first have some tests to define what should be the expected outcome in different cases, and then we may be able to remove the unnecessary complexity from the implementation.

packages/ts/react-crud/src/autoform.tsx
krissvaa: Will show all errors

taefi: Would it make sense that instead of joining error messages with `'; '` we render them under the form like this:
```
There are validation errors in the form:
 * foobar
 * just a message
 * ...
``` 
I think this improves the UX to make the errors more understandable. WDYT?

krissvaa: Added this options

taefi (approved):

PR #1946, opened 2024-01-09
feat: Add aria-controls to autocrud grid to control the form

### LINKED ISSUE #1666: [autocrud] Create semantic association between grid and form
For accessibility it should be clear that selecting something something in the grid edits that item in the form. Technically the grid "controls" the form, and there should be an aria attribute to reflect that association.

krissvaa:
Add aria-controls to autocrud grid to control the form

Fixes #1666


packages/ts/react-crud/src/autocrud.tsx
```diff
@@ -146,6 +146,7 @@ export function AutoCrud<TModel extends AbstractModel>({
           setItem(activeItem ?? undefined);
         }}
         ref={autoGridRef}
+        aria-controls={`${id ?? 'auto'}-form`}
```
taefi, line 149: In case the there are more than one AutoCrud in a view, and the `id` prop is `undefined/null`, the AutoForm's `id` and the `aria-controls` of AutoGrid would clash, right?

taefi (changes requested):
Could we have a couple of tests validating the presence of `aria-controls` and the equality of its value with the `id` of the form in the two situations: 
- an id is provided
- no id is passed in

packages/ts/react-crud/src/autocrud.tsx
krissvaa: Now id is generated and will be unique everytime

taefi (approved):
LGTM

PR #1951, opened 2024-01-11
chore: use snapshots for Flow and TestBench

cromoteca:
Set both to `24.4-SNAPSHOT`. Other changes are findings while making tests pass.


ZheSun88:
i pushed a commit to update the testbench version

taefi (approved):

PR #1985, opened 2024-01-24
feat: add file-based router

### LINKED ISSUE #2024: [FS routing] filename conventions for views, layouts, and helpers
As a Hilla application developer, I want to be able to create non-view `.tsx` files among route views.

Supported conventions and use cases for the `frontend/views/`:

- Files and folders prefixed with `_` are ignored by FS router, can be used for shared utils and helpers.
- Except above, every `.tsx` file defines a view route for FS router.
- Every file prefixed with `$` is processed by FS router, but is not a regular view route.
  - `$index.tsx` defines index view route for a directory.
  - `$layout.tsx` defines a layout route for a directory.
  - Any other `$` prefixed file throws an exception.
 
This includes updates to Hilla documentation to describe and use the above conventions.

Lodin:
Fixes #2024


Lodin:
Ok, the issue with codecov is resolved. It seems there were some problems with [Github Operations](https://www.githubstatus.com/history).

packages/ts/hilla-file-router/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@vaadin/hilla-file-router",
+  "version": "24.4.0-alpha1",
```
taefi, line 3:
```suggestion
  "version": "24.4.0-alpha2",
```

packages/ts/hilla-file-router/src/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+  module?: T;
+  children?: ReadonlyArray<AgnosticRoute<T>>;
+}>;
+
+export function processPattern(blank: string): string {
+  return blank
+    .replaceAll(/\{\.{3}.+\}/gu, '*')
+    .replaceAll(/\{{2}(.+)\}{2}/gu, ':$1?')
+    .replaceAll(/\{(.+)\}/gu, ':$1');
+}
```
taefi, line 70: I suggest:
1. Naming this function to reflect what it exactly does.
2. a better name for the input parameter, I really don't understand what `blank` means here, especially when I look at the possible REGEX patterns that can match the contents of the input.
3. and to define either named variables for each REGEX or at least comments with examples of what each REGEX matches and replaces, for instance: 
```
.replaceAll(/\{\.{3}.+\}/gu, '*') //  e.g. `{...name}` would become `*.`
.replaceAll(/\{{2}(.+)\}{2}/gu, ':$1?') // e.g. `{{userId}}` becomes `:userId?`
.replaceAll(/\{(.+)\}/gu, ':$1'); // e.g.  `{id}` becomes `:id`
```

packages/ts/hilla-file-router/src/generateJson.ts
```diff
@@ -0,0 +1,74 @@
...
+
+function* walkAST(node: Node): Generator<Node> {
+  yield node;
+
+  for (const child of node.getChildren()) {
+    yield* walkAST(child);
+  }
+}
+
+export default async function generateJson(views: RouteMeta, configExportName: string): Promise<string> {
```
taefi, line 35: I suggest renaming this function and module to reflect what it exactly does, such `extractViewConfigs`, as `generateJson` is a very common name.

packages/ts/hilla-file-router/src/vite-plugin-file-router.ts
```diff
@@ -0,0 +1,109 @@
...
+   */
+  configExportName?: string;
+}>;
+
+type GeneratedUrls = Readonly<{
+  json: URL;
+  code: URL;
+}>;
+
+async function generate(code: string, json: string, urls: GeneratedUrls) {
```
taefi, line 43: If this function creates some files, I suggest renaming it to `createXXXfile(s)` or simply `writeToFile`, as `generate` is misleading.

packages/ts/hilla-file-router/src/vite-plugin-file-router.ts
```diff
@@ -0,0 +1,109 @@
...
+  outDir: URL,
+  generatedUrls: GeneratedUrls,
+  extensions: readonly string[],
+  configExportName: string,
+): Promise<void> {
+  const routeMeta = await collectRoutes(viewsDir, { extensions });
+  const code = generateRoutes(routeMeta, outDir);
+  const json = await generateJson(routeMeta, configExportName);
+
+  await generate(code, json, generatedUrls);
```
taefi, line 58:
```suggestion
  const viewRoutesCode = generateRoutes(routeMeta, outDir);
  const viewConfigsJson = await generateJson(routeMeta, configExportName);

  await generate(viewRoutesCode, viewConfigsJson, generatedUrls);
```

packages/ts/hilla-file-router/src/generateRoutes.ts
```diff
@@ -0,0 +1,94 @@
...
+      transformer((node) =>
+        ts.isImportDeclaration(node) && (node.moduleSpecifier as StringLiteral).text === 'IMPORTS' ? imports : node,
+      ),
+      transformer((node) => (ts.isIdentifier(node) && node.text === 'ROUTE' ? routes : node)),
+    ],
+  );
+
+  const file = createSourceFile(routeDeclaration, 'views.ts');
+
+  return printer.printFile(file);
```
taefi, line 93: Just a question:
Why the view routes are being persisted to a file here, and then in the vite plugin they are being read and written to another file? Couldn't the transformed `routeDeclaration` be returned and persisted there once?

taefi (changes requested):

packages/ts/hilla-file-router/src/react.ts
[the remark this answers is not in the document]
Lodin: It is intended to be used as a main import file. So, it's going to be like `import {toReactRouter} from '@vaadin/hilla-file-router/react.js';`. I believe that making the file name even more detailed is redundant.

packages/ts/hilla-file-router/src/utils.ts
```diff
@@ -0,0 +1,114 @@
...
+  lazy?: boolean;
+
+  /**
+   * If set to false, then the route will not be registered with React Router,
+   * but it will still be included in the main menu and used to configure
+   * Spring Security
+   */
+  register?: boolean;
+
+  menu?: Readonly<{
```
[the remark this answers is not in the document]
Lodin: For now, that's kind of a placeholder for the further implementation, so I'm not going to fix that part is this PR. However, I will address that comment in my follow-on PR.

packages/ts/hilla-file-router/src/generateRoutes.ts
Lodin: Yeah, that looks a bit misleading here but that's more like a problem of Typescript API naming. The `printFile` name here means that the Typescript prints a structure called `SourceFile` which basically contains the content of the whole file AST. It has pretty thin relation to the real file.

packages/ts/hilla-file-router/src/react.ts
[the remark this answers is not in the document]
taefi: Well, for importing in the user codes, I guess even `'@vaadin/hilla-file-router/react.js'` could become more delightful, if we have an `index.ts` that exports everything needed to be exposed from `reactRouterUtils` and users can have something like the following instead:
```tsx
import { toReactRouter } from '@vaadin/hilla-file-router';
```
Just a thought.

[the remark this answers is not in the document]
Lodin: Ok, after some thinking, I decided to make it clearer. Now we have a `vite-plugin-...` file and a `runtime` file exported, as well as appropriate folders. So, it makes a clear distinction for the user to import either a runtime part or a Vite plugin

packages/ts/hilla-file-router/src/vite-plugin/collectRoutesFromFS.ts
```diff
@@ -0,0 +1,68 @@
...
+  extensions: readonly string[];
+  parent?: URL;
+}>;
+
+function cleanUp(blank: string) {
+  return blank
+    .replaceAll(/\{\.{3}(.+)\}/gu, '$1')
+    .replaceAll(/\{{2}(.+)\}{2}/gu, '$1')
+    .replaceAll(/\{(.+)\}/gu, '$1');
+}
```
taefi, line 22: I guess a slightly different black magic is in effect here, which can leverage from some comments for the maintainers.

packages/ts/hilla-file-router/src/vite-plugin/createRoutesFromFS.ts
```diff
@@ -0,0 +1,95 @@
...
+      (statement as VariableStatement).declarationList.declarations[0].initializer as ObjectLiteralExpression,
+    [
+      transformer((node) =>
+        ts.isIdentifier(node) && node.text === 'CHILDREN' ? ts.factory.createArrayLiteralExpression(children) : node,
+      ),
+    ],
+  );
+}
+
+export default function createRoutesFromFS(views: RouteMeta, generatedDir: URL): string {
```
taefi, line 53: nit picking: 
```suggestion
export default function createRoutesFromRoutesMeta(views: RouteMeta, generatedDir: URL): string {
```

taefi (changes requested):

packages/ts/hilla-file-router/src/react.ts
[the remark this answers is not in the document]
taefi: So, now the user code would be like
```ts
import { toReactRouter } from '@vaadin/hilla-file-router/runtime.js';
```
right?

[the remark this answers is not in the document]
Lodin: Yep, exactly

taefi (approved):
LGTM!

platosha (approved):

PR #1988, opened 2024-01-25
chore: figure out correct goalPrefix when hilla-maven-plugin is packed as vaadin-maven-plugin

manolo:


packages/java/engine-runtime/src/main/java/com/vaadin/hilla/internal/AbstractTaskEndpointGenerator.java
```diff
@@ -70,9 +72,19 @@ protected void prepareEngineConfiguration()
...
             try {
+            	String goal = "hilla:configure";
+            	File pom = new File("pom.xml");
+            	if (pom.exists()) {
+                    String content = new String(Files.readAllBytes(Paths.get(pom.toURI())));
+                    String regex = "<plugin>\\s*<groupId>\\s*com\\.vaadin\\s*</groupId>\\s*<artifactId>\\s*vaadin-maven-plugin\\s*</artifactId>";
+                    if (Pattern.compile(regex, Pattern.DOTALL).matcher(content).find()) {
+                    	goal = "vaadin:configure";
+                    };
+            	}
```
taefi, line 83: Are we trying to support a project that uses V24.4+, but instead of using Vaadin bom, user manually changes the dependencies e.g. from `hilla-maven-plugin` instead of `vaadin-maven-plugin`? If not, please describe what are we trying to solve here by parsing the immediate `pom.xml` file? What happens in multi-module projects?
Why aren't we simply always run the `vaadin:confgiure` goal?

tltv: This is related to unifying task in platform where `vaadin-maven-plugin` "becomes" a `hilla-maven-plugin` and works with flow too ([4944](https://github.com/vaadin/platform/issues/4944)). 

This change could fail in multi-module project indeed. 
Just dropping Manolo's own idea here for a fix: run first with `vaadin:configure` and if it fails, then rerun with `hilla:configure`.

mcollovati: Another option could be `mvn help:describe -Dcmd='hilla:configure'`, to prevent false negatives if the `configure` goal may exist but fail for different reasons.

taefi: > Just dropping Manolo's own idea here for a fix: run first with vaadin:configure and if it fails, then rerun with hilla:configure.
Still I don't get why a `hilla:configure` should exist at all after the unified platform is out? If Hilla is not excluded from the dependencies by the "Vaadin" user, both Hilla (and Flow if needs it) should simply call the `vaadin:configure`. Why we are trying to support a fallback? Unless I'm missing a point from a discussion happened somewhere else.

taefi: If it is for internal Hilla usages e.g. tests, I would suggest getting rid of the complexities altogether. What if Hilla does not provide a separate plugin, and then Platform tries to repackage it at the platform level just to rename it? Flow should remain the same IMO, Hilla publishes it's own goals under `vaadin` names, then platform can just republish the artifact as it is (would be).

cromoteca:
We could also play with the `goalPrefix` like we did before: https://github.com/vaadin/hilla/blob/9ed14940077118014ca8dc3bd15c396507b3c83b/packages/java/maven-plugin/pom.xml#L96

That way, two different plugins can be seen under the same name, provided that they are not used at the same time.

Although it's not clear to me what the final goal is, and why two names are needed.

packages/java/engine-runtime/src/main/java/com/vaadin/hilla/internal/AbstractTaskEndpointGenerator.java
taefi: Anyway, if we want to check the availability of hilla-maven-plugin before deciding the about the goal to be executed, I'm more inclined towards @mcollovati suggestion of executing `mvn help:describe -Dcmd=hilla:configure` with a `-q` flag to just check its existence. Parsing the immediate pom.xml is a bit more prone to failures...

taefi:
> We could also play with the `goalPrefix` like we did before:
> 
> https://github.com/vaadin/hilla/blob/9ed14940077118014ca8dc3bd15c396507b3c83b/packages/java/maven-plugin/pom.xml#L96
> 
> That way, two different plugins can be seen under the same name, provided that they are not used at the same time.
> 
> Although it's not clear to me what the final goal is, and why two names are needed.

I was kinda thinking the same, the only usage I can think of for a `hilla-maven-plugin` is when we want to build/test Hilla project itself and still the platform plugin called `vaadin-maven-plugin` is not released. But, we can keep having the `hilla-maven-plugin` and include that dependency in Hilla project's tests, but the goal prefix can change to `vaadin` so that we don't need to distinguish between `hilla` or `vaadin`.

manolo:
Added a commit setting the goalPrefix as @taefi suggests

tltv (approved):

PR #1990, opened 2024-01-25
test: Use vaadin-router in tests where Lit is used

mshabarov:
Sets the tests to use vaadin-router by default as they almost all written in Lit.
When it's a React-based test, it sets react-router.
Also sets the serverSideRoutes item for React-based tests as a workaround for know Flow issue.

Fixes Hilla snapshot validation issues.


mshabarov:
Waits for https://github.com/vaadin/flow/pull/18542

platosha (approved):

PR #1992, opened 2024-01-26
chore: change hilla-dev to hilla-dev-mode in hilla

tltv:
Related-to: vaadin/platform/issues/4946

### DIFF
--- a/packages/java/hilla-bom/pom.xml
+++ b/packages/java/hilla-bom/pom.xml
@@ -106,6 +106,11 @@
             </dependency>
             <dependency>
                 <groupId>com.vaadin.hilla</groupId>
+                <artifactId>hilla-dev-mode</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.vaadin.hilla</groupId>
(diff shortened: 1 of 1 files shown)


manolo (approved):

mshabarov:
@manolo @tltv how in the end the exclusion of the development dependencies is supposed to work ? Initially, I assumed that `hilla-dev` would be a part of `vaadin-dev` and for production mode one should exclude only `vaadin-dev` as in previous V24 versions.

But with this PR and PRs in platform, I'm a bit confused how it will be achievable in the end. Could you please give your thoughts?

tltv:
> @manolo @tltv how in the end the exclusion of the development dependencies is supposed to work ? Initially, I assumed that `hilla-dev` would be a part of `vaadin-dev` and for production mode one should exclude only `vaadin-dev` even for hybrid apps, as in previous V24 versions.
> 
> But with this PR and PRs in platform, I'm a bit confused how it will be achievable in the end. Could you please give your thoughts?

Maybe this discussion can continue in one place here https://github.com/vaadin/platform/issues/4946.

Issue #1979, opened 2024-01-22
Remove hilla-*-spring-boot-starter modules


mshabarov:
### Describe your motivation

Remove `hilla-spring-boot-starter` and `hilla-react-spring-boot-starter` modules from Hilla projects after https://github.com/vaadin/platform/issues/4934, as they wouldn't be needed when Hilla is part of Platform.

### Additional context

Vaadin/Hilla 24.4.

PR #1995, opened 2024-01-29
chore: remove hilla spring boot starter modules

tltv:
Removes `hilla-spring-boot-starter` and `hilla-react-spring-boot-starter` modules. Platform's `vaadin-spring-boot-starter` should be used instead.

Fixes: #1979


manolo (dismissed):

pom.xml
```diff
@@ -27,8 +27,6 @@
     <module>packages/java/hilla-dev</module>
     <module>packages/java/hilla-dev-mode</module>
     <module>packages/java/hilla-react</module>
-    <module>packages/java/hilla-spring-boot-starter</module>
-    <module>packages/java/hilla-react-spring-boot-starter</module>
```
knoobie, line 31: Just asking: 

Now it would be breaking change in a minor version (well Hilla 2 -> 24 is major... so dismiss if not applicable) 

Otherwise you could let those modules still be alive and add maven relocation to say: hey it's deprecated; use the other artefacts: https://maven.apache.org/ref/3.2.1/maven-model/maven.html#class_relocation

manolo (approved):

PR #2015, opened 2024-02-01
feat: Add template package.json for Hilla components

mshabarov:
Adds template files where Hilla components packages are listed.
Replaces `AtNpmPackage` annotations in `hilla` and `hilla-react`, so now Flow can add these packages to package.json automatically depending on whether Hilla is used and React or Lit is used.

Places these templates in `hilla` and `hilla-react`.

So basically developers don't need to do manual swaps or exclusions in their projects:

```xml
<dependency>
   <groupId>com.vaadin</groupId>
   <artifactId>vaadin-spring-boot-starter</artifactId>
   <version>${vaadin.version}</version>
   <exclusions>
       <exclusion>
            <groupId>com.vaadin.hilla</groupId>
            <artifactId>hilla-react</artifactId>
        </exclusion>
      </exclusions>
</dependency>
<dependency>
      <groupId>com.vaadin.hilla</groupId>
      <artifactId>hilla-lit</artifactId>
      <version>${vaadin.version}</version>
</dependency>
```

Related-to https://github.com/vaadin/flow/issues/18564

Depends-on https://github.com/vaadin/flow/pull/18597


mshabarov:
`HillaAppInitUtility::detectFramework` is still based on `hilla` and `hilla-react` artifacts being added into build configurations.
Thus, I cannot remove `hilla-react` artifact right now without changing this part of functionality.

This is proposed for a separate PR, as it's not directly related to goals in https://github.com/vaadin/flow/issues/18564 and most likely needed for `init-app` goal that is going to be removed soon.

mshabarov:
Fails with the same error as other PRs, so probably not caused by this change:

```
ProductionBuildFunctionalIT > validateProductionBuildOutput() FAILED
```

packages/java/hilla-react/src/main/resources/com/vaadin/flow/server/frontend/dependencies/hilla/components/lit/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "hilla-components-dependencies",
```
caalador, line 2: I feel the lit package should not be available in the react module.

packages/java/hilla/src/main/resources/com/vaadin/flow/server/frontend/dependencies/hilla/components/react/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "hilla-components-dependencies",
```
caalador, line 2: I feel the react package should not be available in the lit module.

caalador (changes requested):
Only question I have really is that should we actually have

packages/java/hilla-react/src/main/resources/com/vaadin/flow/server/frontend/dependencies/hilla/components/lit/package.json
mshabarov: My idea was to have only one module `hilla` where you have both packages and this modules isn't bound to lit or react, it just have Hilla components. 

Currently, both `hilla` and `hilla-react` actually have the same content, because init-app logic depends on what artifact is used. 

But init-app will be moved from the Hilla plugin, so I expect `hilla-react` module to be completely removed soon (in a separate PR). 

So finally we'll have one single module `hilla` with both packages and Flow decides which package to use based on auto-detection.

caalador (approved):

ZheSun88:
what is wrong with the release lable and comment? ;-P

mshabarov:
> what is wrong with the release lable and comment?

Referenced issue in Flow has been merged and released, thus this label is there. Maybe.

platosha (approved):

PR #2018, opened 2024-02-02
chore: active full set of spring module in PR validations

ZheSun88:


packages/java/tests/pom.xml
```diff
@@ -281,24 +281,27 @@
             </modules>
         </profile>
 
+        <!-- the spring module constains submodules to run with different profiles -->
```
Artur-, line 284: Shouldn't that be handled by the `it-modules` profile above?

ZheSun88: the above one will be deactived during PR validation. check here 
https://github.com/vaadin/hilla/blob/main/.github/workflows/validation.yml#L246

Artur- (approved):

ZheSun88:
checked all other branches.. it doesn't have this issue.. so removed all branch labels.

PR #2031, opened 2024-02-05
fix: support fallback calling full plugin name

### LINKED ISSUE #2011: Plugin name clash when running Vaadin 24.4 with multiplatform-runtime
### Describe the bug

Vaadin Flow + Multiplatform-Runtime or Vaadin Hilla + Flow + Multiplatform-Runtime don't work with Vaadin 24.4 latest pre-release as Vaadin tries to call `vaadin:configure` goal that clashes with legacy `vaadin-maven-plugin` having the same name `vaadin`:

When invoking the `vaadin:build-frontend` goal, the `TaskGenerateOpenAPIImpl` node task is executed, that in turn runs `mvn vaadin:configure`, see `AbstractTaskEndpointGenerator` class in Hilla.

The problem is that in an MPR project this will call the framework's maven plugin, resulting in the following error
```
[ERROR] Could not find goal 'configure' in plugin com.vaadin:vaadin-maven-plugin:8.20.0 among available goals clean, compile, compile-report, compile-theme, css, debug, eclipse, eclipseTest, generateAsync, help, i18n, mergewebxml, resources, run, run-codeserver, source-jar, test, update-theme, update-widgetset, upgrade8 -> [Help 1] 
```

Flow invokes the `TaskGenerateOpenAPI` and `vaadin:configure`, because Lookup detects the Hilla class in the classpath, as `vaadin-spring-boot-starter` now depends on Hilla.

### Expected-behavior

Flow and Hilla + MPR project should work with V24.4.

### Reproduction

Run `Session IT` test module in MPR project.

### System Info

Vaadin 24.4 pre-release.

cromoteca:
Closes #2011. Implements the solution suggested in the issue.


packages/java/engine-runtime/src/main/java/com/vaadin/hilla/internal/AbstractTaskEndpointGenerator.java
```diff
@@ -69,26 +73,45 @@ protected void prepareEngineConfiguration()
...
-                                        + "Gradle project.", projectDirectory)))
-                        // Run the first valid runner
-                        .run(null);
-                firstRun = false;
-            } catch (CommandRunnerException e) {
+            var mavenConfigure = MavenRunner.forProject(projectDirectory, "-q",
+                    "vaadin:configure");
+            var mavenConfigureVersion = Platform.getVaadinVersion()
+                    .flatMap(version -> MavenRunner.forProject(projectDirectory,
+                            "-q", "com.vaadin:vaadin-maven-plugin:" + version));
```
mcollovati, line 80: It looks like the goal is missing here
```suggestion
                            "-q", "com.vaadin:vaadin-maven-plugin:" + version+":configure"));
```

packages/java/engine-runtime/src/test/java/com/vaadin/hilla/internal/AbstractTaskEndpointGeneratorTest.java
```diff
@@ -47,4 +53,52 @@ public void execute() throws ExecutionFailedException {
...
+
+        try (var staticMock = mockStatic(MavenRunner.class);
+                var platformMock = mockStatic(Platform.class)) {
+            staticMock
+                    .when(() -> MavenRunner.forProject(any(), eq("-q"),
+                            eq("vaadin:configure")))
+                    .thenReturn(Optional.of(mockConfigureNotFound));
+            staticMock
+                    .when(() -> MavenRunner.forProject(any(), eq("-q"),
+                            eq("com.vaadin:vaadin-maven-plugin:1.0.0")))
```
mcollovati, line 87:
```suggestion
                            eq("com.vaadin:vaadin-maven-plugin:1.0.0:configure")))
```

mcollovati (commented):
The change looks good. 

However, we have a problem: `Platform.getVaadinVersion()` uses the current thread context classloader to get the properties file with the version in `vaadin-core`, but when the code runs as part of a maven goal execution, the classpath in use is the one from the plugin, so  `vaadin-core` is not present.
Thus, the method returns an empty Optional and the  `MavenRunner` with full coordinates is not executed

packages/java/engine-runtime/src/test/java/com/vaadin/hilla/internal/AbstractTaskEndpointGeneratorTest.java
```diff
@@ -35,16 +42,67 @@ void shouldThrowIfOutputDirectoryIsNull() {
         }, "Output directory cannot be null");
     }
 
-    static private class TestTaskEndpointGenerator
+    private static final Function<String, URL> resourceFinder = Thread
```
mcollovati, line 45: Is it needed to have this field static? Could potentially cause issues in parallel executions. What about providing the value directly in the call to `super(...)`?

mcollovati (commented):
Tested with MPR, and it works as expected :+1: 
I added a couple of nit comments, other than that LGTM.

packages/java/engine-runtime/src/test/java/com/vaadin/hilla/internal/AbstractTaskEndpointGeneratorTest.java
```diff
@@ -1,11 +1,17 @@
...
+import java.util.Optional;
 
+import com.vaadin.hilla.engine.commandrunner.CommandRunnerException;
+import com.vaadin.hilla.engine.commandrunner.MavenRunner;
 import org.junit.jupiter.api.Test;
 
+import com.vaadin.flow.server.Platform;
 import com.vaadin.flow.server.ExecutionFailedException;
 
+import static org.mockito.Mockito.*;
```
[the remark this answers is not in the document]
cromoteca: We don't. In some rare cases I like a star import. Mockito and asserts are some examples.

PR #2032, opened 2024-02-05
feat: extract and combine routes from fs routing and flow

krissvaa:


Lodin (dismissed):
LGTM

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
```diff
@@ -59,19 +60,23 @@ protected void extractServerViews(
         registry.getRegisteredRoutes().forEach(serverView -> {
             final Class<? extends com.vaadin.flow.component.Component> viewClass = serverView
                     .getNavigationTarget();
-            final String url = "/" + registry.getTargetUrl(viewClass)
-                    .orElseThrow(() -> new RuntimeException(
-                            "Only supporting views without parameters"));
+            final Optional<String> targetUrl = registry.getTargetUrl(viewClass);
+            if (targetUrl.isPresent()) {
+                final String url = "/" + targetUrl.get();
```
Lodin, line 65: I guess, you could do the following here:
```suggestion
            final String url = registry.getTargetUrl(viewClass).map((res) -> "/" + res)
                    .orElseThrow(() -> new RuntimeException(
                            "Only supporting views without parameters"));
```

Lodin (dismissed):
LGTM

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
krissvaa: registry.getTargetUrl(viewClass) will throw "IllegalArgumentException: Navigation target 'com.vaadin.app.RandomView' requires a parameter."
so try/catch is needed here

krissvaa: Dynamic flow route reading will be implemented in the next PR.

Lodin (approved):

PR #2041, opened 2024-02-07
chore: add hilla-dev dependency to tests so as they can be run in devmode

manolo:
This is needed in order not to make hilla tests depend on platform vaadin-dev

### DIFF
--- a/packages/java/tests/pom.xml
+++ b/packages/java/tests/pom.xml
@@ -68,6 +68,11 @@
             <artifactId>engine-runtime</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


packages/java/tests/pom.xml
```diff
@@ -68,6 +68,10 @@
             <artifactId>engine-runtime</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin.hilla</groupId>
+            <artifactId>hilla-dev</artifactId>
```
ZheSun88, line 73: i think you should add a `<version>` tag here.

cromoteca (approved):

PR #2082, opened 2024-02-16
refactor!: remove hilla-react

platosha:
As `hilla` and `hilla-react` have the same content after recent changes,
keeping both is redundant. Let me suggest to keep only `hilla` as the
one with the shorter and simpler name.

Depends on: https://github.com/vaadin/platform/pull/5061


ZheSun88:
platform PR got merged.

cromoteca (approved):

PR #2094, opened 2024-02-22
chore: upgrade spring.boot to 3.1.9 (2.4)

ZheSun88:


taefi:
@ZheSun88 This is not needed anymore, correct?

PR #2106, opened 2024-02-22
chore: remove Lumo / Material class copies

### LINKED ISSUE #2058: hilla-react causes Lumo icons not to be loaded when using Flow
### Describe the bug

`hilla-react` includes a version of com.vaadin.hilla.theme.Lumo.class which is missing all annotations. When it is used instead of the real class, Flow does not load the files needed.

### Expected-behavior

`hilla-react` does not break Flow

### Reproduction

A Flow 24.4 project with `add(new Icon("lumo", "dropdown"))`

### System Info

24.4.0.alpha3

platosha:
Fixes #2058


packages/java/tests/gradle/single-module-tests/gradle.properties
```diff
@@ -1,3 +1,9 @@
 org.gradle.daemon=false
+<<<<<<< Updated upstream
```
Artur-, line 2: No

Artur- (approved):

PR #2128, opened 2024-02-27
fix(file-router): ignore "config" export for HMR

Lodin:
Allows correct HMR update in case `export const config` exists in the route file.


packages/ts/hilla-file-router/src/vite-plugin.ts
```diff
@@ -87,5 +90,20 @@ export default function vitePluginFileSystemRouter({
       server.watcher.on('change', changeListener);
       server.watcher.on('unlink', changeListener);
     },
+    transform(code, id): Promise<TransformResult> | TransformResult {
+      if (id.startsWith(fileURLToPath(_viewsDir))) {
```
Legioth, line 94: Should maybe also ignore files starting with `_` inside the views directory?

Lodin: Right, good point

taefi:
The java test failure would be alleviated after #2130 is merged, stay tuned...

packages/ts/hilla-file-router/src/vite-plugin.ts
Lodin: Done

Legioth (commented):
I don't pretend to have a good enough understanding of how HMR works without some additional explanations but I might still consider this change to be good enough if it's guarded by the integration test that we discussed.

taefi:
> I don't pretend to have a good enough understanding of how HMR works without some additional explanations but I might still consider this change to be good enough if it's guarded by the integration test that we discussed.

Basically, I was struggling with the fact that I really don't know how HMR is working and how that regex is helping there :)

Legioth:
This is basically one of those cases where it would be helpful with some comments in the code to give some additional contest in the form of an examples of what the targeted code looks before the transformation and what it is supposed to look like after.

platosha (approved):

platosha (approved):

PR #2133, opened 2024-02-28
build(hilla): generate package.json resources for Hilla npm components

### LINKED ISSUE #2114: Add @vaadin/hilla-file-router to devDependencies in Hilla / React project
The `@vaadin/hilla-file-router` is missing here: `https://github.com/vaadin/hilla/blob/main/packages/java/hilla/src/main/resources/com/vaadin/flow/server/frontend/dependencies/hilla/components/react/package.json`

Let's add a `devDependency` entry for this new package.

### LINKED ISSUE #2115: Auto-update version numbers in package.json resources
Currently, the `hilla` artifact brings Hilla's own npm packages to users by declaring them in the `package.json` resources. The challenge is that the versions in that file should be updated every release, otherwise it starts to lag behind, which results in shipping old package versions.

As an idea, we could introduce templates for `package.json` resource files, so that the `{{version}}` tag is replaced on build time with the latest one.

platosha:
Fixes #2114
Fixes #2115


scripts/prepare/index.ts
```diff
@@ -30,6 +30,19 @@ await Promise.all(
   ),
 );
 
+console.log('Generating components list package.json resources.');
+
+const components = await readdir(local.components);
```
Lodin, line 35: Wouldn't it be simpler to create two URLs — like `local.components.lit` and `local.components.react`? Or even `local.litPackageJson` and `local.reactPackageJson`? We could avoid unnecessary IO operation (`readdir`).

scripts/prepare/config.ts
```diff
@@ -64,4 +65,8 @@ export const destination = {
     new URL('packages/java/tests/spring/react-signals/src/main/resources/vaadin-core-versions.json', root),
   ],
   themeDir: new URL('packages/java/hilla/src/main/java/com/vaadin/hilla/theme/', root),
+  components: new URL(
```
Lodin, line 68: Do we use this section anywhere? 🤔

Lodin (changes requested):
LGTM but there is a couple of questions

scripts/prepare/index.ts
platosha: We also need their respective destinations, then there would be 4 URLs in total. Not any simpler IMO.

To avoid the IO operation, I've replaced `readdir` with the `lit`/`react` options exported form the config.

scripts/prepare/config.ts
platosha: Yes, it's the destination for the generated files.

.gitignore
```diff
@@ -44,7 +44,7 @@ scripts/prepare/results/
 packages/java/hilla/*.json
 packages/java/hilla-react/*.json
 packages/java/hilla/src/main/java/com/vaadin/hilla/theme/
-packages/java/hilla-react/src/main/java/com/vaadin/hilla/theme/
+packages/java/hilla/src/main/resources/
```
cromoteca, line 47: This one looks a bit unexpected. Which files are supposed to be ignored there?

cromoteca (dismissed):

ZheSun88 (approved):

.gitignore
platosha: There simply aren't any files yet other than generated ones.

Issue #2117, opened 2024-02-26
Fail to create production build with repos extending JpaSpecificationExecutor


tarekoraby:
### Describe the bug

In Hilla 24.4.0.alpha5, the production build creation fails if the project has a `ListRepositoryService` endpoint that takes as input a Spring Data repo that extends `JpaSpecificationExecutor`. 

In that case, the production build fails with
```
[ERROR] Failed to execute goal com.vaadin:vaadin-maven-plugin:24.4.0.alpha5:build-frontend (default) on project skeleton-starter-hilla-react: Could not execute build-frontend goal: Error occured du
ring goal execution: org/springframework/data/jpa/domain/Specification
```

Running Maven with `-X` option shows
```
Caused by: java.lang.ClassNotFoundException: org.springframework.data.jpa.domain.Specification
    at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass (SelfFirstStrategy.java:50)
    at org.codehaus.plexus.classworlds.realm.ClassRealm.unsynchronizedLoadClass (ClassRealm.java:271)
    at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass (ClassRealm.java:247)
    at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass (ClassRealm.java:239)
    at java.lang.Class.getDeclaredMethods0 (Native Method)
    at java.lang.Class.privateGetDeclaredMethods (Class.java:3402)
    at java.lang.Class.getDeclaredMethods (Class.java:2504)
    at com.vaadin.hilla.parser.models.ClassInfoReflectionModel.prepareMethods (ClassInfoReflectionModel.java:216)
    at com.vaadin.hilla.parser.models.ClassInfoModel.getMethods (ClassInfoModel.java:297)
    at com.vaadin.hilla.parser.plugins.backbone.MethodPlugin.scan (MethodPlugin.java:82)
    at com.vaadin.hilla.parser.core.AbstractCompositePlugin.scan (AbstractCompositePlugin.java:47)
    at com.vaadin.hilla.parser.core.AbstractCompositePlugin.scan (AbstractCompositePlugin.java:47)
```

### Expected-behavior

It works

### Reproduction

1. Clone https://github.com/vaadin/skeleton-starter-hilla-react/tree/JpaSpecificationExecutor-error
2. Run `mvn clean package -Pproduction`

### System Info

- Windows 11
[...]

PR #2136, opened 2024-02-28
fix: provide necessary dependency in plugin classpath

cromoteca:
When generating an endpoint that extends `ListRepositoryService`, for some reason that class is now loaded by a different class loader. That didn't happen in Hilla 2.5. That class loader uses the classpath of the plugin, not the one of the project.

This PR provides the necessary dependency at plugin level.

Fixes https://github.com/vaadin/hilla/issues/2117

### DIFF
--- a/packages/java/maven-plugin/pom.xml
+++ b/packages/java/maven-plugin/pom.xml
@@ -58,6 +58,14 @@
       <artifactId>hilla-engine-core</artifactId>
       <version>${project.version}</version>
     </dependency>
+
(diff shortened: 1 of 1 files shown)


platosha (approved):

Artur-:
This is how Flow gets the correct classpath: https://github.com/vaadin/flow/blob/b042a2fb5599e834702417a1a603c397f2eb82ef/flow-plugins/flow-maven-plugin/src/main/java/com/vaadin/flow/plugin/maven/FlowModeAbstractMojo.java#L312-L318

PR #2138, opened 2024-02-28
fix: mark as visited on change in addition to blur

### LINKED ISSUE #2134: [react-form] Validators are not called for DateTimePicker
### Describe the bug

Unlike `DatePicker` and `TimePicker`, with `DateTimePicker` it's not possible to make custom validators work.

The reason is that in React Form, `visited` state is only updated on `blur`, whereas Lit Form uses both `blur` and `change`:

https://github.com/vaadin/hilla/blob/daf1d14b83c2bdcf580f0e45a3a0874b473c859c/packages/ts/lit-form/src/Field.ts#L401-L407

https://github.com/vaadin/hilla/blob/daf1d14b83c2bdcf580f0e45a3a0874b473c859c/packages/ts/react-form/src/index.ts#L151-L153

### Expected-behavior

Expected the `DateTimePicker` to be marked as visited also on `change` so that validation for the `BinderNode` runs.

### Reproduction

Place the following code in React docs:

```tsx
import { reactExample } from 'Frontend/demo/react-example'; // hidden-source-line
import React, { useEffect } from 'react';
import { DateTimePicker } from '@vaadin/react-components/DateTimePicker.js';
import { useForm, useFormPart } from '@vaadin/hilla-react-form';
import AppointmentModel from 'Frontend/generated/com/vaadin/demo/domain/AppointmentModel';

function Example() {
  const { model, field } = useForm(AppointmentModel);
  const dateTimeField = useFormPart(model.startDateTime);

  useEffect(() => {
    dateTimeField.addValidator({
      message: 'The selected day of week or time is not available',
      validate: (dateTime: string) => {
        const date = new Date(dateTime);
        const validWeekDay = date.getDay() >= 1 && date.getDay() <= 5;

        const time = dateTime.split('T')[1];
        const validTime =
          (time >= '08:00' && time <= '12:00') || (time >= '13:00' && time <= '16:00');

        return validWeekDay && validTime;
      },
    });
  }, []);

  return (
    // tag::snippet[]
    <DateTimePicker
      label="Appointment date and time"
      helperText="Open Mondays-Fridays, 8:00-12:00, 13:00-16:00"
      step={60 * 30}
      {...field(model.startDateTime)}
    />
    // end::snippet[]
  );
}

export default reactExample(Example); // hidden-source-line
```

### System Info

Vaadin 24.4 (latest) and earlier versions.

web-padawan:
## Description

Fixes #2134

## Type of change

- Bugfix

Note: I'm not very happy with my approach to test this (like, I'd prefer to spy on validator). Suggestions are welcome.

Also, ideally we wouldn't need to use `fireEvent` manually, as native `<input type="number">` fires it on <kbd>Enter</kbd>.
But for some reason `user.type(projectInput, '123[Enter]')` didn't work this way, probably RTL doesn't handle this case.

### DIFF
--- a/packages/ts/react-form/test/useForm.spec.tsx
+++ b/packages/ts/react-form/test/useForm.spec.tsx
@@ -350,7 +350,8 @@ describe('@vaadin/hilla-react-form', () => {
       await user.type(projectInput, '123');
       // Mimic change that happens without blur
       fireEvent.change(projectInput);
-      await expect(findByTestId('count')).to.eventually.have.property('textContent', '1');
+      const count = await findByTestId('count');
+      expect(count).to.have.text('1');
     });
   });
 });
--- a/packages/ts/react-form/test/useForm.spec.tsx
+++ b/packages/ts/react-form/test/useForm.spec.tsx
@@ -1,5 +1,5 @@
 import { expect, use } from '@esm-bundle/chai';
-import { act, render, type RenderResult, waitFor } from '@testing-library/react';
+import { act, fireEvent, render, type RenderResult, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import chaiAsPromised from 'chai-as-promised';
 import chaiDom from 'chai-dom';
@@ -320,6 +320,37 @@ describe('@vaadin/hilla-react-form', () => {
       const loadFormBtn = await findByTestId('load');
       await user.click(loadFormBtn);
       await expect(findByTestId('contracts')).to.eventually.have.nested.property('selectedOptions.0.value', '202');
+    });
+
+    it('runs validators on change event', async () => {
+      function ProjectForm() {
+        const { addValidator, field, model } = useForm(EntityModel);
+        const [count, setCount] = useState(0);
+
+        useEffect(() => {
+          addValidator({
+            message: 'ID expected',
+            validate: (entity) => {
+              setCount(count + 1);
+              return !entity.projectId;
+            },
+          });
+        }, []);
+
+        return (
+          <>
+            <input type="number" data-testid="project" {...field(model.projectId)} />
+            <div data-testid="count">{count.toString()}</div>
+          </>
+        );
+      }
+
+      const { findByTestId } = render(<ProjectForm />);
+      const projectInput = await findByTestId('project');
+      await user.type(projectInput, '123');
+      // Mimic change that happens without blur
+      fireEvent.change(projectInput);
+      await expect(findByTestId('count')).to.eventually.have.property('textContent', '1');
     });
   });
 });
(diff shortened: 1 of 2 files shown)


packages/ts/react-form/test/useForm.spec.tsx
```diff
@@ -321,5 +321,36 @@ describe('@vaadin/hilla-react-form', () => {
...
+          </>
+        );
+      }
+
+      const { findByTestId } = render(<ProjectForm />);
+      const projectInput = await findByTestId('project');
+      await user.type(projectInput, '123');
+      // Mimic change that happens without blur
+      fireEvent.change(projectInput);
+      await expect(findByTestId('count')).to.eventually.have.property('textContent', '1');
```
Lodin, line 353: As far as I remember, you can shrink it to just (maybe some Chai plugin is required).
```suggestion
      await expect(findByTestId('count')).to.eventually.have.text('1');
```

Lodin (approved):
LGTM

packages/ts/react-form/test/useForm.spec.tsx
```diff
@@ -321,5 +321,36 @@ describe('@vaadin/hilla-react-form', () => {
...
+            <div data-testid="count">{count.toString()}</div>
+          </>
+        );
+      }
+
+      const { findByTestId } = render(<ProjectForm />);
+      const projectInput = await findByTestId('project');
+      await user.type(projectInput, '123');
+      // Mimic change that happens without blur
+      fireEvent.change(projectInput);
```
sissbruecker, line 352: Was looking into this a bit. I guess it doesn't work with `user.type` because testing library is just calling `dispatchEvent` for the respective keyboard events, it doesn't integrate with the browser driver to natively press keys. As such it probably can't trigger the change event of a native input.

packages/ts/react-form/test/useForm.spec.tsx
```diff
@@ -321,5 +321,36 @@ describe('@vaadin/hilla-react-form', () => {
...
+      function ProjectForm() {
+        const { addValidator, field, model } = useForm(EntityModel);
+        const [count, setCount] = useState(0);
+
+        useEffect(() => {
+          addValidator({
+            message: 'ID expected',
+            validate: (entity) => {
+              setCount(count + 1);
+              return !entity.projectId;
```
sissbruecker, line 335: Regarding spy usage: It's possible to define a spy outside of the component and call the spy here. That requires adding a timeout before the assertion though, it looks like there is some async processing in the binder. Not sure if that makes things easier.

sissbruecker (approved):

packages/ts/react-form/test/useForm.spec.tsx
web-padawan: Thanks. I guess we can keep the current approach as it's not very complicated anyway.

PR #2153, opened 2024-02-29
chore: update components 24.4.0-alpha15

platosha:


Artur- (commented):
react-grid-test and react-signals also use react-components

Artur-:
Maybe they are updated by the build?

platosha:
> Maybe they are updated by the build?

yes, IT test modules will be updated by the snapshot build's PR.

taefi (approved):

PR #2170, opened 2024-03-05
test: disable FS router in react ITs temporarily

taefi:
FS Router should not automatically be enabled in projects with the existing `routes.tsx` that routes are defined manually. This is just a workaround to disable the FS Router processing the views dir for the sake of ITs to pass.


Lodin (approved):

taefi:
Closing in favor of #2171

PR #2177, opened 2024-03-06
refactor(file-router): remove or extract non-public API elements

### LINKED ISSUE #2156: ViewConfig.register has no impact
### Describe the bug

The type definition for `ViewConfig` has a `register` property that is documented to prevent the view from being automatically registered with the router. It is, however, possible to navigate to a view even if `register` has been set to `false`.

### Expected-behavior

Expected that documented properties work. Should either make the property work or remove it from the type definition if it's out of scope.

### Reproduction

1. https://github.com/Lodin/hilla-file-router-example/commit/ec8cd0b6f580f052292082ea598702c49f30d0df
2. Add `export const config : ViewConfig = { register: false }` to `hilla.tsx`
3. Navigate to `path/to/hilla` and note that the view is rendered there

### System Info

Vaadin 24.4.0.alpha7

### LINKED ISSUE #2157: Redundant params property in ViewConfig
### Describe the bug

The `ViewConfig` type has a `params` object that seems to be used internally even though I doesn't make sense for the application developer to define a value for it when configuring a view.

### Expected-behavior

There should be a type that contains only the properties that the application developer is expected to define in an `export const config` definition. If additional properties are introduced when processing the configuration for a view, then another type should be used for that purpose instead.

### Reproduction

N/A

### System Info

Vaadin 24.4.0.alpha7

Lodin:
Resolves #2157.
Resolves #2156.

### DIFF
--- a/packages/ts/hilla-file-router/package.json
+++ b/packages/ts/hilla-file-router/package.json
@@ -23,7 +23,7 @@
     "build": "concurrently npm:build:*",
     "build:esbuild": "tsx ../../../scripts/build.ts",
     "build:dts": "tsc --isolatedModules -p tsconfig.build.json",
-    "build:copy": "cd src && copyfiles **/*.d.ts ..",
+    "build:copy": "cd src && copyfiles *.d.ts **/*.d.ts ..",
     "lint": "eslint src test",
     "lint:fix": "eslint src test --fix",
     "test": "mocha test/**/*.spec.ts --config ../../../.mocharc.cjs",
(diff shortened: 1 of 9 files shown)


taefi (commented):
LGTM. We may consider the followings:
- name of `ServerViewConfig` might be confusing for developers e.g. to assume they can do something about server-side views.
- exporting the internal type `ServiceViewConfig` in the `types.d.ts` along with other public types might be giving some wrong impressions about it to the users. We may consider moving it to another file such as `internal-types.ts`, makes it hidden for future possible internal changes as well.

Lodin:
@taefi My idea here was that it exists in the `Vaadin` object, so it is anyway exports to the public as a part of the global definition. What do you think about that?

taefi (approved):

PR #2181, opened 2024-03-06
feat: enable maven enforce-plugin-versions rule

taefi:
To prevent having old plugins accidentally
resolved by maven during the build. For
example, after renaming groupIds to
com.vaadin there was still hilla-maven-plugin
from `com.vaadin.hilla` being used during the
builds, and maven was resolving the last
existing version of the plugin instead of
only one version in use for the whole project.

NOTE: Enabling enforce-plugin-versions is not enough.


taefi:
Closing as of now just enabling enforce-plugin-versions is not enough.

PR #2184, opened 2024-03-06
test: add flow-react dependency to ITs

platosha:
Depends on #2183

Due to `flow-react` opt-out support feature (`FrontendUtils.isReactModuleAvailable()` checks, for instance, [in `TaskUpdatePackages`](https://github.com/vaadin/flow/blob/d610e3f1d777a72954b33a3b5bcc404806d643ef/flow-server/src/main/java/com/vaadin/flow/server/frontend/TaskUpdatePackages.java#L208-L212)), React components aren't installed if the `flow-react` dependency is missing.

This change adds `flow-react` to IT test modules, making Flow tasks install / update `@vaadin/react-components` where necessary.


packages/java/tests/pom.xml
```diff
@@ -158,6 +163,9 @@
                     <groupId>com.vaadin</groupId>
                     <artifactId>hilla-maven-plugin</artifactId>
                     <version>${project.version}</version>
+                    <configuration>
+                        <generateBundle>false</generateBundle>
```
platosha, line 167: While technically not required, this reduces the noise in the untracked files caused by creating the `prod.bundle` in ITs.

packages/java/tests/pom.xml
```diff
@@ -78,6 +78,11 @@
             <artifactId>vaadin-dev-server</artifactId>
             <version>${flow.version}</version>
         </dependency>
+        <dependency>
```
platosha, line 81:
```suggestion
        <!--
          Due to flow-react opt-out support feature
          `FrontendUtils.isReactModuleAvailable()` checks, for instance,
          `TaskUpdatePackages`, React components aren't installed
          if the flow-react dependency is missing.
        -->
        <dependency>
```

cromoteca (dismissed):

Issue #2104, opened 2024-02-22
[i18n] Prevent app rendering while loading initial translations


sissbruecker:
It should be possible to prevent the app from rendering before the initial set of translations have been loaded, either by delaying the rendering completely, or by enabling developers to show a fallback UI (loading indicator / skeleton).

This prototype shows some possible approaches: https://github.com/sissbruecker/hilla-i18n-client-prototype

```tsx
import { i18n } from "@vaadin/i18n";

// Block app rendering with top-level await
await i18n.configure();

// Or render fallback using initialized signal
i18n.configure();

function App() {
  return i18n.initialized.value 
    ? <RouterProvider />
    : <Loading />
}
```

PR #2186, opened 2024-03-07
feat: allow checking for i18n initialized state

sissbruecker:
Allows checking whether the I18n instance has been initialized or not, for example to show a placeholder component or loading indicator while initial translations are loading.

Fixes https://github.com/vaadin/hilla/issues/2104


packages/ts/react-i18n/test/i18n.spec.tsx
```diff
@@ -176,6 +182,21 @@ describe('@vaadin/hilla-react-i18n', () => {
         await i18n.setLanguage('de-DE');
         expect(effectSpy.calledOnceWith('de-DE', 'Stadt')).to.be.true;
       });
+
+      it('should run effects when initialized changes', async () => {
```
ugur-vaadin, line 186: Should we also check for not running effects when reconfigured after initializing?

sissbruecker: We haven't really defined what should happen if developers call `configure` multiple times, or if that should even be allowed. Besides, re-running effects doesn't seem problematic on its own?

ugur-vaadin (approved):

sissbruecker:
@platosha Sure, done

platosha (approved):

PR #2200, opened 2024-03-11
chore(java): Remove unnecessary version declarations from pom.xml files

cromoteca:


packages/java/parser-jvm-utils/pom.xml
```diff
@@ -19,11 +19,6 @@
     </properties>
 
     <dependencies>
-        <dependency>
```
cromoteca, line 22: This is a duplicate declaration

platosha (approved):
Thank you! Nice to have some cleanup there.

PR #2207, opened 2024-03-12
refactor(file-router): make views.json data tree-like

### LINKED ISSUE #2192: [FS router] inheritance of layout access checks
For layouts with restricted access, accessing any child view should be checked against the parent rules. It would be unexpected if a view is able to override access restrictions of the parent layout and render it.

Make sure the server and client side parts have aligned behaviour regarding inheriting restrictions.

Lodin:
Fixes #2192


packages/java/endpoint/src/main/java/com/vaadin/hilla/route/records/ClientViewConfig.java
```diff
@@ -42,25 +50,103 @@ public void add(String key, Object value) {
...
 
+    public String getTitle() {
+        return title;
+    }
+
+    public String[] getRolesAllowed() {
+        return rolesAllowed;
+    }
+
+    public boolean isRequiresLogin() {
```
taefi, line 69: As these are going to be part of the public API, I suggest to rename `requiresLogin` to `loginRequired` so that the accessor could be read naturally as `isLoginRequired`.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/records/ClientViewConfig.java
```diff
@@ -42,25 +50,103 @@ public void add(String key, Object value) {
...
+
+    public String getRoute() {
+        return route;
+    }
+
+    public boolean isLazy() {
+        return lazy;
+    }
+
+    public boolean isRegister() {
```
taefi, line 81: Same as previous comment, `register` is not informative enough about what is it exactly doing/affecting, if it's enabling the users to override the default automatic registration of views, I'd suggest renaming it to `autoRegistered` (or `automaticallyRegistered` to be more verbose) which leads to having something like `isAutoRegistered` instead of `isRegister` as the accessor.

platosha (approved):

PR #2212, opened 2024-03-13
feat: add react-i18n package to Hilla React dependencies

platosha:


cromoteca (approved):

Legioth:
This feature was originally intended to be introduced in 24.5 alpha but it's not going into 24.4 since that version is late. I wonder if we should make sure we remember to revert this change from the 24.4 branch once it's created?

Issue #2248, opened 2024-03-21
File System router generates router config for JS/TS files


mshabarov:
### Describe the bug

FileSystem-based router takes into account any `.ts`, `.js` files in `frontend/views/`, for example, Lit templates or any JS/TS scripts and included them into React Router meta data for routing config.

This breaks Flow+Lit templates or Flow+JS/TS add-on if the files are in `views`, as the Hilla FS plugin generates non-empty routing config and this crashes the React Router, as these source aren't React component in reality.

Hilla's FS router should take into account only `.tsx` and `.jsx` files that usually refer to React components by convention.

Workaround is to set `reactEnabled=false`, that prevents the plugin to run.

### Expected-behavior

Projects should start normally without router crash if the app has non-route files in `views`.

### Reproduction

Run any Flow apps with Lit templates in `views`, or even a simple JS file with `console.log('hello')` inside.

### System Info

Vaadin 24.4.0.alpha15 and alpha16.

PR #2249, opened 2024-03-21
fix: Ignore non-JSX files in views folder

mshabarov:
Removes .ts and .js from extension list, making them ignored.
Shows a warning message if sees any .ts or .js files in /views.
Changes test files extensions to .tsx and .jsx and adds a couple of .js and .ts files, which shouldn't be processes by the plugin.

Fixes https://github.com/vaadin/hilla/issues/2248


packages/ts/hilla-file-router/src/vite-plugin/collectRoutesFromFS.ts
```diff
@@ -97,6 +97,10 @@ export default async function collectRoutesFromFS(
           children: [],
         });
       }
+    } else {
+      logger.warn(
```
Legioth, line 101: Should not log for files prefixed with `_` regardless of the extension. I'm also not sure if we should log the warning also for e.g. .css files inside `views/`

mshabarov: Changed to show warning only for js and ts.

Legioth: Based on how I read the code, it would still log a warning for `_foo.js` and that's not right since `_` files are supposed to be ignored regardless of the extension

mshabarov: Ah, I forgot to commit this. Thanks, added.

mshabarov:
Tested with the attached project in https://github.com/vaadin/copilot-internal/issues/1197 and the app works as expected now.

platosha (approved):

PR #2258, opened 2024-03-25
fix(fs-router) check for routes with / first, then without

### LINKED ISSUE #2244: (file router) the view for the root path is never found in Java view registry
### Describe the bug

Looks like there is a mismatch between the root path `"/"` and the root view `""`, and the `AntPathMatcher` is unable to make the match happen.

image

The root always redirects to login page.

### Expected-behavior

The root URL should redirect to the login page only if explicitly protected.

### Reproduction

Use
```java
http.authorizeHttpRequests(authorize -> authorize.requestMatchers(routeUtil::isRouteAllowed).permitAll());
```

### System Info

Hilla 24.4-SNAPSHOT (as of 21/03)

cromoteca:
Fixes #2244

### DIFF
--- a/packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
+++ b/packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
@@ -16,6 +16,7 @@
 package com.vaadin.hilla.route;
 
 import com.vaadin.hilla.route.records.ClientViewConfig;
+
 import org.springframework.stereotype.Component;
 import org.springframework.util.AntPathMatcher;
 
@@ -23,6 +24,8 @@ import java.io.Serializable;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.Objects;
+import java.util.stream.Stream;
 
 /**
  * Keeps track of registered client side routes.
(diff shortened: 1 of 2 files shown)


packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
```diff
@@ -84,11 +87,34 @@ public void removeRoute(String route) {
...
+            return path + "/";
+        }
+        return path;
+    }
+
+    private String removeTrailingSlash(String path) {
+        if (path.isEmpty()) {
+            return "";
+        }
+        if (path.charAt(path.length() - 1) == '/') {
```
taefi, line 115: Any specific reason for not using
```suggestion
        if (path.endsWith("/")) {
```

taefi (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
```diff
@@ -84,11 +87,34 @@ public void removeRoute(String route) {
...
+                    }
+                    return null;
+                }).filter(Objects::nonNull).findFirst().orElse(null);
+    }
+
+    private String addTrailingSlash(String path) {
+        if (path.isEmpty()) {
+            return "/";
+        }
+        if (path.charAt(path.length() - 1) != '/') {
```
taefi, line 105: Same comment about using `String#endsWith`

taefi (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
```diff
@@ -84,11 +87,34 @@ public void removeRoute(String route) {
...
+        return Stream.of(addTrailingSlash(path), removeTrailingSlash(path))
+                .map(p -> {
+                    for (String route : routes) {
+                        if (pathMatcher.match(route, p)) {
+                            return registeredRoutes.get(route);
+                        }
+                    }
+                    return null;
+                }).filter(Objects::nonNull).findFirst().orElse(null);
+    }
```
knoobie, line 99: Just bypassing here: Did you consider security here? I remember some CVEs in the recent past which originated from security confusion based on trailing slash.. 

Additionally there is this recent Spring Framework Change to keep in mind: https://github.com/spring-projects/spring-framework/issues/28552

cromoteca: This code doesn't actually change which content is rendered. Its purpose is to try to give the same view permission response that you would get on the client where the navigation is handled by the React Routed.

By doing some manual testing, I've seen that when two routes are present which are only different on the trailing slash, React Router always renders the one with the slash (e.g. you cannot see `/about` if `/about/` exists). But it is also forgiving: `/about/` will show the view for `/about` if an exact match is not there.

We don't have this information here, since the new File Router only stores the path items (e.g. `["about"]`). Also, the root path is stored as `""`, but it must match `"/"`.

This PR modifies the code that tries to find the route that React Router would render given a path. In that sense, it should not open any additional security breach.

taefi (approved):

PR #2259, opened 2024-03-25
feat(fs-router) add console errors for duplicated paths

### LINKED ISSUE #2245: (file router) conflicts between files and directories are not handled correctly
The current file router behavior in case of conflict between `/about.tsx` and `/about/$index.tsx` is definitely confusing: I see the title from the former in the menu and the contents of the latter when clicking.

cromoteca:
Console messages are generated for duplicate paths in the filesystem, like `about.tsx` and `about/$index.tsx`.

Closes #2245

### DIFF
--- a/packages/ts/hilla-file-router/src/vite-plugin/createRoutesFromMeta.ts
+++ b/packages/ts/hilla-file-router/src/vite-plugin/createRoutesFromMeta.ts
@@ -76,11 +76,19 @@ export default function createRoutesFromMeta(views: RouteMeta, { code: codeFile 
       ([statement]) => statement as ts.ImportDeclaration,
     ),
   ];
+  const errors: string[] = [];
   let id = 0;
 
   const routes = transformTreeSync<RouteMeta, CallExpression>(
     views,
-    (view) => view.children.values(),
+    (view) => {
+      const paths = view.children.map((c) => c.path);
+      const uniquePaths = new Set(paths);
+      paths
+        .filter((p) => !uniquePaths.delete(p))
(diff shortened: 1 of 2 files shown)


taefi (approved):

packages/ts/file-router/src/vite-plugin/createRoutesFromMeta.ts
```diff
@@ -76,11 +76,19 @@ export default function createRoutesFromMeta(views: RouteMeta, { code: codeFile
...
 
   const routes = transformTreeSync<RouteMeta, CallExpression>(
     views,
-    (view) => view.children.values(),
+    (view) => {
+      const paths = view.children.map((c) => c.path);
+      const uniquePaths = new Set(paths);
+      paths
+        .filter((p) => !uniquePaths.delete(p))
+        .forEach((p) => errors.push(`console.error("Two views share the same path: ${p}");`));
```
taefi, line 89: Would it possible to change the error message to include the name of all the views (not just "Two") sharing the same path?

taefi (changes requested):

packages/ts/file-router/src/vite-plugin/createRoutesFromMeta.ts
cromoteca: To my understanding, the concerned views cannot be more than two. And for the full path, the current object has no information about parents, so the only thing we can easily access is the current path item (e.g. "about").

taefi (approved):

PR #2279, opened 2024-03-31
feat: add new attribute 'hiddenColumns' to configure which columns sh…

### LINKED ISSUE #1507: [autogrid][dx] Consider providing a way for listing only the columns that should be hidden
### Describe your motivation

As the default for `AutoGrid` is to show all the columns (there could be 30-40 columns), using `visibleColumns` to hide a handful of columns is not convenient at all. 

Something like `hiddenColumns` that only gets the list of columns to be hidden is more convenient.
In this case, the `visibleColumns` can also be named as something like `columnOrder` since that's the only usage for it after defining a `hiddenColumns` property.

_No response_

### Additional context

_No response_

rbrki07:
…ould hide in autogrid

## Description

Adding a new attribute `hiddenColumns` for AutoGrid component allowing to hide certain columns in the grid. The new attribute will be applied after `visibleColumns` and `customColumns` have been processed. This ensures a high flexibility when combining this attributes without being mutually exclusive. 

Fixes #1507

## Type of change

- [ ] Bugfix
- [X] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [X] The issue is created in the corresponding repository and I have referenced it.

- [X] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.


cromoteca:
Hi @rbrki07,

First of all thank you for your contribution. I tested your PR locally and it works as expected. In that sense, this PR could be merged right away. I think that being able to simply exclude some columns, without redefining the whole list of visible columns, is a great addition.

I see you were in a discussion with @taefi on the feature details. There are some possible improvements I can see:

- forbid defining both `hiddenColumns` and `visibleColumns`; this is possible if we change interfaces to types where needed, for example https://github.com/vaadin/hilla/blob/0961d0bb9b527ab8d27402b32cfe924501b70cc3/packages/ts/react-crud/src/autogrid.tsx#L44-L147
- as the new property will be also available in the CRUD (in `gridProps`), users could expect a similar property in the form, just like `visibleColumns` has its counterpart in `visibleFields`

It is not necessary to make those changes in this PR, we can follow up with another one, but I wanted to hear your opinion about those points.

rbrki07:
Hi @cromoteca,
thank you for reviewing my pr and thank you for your feedback. Regarding the possible improvements you mentioned:
> forbid defining both `hiddenColumns` and `visibleColumns`;

I don't think this is necessary, but I guess my opinion on this is biased. As a developer, who is using Hilla the solution proposed with this pr would be sufficient for me. As a developer, who is maintaining Hilla I guess there other valid opinions.

> as the new property will be also available in the CRUD (in gridProps), users could expect a similar property in the form, just like `visibleColumns` has its counterpart in `visibleFields`

I think this is a valid point and a very good idea. I would prefer to handle this in a follow-up issue/pr.

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -57,7 +57,8 @@ interface AutoGridOwnProps<TItem> {
    *
    * By default, the grid shows columns for all properties of the model which
    * have a type that is supported. Use the `visibleColumns` option to customize
-   * which columns to show and in which order.
+   * which columns to show and in which order. Use the `hiddenColumns` option to
+   * hide certain columns and to keep the default order.
```
cromoteca, line 61: I'd avoid sponsoring a parameter in another parameter's comment.

packages/ts/react-crud/test/autogrid.spec.tsx
```diff
@@ -1103,6 +1103,51 @@ describe('@vaadin/hilla-react-crud', () => {
         await assertColumns(grid, 'email', 'firstName');
       });
 
+      it('should hide configured columns and keep default order', async () => {
```
cromoteca, line 1106:
```suggestion
      it('should hide columns and keep default order', async () => {
```

cromoteca (changes requested):

cromoteca (approved):

packages/ts/react-crud/src/autogrid.tsx
```diff
@@ -258,6 +265,17 @@ function useColumns(
...
 
+  // When using `hiddenColumns` option, remove columns to hide using their `key`
+  if (options.hiddenColumns) {
+    columns = columns.filter((column) => {
+      const { key } = column;
+      if (key) {
+        return !options.hiddenColumns?.includes(key);
+      }
+      return true;
+    });
```
cromoteca, line 276: Line 275 is not covered by tests. Now, the case of a null key is not a real one, since all columns are supposed to have one. Yet, that makes me think of a shorter syntax for this filter:
```suggestion
    columns = columns.filter(({ key }) => !(key && options.hiddenColumns?.includes(key)));
```

cromoteca (changes requested):
I have a new change suggestion. And if possible avoid force-pushing to ease further review. Commits will be squashed anyway when merging to `main`. Thanks.

packages/ts/react-crud/src/autogrid.tsx
rbrki07: I also noticed the missing test coverage too and added an additional test `should ignore unknown columns when using hiddenColumns` with https://github.com/vaadin/hilla/compare/ce783d2e6ad87a7c921edfe733c9d89ca6396369..e89d85722221ca1641fb841447a1f4dd9830bb85. Could you rerun the code coverage test? The codecov report is based on an older commit.

cromoteca: The report seems to be up-to-date. The `key` is related to `columns` here, while your test adds unknown column names to `hiddenColumns`. But, again, all columns are supposed to have a key, even custom ones, see https://github.com/rbrki07/hilla/blob/376e8f2017ae60865b8561cf4b61e26633d9a9a6/packages/ts/react-crud/test/autogrid.spec.tsx#L1310 for example.

rbrki07: Thank you for clarification on this. I updated the pr with you suggestion.

cromoteca (approved):

rbrki07:
Thank you @cromoteca for your guidance and support in my first pull request for Vaadin.

PR #2297, opened 2024-04-04
fix: ignore change/addition in empty view/layout/index files

### LINKED ISSUE #2204: Server is redeployed when an empty $index.tsx is created inside views/
### Describe the bug

A redeploy is triggered immediately when creating a new empty `$index.tsx` file inside `frontend/views/` (either as in the root or in a sub directory). It seems like a new empty entry is added to `views.json` even though that entry doesn't seem to serve any purpose.

### Expected-behavior

Expected that a redeploy would happen only when some actual content is saved in the file since the view is functional only at that point. More specifically, I would expect that an `$index.tsx` file would be handled in the same way as an empty file with some other name.

### Reproduction

1. Create a `foo` sub directory inside `frontend/views`. No redeploy
2. Create an empty `$index.tsx` file inside `foo`. This causes a redeploy even though the file is empty.

### System Info

Vaadin 24.0.0.alpha11. File created through VS Code.

taefi:
## Description

FS Router's vite-plugin watches for the change, addition, or removal of the files under `views` directory, and determines the FS router configurations based on the updates. When a new file is added anywhere under views it can be empty or doesn't export any components or route configurations. In this case, that file should be ignored.

Fixes #2204

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


packages/ts/file-router/src/vite-plugin/collectRoutesFromFS.ts
```diff
@@ -33,10 +33,11 @@ export type CollectRoutesOptions = Readonly<{
 async function checkFile(url: URL | undefined, logger: Logger): Promise<URL | undefined> {
   if (url) {
     const contents = await readFile(url, 'utf-8');
-    if (contents === '') {
+    if (contents === '' || contents.trim() === '') {
```
Lodin, line 36: Using `trim` is enough here
```suggestion
    if (contents.trim() === '') {
```

packages/ts/file-router/src/vite-plugin/collectRoutesFromFS.ts
```diff
@@ -74,6 +75,10 @@ export default async function collectRoutesFromFS(
       children.push(await collectRoutesFromFS(new URL(`${d.name}/`, dir), { extensions, logger, parent: dir }));
     } else if (d.isFile() && extensions.includes(extname(d.name))) {
       const file = new URL(d.name, dir);
+      const url = await checkFile(file, logger);
```
Lodin, line 78: That's definitely not the best place to check files because there is a lot of files we're skipping. That's why I was checking the file in the array after that.

packages/ts/file-router/src/vite-plugin/collectRoutesFromFS.ts
```diff
@@ -106,16 +111,11 @@ export default async function collectRoutesFromFS(
...
-        let { file: f, layout: l } = child;
-        [f, l] = await Promise.all([checkFile(f, logger), checkFile(l, logger)]);
-
-        return {
-          ...child,
-          file: f,
-          layout: l,
-        };
-      }),
+      children.map(async (child) => ({
```
Lodin, line 114: What if instead of removing the filecheck here, we filter the array according to the results of `checkFile`?

Lodin (changes requested):

packages/ts/file-router/src/vite-plugin/collectRoutesFromFS.ts
taefi: > That's definitely not the best place to check files because there is a lot of files we're skipping. That's why I was checking the file in the array after that.

Could you please elaborate more about the files we are skipping if we do the checking here?
To my understanding, this branch of `else if` is the one adding the entires to the `children` and we only care about those files with the specified extensions, so I couldn't understand what are the skipped files, and why is it important to first add some entires to `children` for empty files (or those that are not exporting a component) and then filter them later. 
A brief description of the rule in my mind: 
Files with `.tsx` or `.jsx` extension that are empty or doesn't have a export default (yet) can be considered non-existent from the FS Router POV. No need to consider their existence in the first place.

packages/ts/file-router/src/vite-plugin/collectRoutesFromFS.ts
taefi: My reason for removing the `checkFile` call for the `file` was due to the fact that it's done earlier before adding it to the `children` entries. In case we want to add filtering to the pipeline before the map, we still don't need to call the `checkFile` on `file` as it is done earlier, unless we don't check the `file` earlier.

packages/ts/file-router/src/vite-plugin/collectRoutesFromFS.ts
taefi: Done.

packages/ts/file-router/src/vite-plugin/collectRoutesFromFS.ts
Lodin: The idea behind my comment was that we're skipping the files that start with `_`. However, in your implementation, they are also checked. However, each check is IO operation which is expensive. That's why I perform the check only with already added files.

Also, for me, it looks more logically correct when all checks are done in the same place somehow.

Lodin (approved):
Ok, let's go with this implementation for now

PR #2299, opened 2024-04-05
fix(form): force calling `clear` at form creation/model change

### LINKED ISSUE #2282: React form not showing required indicators before first interaction
### Describe the bug

I upgraded my project to 24.4.0.alpha18. My form doesn't display required indicators for fields before I interact with the form once. 

Before interacting:
Screenshot 2024-04-02 at 13 41 29

After focusing and blurring the first field:
Screenshot 2024-04-02 at 13 41 36

### Expected-behavior

That required indicators are shown immediately

### Reproduction

Source here https://github.com/vaadin/expo-hilla-react/blob/main/src/main/frontend/views/form.tsx

### System Info

Vaadin 24.4.0.alpha18

cromoteca:
Calling `clear` at form creation allow the field validation strategies to be in place. This does not seem to have any negative effect.

Fixes #2282.


platosha:
Could we have a test here to prevent future regressions?

cromoteca:
I added a simple test and verified that the test fails without the fix.

platosha (approved):

PR #2309, opened 2024-04-09
fix: use order index and paths comparison to sort menu items

### LINKED ISSUE #2222: Default menu order puts ordered items after unordered items
### Describe the bug

It seems like the default value used if `menu.order` is not defined in `export const config` is 1 which means that setting the order to 1 for a view causes it to end up after views without an explicitly defined menu order rather than before.

As a workaround, you can use negative numbers for items that should be placed before items without an explicit order.

### Expected-behavior

Views with an explicitly defined order should be shown before other views in the menu. In practice, this means that the default `order` value should be `Number.MAX_VALUE` rather than `0`. 

### Reproduction

1. https://github.com/Lodin/hilla-file-router-example.
3. Add `export const config : ViewConfig = { menu: { order: 1 } }` to the view that is in the middle in the menu.
4. Notice that this makes the menu entry for that view show up last in the menu rather than first

### System Info

Vaadin 24.4.0.alpha14

taefi:
## Description
This will fix the menu item orders by interpreting ViewConfig's menu order as the natural index, so that menu items with lower order will show up on top menu items with higher order. This also will adjust sort behavior in a way that views without any defined order would be show after the menu items with defined order. Ties are resolved based on the `path` string natural comparison.

Fixes #2222

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


packages/ts/file-router/src/runtime/createMenuItems.ts
```diff
@@ -30,7 +30,13 @@ export function createMenuItems(vaadinObject = (window as VaadinWindow).Vaadin):
             to: _path,
             icon: config.menu?.icon,
             title: config.menu?.title ?? config.title,
+            order: config.menu?.order,
           };
         })
+        // Sort views according to the order specified in the view configuration.
+        .sort((menuA, menuB) => {
+          const ordersDiff = (menuA.order ?? Number.MAX_VALUE) - (menuB.order ?? Number.MAX_VALUE);
+          return ordersDiff !== 0 ? ordersDiff : menuA.to.localeCompare(menuB.to);
```
Lodin, line 39: I guess, you have to use `Intl.Collator` here as it is the most correct way to do the locale comparison nowdays.

Lodin (approved):

PR #2316, opened 2024-04-11
feat: utilize access control when populating client-side routes in menu

taefi:
This enables filtering of Hilla views based on the security configurations exported from views, and uses the same access control API that could be optionally used Spring Security configuration for http level security.

It is important to note that filtering Hilla views in the menu is a usability feature not a security one.

### DIFF
--- a/packages/java/endpoint/src/main/java/com/vaadin/hilla/startup/RouteUnifyingServiceInitListener.java
+++ b/packages/java/endpoint/src/main/java/com/vaadin/hilla/startup/RouteUnifyingServiceInitListener.java
@@ -21,6 +21,7 @@ import com.vaadin.flow.server.VaadinServiceInitListener;
 import com.vaadin.hilla.route.ClientRouteRegistry;
 import com.vaadin.hilla.route.RouteUnifyingIndexHtmlRequestListener;
 import com.vaadin.hilla.route.RouteUnifyingConfigurationProperties;
+import com.vaadin.hilla.route.RouteUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -39,6 +40,8 @@ public class RouteUnifyingServiceInitListener
 
     private final ClientRouteRegistry clientRouteRegistry;
 
+    private final RouteUtil routeUtil;
+
     private final RouteUnifyingConfigurationProperties routeUnifyingConfigurationProperties;
 
     /**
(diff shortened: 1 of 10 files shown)


Legioth:
I would suggest having the filtering enabled by default with no way of customizing it for now (other than the existing `application.properties` option to completely disables the feature). The reason for this is that we want to introduce a filtering API that gives unified control over both Flow views and Hilla views rather than having solutions specific to each framework.

taefi:
> I would suggest having the filtering enabled by default with no way of customizing it for now (other than the existing `application.properties` option to completely disables the feature). The reason for this is that we want to introduce a filtering API that gives unified control over both Flow views and Hilla views rather than having solutions specific to each framework.

I think what's done in this PR is _enabling the filtering by default_ but only for the client-side views. So is the requirement to to enabled it for the server-side views as well?

Legioth:
> I think what's done in this PR is enabling the filtering by default but only for the client-side views. So is the requirement to to enabled it for the server-side views as well?

Only client-side views. The reason it seems like it's not enabled by default is that there's a line of code in application code that is needed to enable the filtering.

packages/java/endpoint/src/main/java/com/vaadin/hilla/startup/RouteUnifyingServiceInitListener.java
```diff
@@ -39,21 +40,27 @@ public class RouteUnifyingServiceInitListener
...
+
     private final RouteUnifyingConfigurationProperties routeUnifyingConfigurationProperties;
 
     /**
      * Creates a new instance of the listener.
      *
      * @param clientRouteRegistry
      *            the registry to add the client routes to
+     * @param routeUtil
+     *            the ClientRouteRegistry aware utility for checking if user is
```
cromoteca, line 53: This JavaDoc and the related method signature make me wonder why we need the `RouteUtil` class in the first place.
That's not related to this PR as the class was already there before, but it exposes a single public method which could be a `ClientRouteRegistry` method, and that would simplify passing `RouteUtil` around.
That's just a thought, not necessarily a request.

cromoteca (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/startup/RouteUnifyingServiceInitListener.java
taefi: I think from the cohesion point of view, separating these two classes was a good move, as IMO the `ClientRouteRegistry` should/could remain just a registry and nothing more. Then whoever uses the registry, such as `RouteUtil`, can deal with the content of the registry based on their business.

Though, I think we could at least rename the `RouteUtil` to something more specific, such as `ViewAccessChecker` or `HillaViewAccessChecker` to not get mixed up with the one from Flow.

cromoteca (approved):

PR #2321, opened 2024-04-12
fix(fs-router): pre-populate view titles from filenames

### LINKED ISSUE #2318: [fs-router] default view names are minified in production mode
### Describe the bug

It has been reported that minified view names (e.g. `j1` or `fK`) are shown in the menu in production mode builds.

### Expected-behavior

Default names must still come from the original view name.

### Reproduction

Create some views without a title in the configuration and see the difference between dev mode and prod mode.

### System Info

Hilla 24.4.0.alpha21

cromoteca:
This pull request adds functionality to pre-populate view titles from filenames. It ensures that the view titles are automatically generated based on the filenames, as otherwise doing that in production mode can lead to use minified names.

Fixes #2318.


taefi (dismissed):
LGTM. 
The codecov error seems to be complete non-sense, and it could be ignored.

packages/ts/file-router/src/vite-plugin.ts
```diff
@@ -102,26 +102,39 @@ export default function vitePluginFileSystemRouter({
...
-        nextExports = { ...nextExports, config: currentExports.config };
-      }`,
-          ),
+            if (Object.keys(nextExports).length === 2 && 'default' in nextExports && 'config' in nextExports) {
+              nextExports = { ...nextExports, config: currentExports.config };
+            }`,
+          );
+        } else {
+          // In production mode, the function name is assigned as name to the function itself to avoid minification
+          const functionNames = /export\s+default\s+function\s+(\w+)/u.exec(modifiedCode);
```
Lodin, line 128: Here could be a problem. Sometimes, the user would like to use `const`, so the export could look like this:
```ts
const MyComponent: FC = () => {
  // ...
};

export default MyComponent;
```
In that case, the current RegExp won't work.

There is also a kind of an edge case:
```ts
function MyComponent() {
  // ...
}

export {
  MyComponent as default,
}
```
I don't think if anyone would really use that kind of export, so we can skip it for now.

Lodin (changes requested):

packages/ts/file-router/src/vite-plugin.ts
platosha: Would it be enough to only support the `export default function X` form for now? @Lodin

Lodin: No. I know a lot of devs who like to use `const` (I was one of them for some time as well), and that would be pretty inconvenient and frustrating for them.

Lodin: I guess, this change should be enough to take the additional case into account:
```suggestion
          const functionNames = /export\s+default\s+(?:function\s+)?(\w+)/u.exec(modifiedCode);
```

Lodin: image
It looks like to work as expected

platosha (approved):

PR #2338, opened 2024-04-18
chore: fetch access controlled server routes

tltv:
Use `RouteRegistry#getRegisteredAccessibleMenuRoutes` to get only accessible server Menu routes. Also updated to read Menu information from MenuData instead of directly reading the Menu annotation.

Related-to: vaadin/flow/issues/19127


packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
```diff
@@ -95,7 +107,8 @@ public void modifyIndexHtmlResponse(IndexHtmlResponse response) {
         if (exposeServerRoutesToClient) {
             LOGGER.debug(
                     "Exposing server-side views to the client based on user configuration");
-            availableViews.putAll(collectServerViews());
+            availableViews
+                    .putAll(collectServerViews(response.getVaadinRequest()));
```
taefi, line 111: Earlier we added an application property called `vaadin.expose-server-routes-to-client` with a default value of `false` which is checked here to prevent accidental leakage of server routes to the client, and now that the `populateClientSideMenu` property defined in vaadin/flow#19202, this `exposeServerRoutesToClient` check seems to be a duplicate, and none of the changes in this PR will affect anything until that property and this condition is removed from Hilla codebase.

taefi (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
tltv: Either default can be changed to `true` or `vaadin.expose-server-routes-to-client` property can be removed from Hilla. I'm not sure if it would make sense to still have it (with different name) in Flow side to make up default value for populateClientSideMenu in `DefaultMenuAccessControl`. There's no such system property currently in the other PR.

Legioth: Like discussed in the Flow daily, the decision on whether to send Flow views should be made on the Hilla side. Flow should only make the data available to Hilla in case Hilla wants to use it.

This can very well be in the form of the existing `expose-server-routes-to-client` property (even though we usually have a hierarchy of properties separated by `.` rather than a long phrase separated by `-`). We could just change it so the default in case the property is not explicitly set varies depending on whether Hilla knows that it has a client-side main layout.

taefi: > the decision on whether to send Flow views should be made on the Hilla side

Could you please elaborate more about the "why"?

Legioth: It's based on a technical practicality: Hilla knows whether there is a client-side main layout but Flow doesn't (without making Flow using Hilla APIs which would lead to a circular dependency).

taefi: Thanks. We agreed on the same in https://github.com/vaadin/flow/pull/19202#discussion_r1572233777

packages/java/endpoint/src/main/java/com/vaadin/hilla/startup/RouteUnifyingServiceInitListener.java
```diff
@@ -60,10 +67,14 @@ public class RouteUnifyingServiceInitListener
     @Autowired
     public RouteUnifyingServiceInitListener(
             ClientRouteRegistry clientRouteRegistry, RouteUtil routeUtil,
-            RouteUnifyingConfigurationProperties routeUnifyingConfigurationProperties) {
+            RouteUnifyingConfigurationProperties routeUnifyingConfigurationProperties,
+            @Nullable NavigationAccessControl accessControl,
+            @Nullable ViewAccessChecker viewAccessChecker) {
```
mshabarov, line 72: This was probably discussed already. For non-Spring apps, how the access control `BeforeEnterListener` instances are used and passed to the Menu-related logic, as they aren't auto-registered? 

Well, if the Spring-Boot not used, Hilla isn't used and Hilla main layout isn't there anyway, so maybe it's not that important.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
```diff
@@ -207,4 +253,37 @@ private Map<String, RouteParamType> getRouteParameters(
...
+                    return IOUtils.toString(inputStream, UTF_8);
+                }
+                return "";
+            }
+        }
+        return FileUtils.readFileToString(new File(
+                FrontendUtils.getFrontendGeneratedFolder(
+                        FrontendUtils.getProjectFrontendDir(configuration)),
+                FILE_ROUTES_JSON_NAME), UTF_8);
+    }
```
taefi, line 288: This logic to decide where should we look for `file-routes.json` in different modes reminds me of the `ClientRouteRegistry::getViewsJsonAsResource` (which is currently private). I suggest not to redefine the `FILE_ROUTES_JSON_NAME` and `FILE_ROUTES_JSON_PROD_PATH` duplicate constants and logic here and reuse the method from `ClientRouteRegistry` for getting it as a resource.

taefi: It also makes sense to move the newly introduced method `isClientMenuUsed` to `clientRouteRegistry` and let it be the only place that has to deal with the logic of where to load the `file-routes.json`.

taefi (changes requested):
Apart from the following notes, now that the Flow introduced `@Menu` and the needed APIs for checking what leaks to the client-side menu properly, I wonder, shouldn't this PR also change the default for `vaadin.expose-server-routes-to-client` to `true`?

packages/java/endpoint/src/main/java/com/vaadin/hilla/startup/RouteUnifyingServiceInitListener.java
taefi: I understand the same, when there's no Hilla in the picture.

tltv: Listeners should be registered in `UIInternals`, so that's where they can be read like this:
```
UI.getCurrent().getInternals()
  .getListeners(BeforeEnterHandler.class).stream()
  .filter(handler -> handler instanceof BeforeEnterListener)
  .map(BeforeEnterListener.class::cast).toList()
```

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
```diff
@@ -207,4 +253,37 @@ private Map<String, RouteParamType> getRouteParameters(
...
+
+    boolean isClientMenuUsed(VaadinService vaadinService) {
+        try {
+            String fileRoutesJson = readFileRoutesJsonFile(
+                    ApplicationConfiguration.get(vaadinService.getContext()));
+            var json = mapper.readTree(fileRoutesJson);
+            // as the client registry has no information about layouts, the JSON
+            // is parsed again to search for a root node with children, which
+            // means that we have a root layout.
+            return json.has(0) && json.get(0).has("children")
```
mshabarov, line 265: Can we be confident, that the layout object is always on the first position in the array?

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
tltv: Makes sense. I'll refactor it a bit.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
taefi: Good point. By definition, regarding the automatic menu, we only consider the existence of a `@layout.tsx` (or `.jsx`) directly under the `Frontend/views/` directory (though, `@layout.tsx` can exist in deeper levels, as well). If such a layout exists, it will be the parent of all the other route objects.  Then, shape of the objects in `file-routes.json` is as:
```json
[
    {
        "route": "",
        "params": {},
        "title": "Main Layout",
        "children": [
             {
                  ...
             },
        ]
    }
]
```
or
```json
[
    {
        "route": "",
        "params": {},
        "title": "Main Layout",
        "children": []
    }
]
```
@platosha can say wether there would any possibility about having more than one root-level objects in the array? Or is it defined as `[]` just to support the empty an array (due to other technical considerations)?

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
tltv: Done, also added couple tests.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
```diff
@@ -207,4 +253,37 @@ private Map<String, RouteParamType> getRouteParameters(
...
+                    && json.get(0).get("children").isArray()
+                    && !json.get(0).get("children").isEmpty();
```
taefi: @cromoteca I think we need to also check that the `json.get(0).get("route")` is empty, and also this last check is not needed here since a main layout should satisfy the following criteria:
1. It should have a `"route": ""`, and
2. It should have a `"children": []` or `"children": [...]`

WDYT?

taefi (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
taefi: Now that I look again, for checking the existence of a main layout, I suggest not to read the `file-routes.json` again, as it is done once for when `ClientRouteRegistry` registering the routes. Based on what we already have in `registeredRoutes` (in ClientRouteRegistry) we can determine whether a main layout with empty string route ("") and a non-null children property exists or not.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
platosha: The position of the main layout as the first item in the array isn't guaranteed anymore. One practical case is prepending a `/login` route in the root level before the main layout.

mshabarov: Could we then change the implementation to the following: iterate through the objects in the root array and look for the one that has non-empty children ? If it's found, then we have a layout.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
platosha: Using the `registeredRoutes` map would be indeed more practical, considering that the main layout might not be the first item.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
mshabarov: See https://github.com/vaadin/hilla/pull/2338#discussion_r1576234329

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
```diff
@@ -219,6 +219,24 @@ private void registerAndRecurseChildren(String basePath,
...
+     * </ul>
+     * <li>has an empty route path</li>
+     * <li>has a non-null children list (can be empty)</li>
+     * <li>has no parent</li>
+     *
+     * @return {@code true} if the registry has a main layout, {@code false}
+     *         otherwise
+     */
+    public synchronized boolean hasMainLayout() {
+        return getAllRoutes().values().stream()
```
tltv, line 235: This would not work with a structure like below as route with a parent is not added in the `registeredRoutes` Map. it should probably check parent here too.

```
[{
 "route":""
 "title":"Layout",
 "children":[
	{
	 "route":""
	 "title":"Public"
	}
	]
}]
```

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
```diff
@@ -170,6 +173,48 @@ public void when_developmentMode_then_loadLatestDevModeFileRoutesJsonIfNeeded_lo
         MatcherAssert.assertThat(allRoutes, Matchers.aMapWithSize(12));
     }
 
+    @Test
+    public void when_developmentMode_fileRoutesJsonWithMainLayout_then_hasMainLayout_is_true()
+            throws IOException {
+        mockDevelopmentMode();
+        String fileRoutesJson = "[{ \"route\": \"\", \"children\": []}]";
```
cromoteca:
```suggestion
        String fileRoutesJson = """
                [{
                    "route": "",
                    "children": []
                }]
                """;
```

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
```diff
@@ -170,6 +173,48 @@ public void when_developmentMode_then_loadLatestDevModeFileRoutesJsonIfNeeded_lo
...
+                LocalDateTime.now());
+        Assert.assertTrue("Main layout should be present",
+                clientRouteRegistry.hasMainLayout());
+    }
+
+    @Test
+    public void when_developmentMode_fileRoutesJsonWithoutMainLayout_then_isClientMenuUsed_false()
+            throws IOException {
+        mockDevelopmentMode();
+        String fileRoutesJson = "[{ \"route\": \"some-route\", \"children\": [ {\"route\": \"\", \"params\": {}, \"title\": \"Index\" } ]}]";
```
cromoteca:
```suggestion
        String fileRoutesJson = """
                [{
                    "route": "some-route",
                    "children": [
                        {
                            "route": "",
                            "params": {},
                            "title": "Index"
                        }
                    ]
                }]
                """;
```

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
```diff
@@ -170,6 +173,48 @@ public void when_developmentMode_then_loadLatestDevModeFileRoutesJsonIfNeeded_lo
...
+                LocalDateTime.now());
+        Assert.assertFalse("Main layout should not be present",
+                clientRouteRegistry.hasMainLayout());
+    }
+
+    @Test
+    public void when_developmentMode_fileRoutesJson_onlyWithNestedLayout_then_isClientMenuUsed_false()
+            throws IOException {
+        mockDevelopmentMode();
+        String fileRoutesJson = "[{ \"route\": \"some-route\", \"children\": [ {\"route\": \"\", \"params\": {}, \"title\": \"Index\", \"children\": [] } ]}]";
```
cromoteca:
```suggestion
        String fileRoutesJson = """
                [{
                    "route": "some-route",
                    "children": [
                        {
                            "route": "",
                            "params": {},
                            "title": "Index",
                            "children": []
                        }
                    ]
                }]
                """;
```

cromoteca (changes requested):
Also (not modified by the PR):

```java
        String addedDevToRootRoute = """
                [{
                    "route": "dev",
                    "children": %s
                }]
                """.formatted(hierarchicalRoutesAsString);
```

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
taefi: Good catch! I changed the detection algorithm, and added tests for covering similar cases.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
```diff
@@ -209,6 +212,9 @@ private void registerAndRecurseChildren(String basePath,
         var path = view.getRoute() == null || view.getRoute().isEmpty()
                 ? basePath
                 : basePath + '/' + view.getRoute();
+        if (isMainLayout(view)) {
```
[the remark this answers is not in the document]
mshabarov: to not check for a main layout, if a main layout is already found.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
```diff
@@ -62,6 +62,8 @@ public class ClientRouteRegistry implements ClientRoutesProvider {
 
     private volatile LocalDateTime lastUpdated;
 
+    private boolean hasMainLayout;
```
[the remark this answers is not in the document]
mshabarov: as this flag is used in the synchronized methods.

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
```diff
@@ -170,6 +173,161 @@ public void when_developmentMode_then_loadLatestDevModeFileRoutesJsonIfNeeded_lo
...
+        String fileRoutesJson = """
+                [
+                    {
+                        "route": "some-route",
+                        "children": [
+                            {
+                                "route": "",
+                                "params": {},
+                                "title": "Index",
+                                "children": []
```
cromoteca, line 262: This is not a layout if no children, so this test case looks identical to the previous one to me.

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
```diff
@@ -170,6 +173,161 @@ public void when_developmentMode_then_loadLatestDevModeFileRoutesJsonIfNeeded_lo
...
+                        "route": "some-route",
+                        "children": [
+                            {
+                                "route": "",
+                                "params": {},
+                                "title": "Index"
+                            }
+                        ]
+                    }
+                ]""".stripIndent();
```
cromoteca, line 239: This does not look like the JSON generated in a real application when there's no root layout.
```suggestion
        String fileRoutesJson = """
                {
                    "route": "",
                    "params": {},
                    "children": [
                        {
                            "route": "some-route",
                            "params": {},
                            "title": "Some Layout",
                            "children": [
                                {
                                    "route": "",
                                    "params": {},
                                    "title": "Index"
                                }
                            ]
                        }
                    ]
                }""".stripIndent();
```

cromoteca (changes requested):

mshabarov:
Tested with [starter](https://github.com/vaadin/flow-hilla-hybrid-example/pull/35) and it works as expected.
I have a couple of suggestions above, but otherwise LGTM 👍

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
taefi: To my understanding a `"children": []` is different from no `children` present in the json. Having `"children"` with either an empty array or actual children is one of criteria of distinction between layouts and normal routes.

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
taefi: The case we are testing here is actually not to have a layout as the parent of the all the other routes (which is translated to "Main Layout").

Anyway, I have an application with the following structure (no main layout):
```
views/
|_ some-route/
    |_ @index.tsx 
```
I get the following entry in the `file-routes.json`:
```json
[
  {
    "route": "some-route",
    "params": {},
    "children": [
      {
        "route": "",
        "params": {},
        "title": "Some Route"
      }
    ]
  }
]
```
Thus, apart from the absence of `"params": {},` and the `"title": "Some Route"` instead of "Index" (which are not important to the logic we are testing here), it seems to be a real case.

mshabarov (dismissed):

mshabarov:
> shouldn't this PR also change the default for vaadin.expose-server-routes-to-client to true?

Yes , I think it should. I tested with this flag set to true.

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
cromoteca: OK I see, a layout with no views has an empty array.

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
```diff
@@ -170,6 +173,167 @@ public void when_developmentMode_then_loadLatestDevModeFileRoutesJsonIfNeeded_lo
...
+            throws IOException {
+        mockDevelopmentMode();
+        String fileRoutesJson = """
+                [
+                    {
+                        "route": "login",
+                        "title": "Login"
+                    },
+                    {
+                        "route": "",
```
cromoteca, line 318:
```suggestion
                        "route": "",
                        "title": "Something",
```

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
```diff
@@ -170,6 +173,167 @@ public void when_developmentMode_then_loadLatestDevModeFileRoutesJsonIfNeeded_lo
...
+    }
+
+    @Test
+    public void when_developmentMode_fileRoutesJson_onlyWithNestedLayout_then_hasMainLayout_false()
+            throws IOException {
+        mockDevelopmentMode();
+        String fileRoutesJson = """
+                [
+                    {
+                        "route": "some-route",
```
cromoteca, line 262:
```suggestion
                        "route": "some-route",
                        "title": "Something",
```

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
```diff
@@ -170,6 +173,167 @@ public void when_developmentMode_then_loadLatestDevModeFileRoutesJsonIfNeeded_lo
...
+    }
+
+    @Test
+    public void when_developmentMode_fileRoutesJson_MainLayoutWithChildren_then_hasMainLayout_true()
+            throws IOException {
+        mockDevelopmentMode();
+        String fileRoutesJson = """
+                [
+                  {
+                    "route": "",
```
cromoteca, line 203:
```suggestion
                    "route": "",
                    "title": "Something",
```

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/ClientRouteRegistryTest.java
```diff
@@ -170,6 +173,167 @@ public void when_developmentMode_then_loadLatestDevModeFileRoutesJsonIfNeeded_lo
...
     }
 
+    @Test
+    public void when_developmentMode_fileRoutesJson_MainLayoutNoChildren_then_hasMainLayout_is_true()
+            throws IOException {
+        mockDevelopmentMode();
+        String fileRoutesJson = """
+                [
+                    {
+                        "route": "",
```
cromoteca, line 183:
```suggestion
                        "route": "",
                        "title": "Something",
```

cromoteca (changes requested):
Everything has a `title` in real apps, so let's make sure this does not change the test outcome.

cromoteca:
> Everything has a `title` in real apps, so let's make sure this does not change the test outcome.

Actually, that's not true. Let me have a second look.

mshabarov (approved):

PR #2349, opened 2024-04-22
chore: Change ProtectedRoute export to a type property

Artur-:
This is to be in sync how Flow had to be done

### DIFF
--- a/packages/ts/react-auth/src/ProtectedRoute.tsx
+++ b/packages/ts/react-auth/src/ProtectedRoute.tsx
@@ -24,7 +24,7 @@ interface ProtectedRouteProps {
   element: JSX.Element;
 }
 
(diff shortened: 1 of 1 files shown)


packages/ts/react-auth/src/ProtectedRoute.tsx
```diff
@@ -42,6 +42,7 @@ export function ProtectedRoute({ redirectPath, access, element }: ProtectedRoute
 
   return element;
 }
+ProtectedRoute.type = 'ProtectedRoute'; // This is for copilot to recognize this
```
caalador, line 45: Should this also still be exported?

Artur-: Exporting was added in https://github.com/vaadin/hilla/pull/2347 to solve the same issue. It no longer needs to be exported for copilot to work

platosha (approved):

PR #2353, opened 2024-04-24
fix: adjust Hilla statistics collection

### LINKED ISSUE #2129: Fix Hilla stats to detect Lit vs. React usage based on deployment configuration
### Describe the bug

Hilla stats where being gathered based on the usage of `hilla-react` artifact, and now that the `hilla-react` is not there anymore, we can decide whether is it a Lit or React app based on `DeplymentConfiguration`'s method called `isReactEnabled`. 

### Expected-behavior

* The statistics should be gathered once per application startup.
* It would be ideal if the statistics gathered based on real usages of Hilla, not just Hilla artifacts being present on classpath, as in Vaadin 24.4+, Hilla is automatically part of the platform, unless it is excluded explicitely.

### Reproduction

Previous unit test for HillaStats is failing.

### System Info

N/A

taefi:
## Description

This adjusts the Hilla statistics gathering based on what's explained in vaadin/platform#6430

Fixes #2129

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


Legioth:
The description mentions three out of the four potential states of the two booleans. Is there a possibility that the fourth option could ever happen, i.e. `FrontendUtils.isReactRouterRequired() == false && deploymentConfiguration.isReactEnabled() == true`?

packages/ts/file-router/src/runtime/createMenuItems.ts
```diff
@@ -12,6 +12,9 @@ import type { MenuItem, ViewConfig } from '../types.js';
  * @returns A list of menu items.
  */
 export function createMenuItems(vaadinObject = (window as VaadinWindow).Vaadin): readonly MenuItem[] {
+  // @ts-expect-error: esbuild injection
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+  __REGISTER__('createMenuItems', vaadinObject);
```
mshabarov, line 17: This has no restrictions to run in the production mode, but should be tracked only in the dev mode. 
I propose to use this:
```
if (vaadinObject.developmentMode) {
  __REGISTER__('createMenuItems', vaadinObject);
}
```
and even better to add this check into __REGISTER__ function.

taefi: Calling the register is only adding an entry to the `Vaadin.registrations`, but AFAIK, those are not gathered nor sent in PROD mode. If that's not the case, then all Hilla TS packages should change since the same call to `__REGISTER__()` is happening in their index upon loading on the browser.

platosha (approved):

PR #2361, opened 2024-04-25
fix: [autogrid] ignore default object json renderer if columnOption `path` is set

### LINKED ISSUE #2322: AutoGrid uses JSON renderer for column of type object, although a custom path is set.
### Describe the bug

Having two entities `Order` and `Customer` in a Many-To-One-Relation, I use `AutoGrid` to display `Orders` and I would like to configure `AutoGrid` to display `customer.name` in the column for `Customer` using a custom `path`:
```tsx
<AutoGrid 
    model={OrderModel} 
    service={OrderService}
    columnOptions={{
      customer: {
        path: "customer.name",
      }
    }} 
/>
```
Unfortunately, `AutoGrid` keeps using the `AutoGridJsonRenderer` for column `customer` and displays the whole object, although a custom `path` is set, that points to a single attribute:
image

If I set the renderer to `null`, I get the desired result:
```tsx
<AutoGrid 
    model={OrderModel} 
    service={OrderService}
    columnOptions={{
      customer: {
        path: "customer.name",
        renderer: null
      }
    }} 
/>
```
image

### Expected-behavior

AutoGrid should not use the `AutoGridJsonRenderer` for a column of type `object`, if a custom path is set. It should be sufficient to configure a custom `path`. Setting the `renderer` to `null` should not be required.

Please also see the [comment](https://vaadin.com/forum/t/manytoone-relation-in-autogrid/165954/6) by @sissbruecker in this Vaadin Forum Thread, where this issue was raised as a question.

### Reproduction

Create two entities like `Order` and `Customer`:
```java
@Entity
@Table(name = "orders")
public class Order {

    @Id
    @GeneratedValue(strategy= GenerationType.IDENTITY)
    Long id;

    @NotNull
    private LocalDate created;

    @NotNull
    @ManyToOne(targetEntity = Customer.class, optional = false)
    private Customer customer;

   // Getters and Setters omitted

}

@Entity
@Table(name = "customers")
public class Customer {

    @Id
    @GeneratedValue(strategy= GenerationType.IDENTITY)
    Long id;

    @NotBlank
    String name;

    @NotBlank
    @Email
    String mail;

   // Getters and Setters omitted   
}
```

Create a `OrderService`:
```java
@BrowserCallable
@AnonymousAllowed
public class OrderService extends CrudRepositoryService<Order, Long, OrderRepository> {}
```

Use `AutoGrid` like this:
```tsx
<AutoGrid 
    model={OrderModel} 
    service={OrderService}
    columnOptions={{
      customer: {
        path: "customer.name",
      }
    }} 
/>
```

Actual result:
image

Expected result:
image

### System Info

Hilla: 24.4.0.alpha21
Hilla CLI: 2.0.1
Java: 21.0.2
OS: macOS Sonoma 14.4.1 (23E224)

rbrki07:
## Description

This should fix that AutoGrid uses the build-in JSON renderer for a column of type `object`, although a custom `path` is set for this column. Instead, it should use the default string renderer for the property target with `path`.

Fixes #2322

## Type of change

- [X] Bugfix
- [ ] Feature

## Checklist

- [X] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [X] The issue is created in the corresponding repository and I have referenced it.


taefi (commented):
LGTM

packages/ts/react-crud/src/autogrid-columns.tsx
```diff
@@ -112,6 +112,14 @@ export function getColumnOptions(
...
   return customColumnOptions
-    ? { ...typeColumnOptions, headerFilterRenderer, ...customColumnOptions }
+    ? {
+        ...typeColumnOptions,
+        headerFilterRenderer,
+        ...customColumnOptions,
+        // Use default renderer for column of type object, if custom path is set and no other renderer is configured
+        ...(propertyInfo.type === 'object' &&
+          customColumnOptions.path &&
+          customColumnOptions.renderer === undefined && { renderer: null }),
```
taefi, line 122: Functionality wise this looks fine.

Nit: Though, I would prefer that the `getTypeColumnOptions` could have the proper info to not return the `AutoGridJsonRenderer` at all in case no custom renderer is defined and the `customColumnOptions?.path !== undefined`.

taefi: @Lodin WDYT?

taefi: I believe your comment on the issue could be another feature. Here, I simply think that the `getTypeColumnOptions` should actually return the correct renderer based on the `propertyInfo` and `customColumnOptions` so that it doesn't need to be patched later down the stream. Having the logic that decides on the renderer isolated in some method is important for future maintenance.

taefi: What I'm suggesting is to have this taken care of in `getTypeColumnOptions` by adding the `columnOptions` as the parameter to it:
```ts
function getTypeColumnOptions(propertyInfo: PropertyInfo, customColumnOptions?: ColumnOptions)
```
and change the `case 'object':` e.g. as follows:
```tsx
case 'object':
    return {
        autoWidth: true,
        renderer:
          customColumnOptions?.path !== undefined && customColumnOptions.renderer == null ? null : AutoGridJsonRenderer,
        headerFilterRenderer: NoHeaderFilter,
    };
```

taefi (changes requested):

taefi (approved):
Nice work @rbrki07! LGTM!

PR #2381, opened 2024-05-02
feat(generator): support `@JsonCreator` and `@JsonValue` annotations

### LINKED ISSUE #2055: Support @JsonCreator and @JsonValue when generating TypeScript from Java
### Describe your motivation

In Java, it is a good practice to use domain primitives rather than ordinary primitives in certain situations. For instance, instead of using a `String` or `long` as the customer ID, you should create a `CustomerId` type that wraps a `String` or `long`. It should be possible to use domain primitives like this in `@BrowserCallable` services.

Although having domain primitives in TypeScript would be nice, it isn't as important as having them in Java. I would be fine with a solution where I can use domain primitives in Java and basic types in TypeScript. The easiest way to do this would be to support `@JsonValue` and `@JsonCreator` on the Java side.

My `CustomerId` class would have a method (maybe `toString()`) annotated with `@JsonValue`. An outgoing `CustomerId` object would then be serialized to a `string` in TypeScript.

I would also have a corresponding constructor that accepts a single `String` parameters and is annotated with `@JsonCreator`. An incoming TypeScript `string` would then be deserialized to a `CustomerId` object.

I could make separate `@BrowserCallable` services and DTOs that use ordinary primitives and do the conversion manually on the Java side before delegating to the "real" services and DTOs that use domain primitives. However, this feels like unnecessary complexity if the only difference between these services and DTOs is that the domain primitives have been replaced with their wrapped ordinary primitives.

### Additional context

_No response_

cromoteca:
## Description

Adds support for `@JsonCreator` and `@JsonValue` annotations to allow to customize how a type is transferred.

The main purpose is to provide support for domain primitives, like:

```java
    public record Email(@JsonValue String value) {
        @JsonCreator public Email {}
    }
```

Fixes #2055

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/JsonValuePlugin.java
+++ b/packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/JsonValuePlugin.java
@@ -23,10 +23,7 @@ import java.util.stream.Stream;
  */
 public class JsonValuePlugin
         extends AbstractPlugin<BackbonePluginConfiguration> {
-    private static class NotAJsonValue {
-    }
-
-    private final Map<Class<?>, Class<?>> jsonValues = new HashMap<>();
+    private final Map<Class<?>, Optional<Class<?>>> jsonValues = new HashMap<>();
 
     @Override
     public void enter(NodePath<?> nodePath) {
@@ -62,12 +59,10 @@ public class JsonValuePlugin
 
     private Optional<Class<?>> getValueType(Class<?> cls) {
         // Use cached results to avoid recomputing the value type.
-        var valueType = jsonValues.computeIfAbsent(cls, this::findValueType);
-        return valueType == NotAJsonValue.class ? Optional.empty()
-                : Optional.of(valueType);
+        return jsonValues.computeIfAbsent(cls, this::findValueType);
     }
 
-    private Class<?> findValueType(Class<?> cls) {
+    private Optional<Class<?>> findValueType(Class<?> cls) {
         // First of all, we check that the `@JsonValue` annotation is
         // used on a method of the class.
         var jsonValue = Arrays.stream(cls.getMethods())
@@ -95,7 +90,7 @@ public class JsonValuePlugin
                     + " Hilla only supports classes with both annotations.");
         }
 
-        return jsonValue.orElse(NotAJsonValue.class);
+        return jsonValue;
     }
 
     // this shouldn't be a runtime exception, but `resolve` doesn't allow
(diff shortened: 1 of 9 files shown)


packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/JsonValuePlugin.java
```diff
@@ -0,0 +1,106 @@
...
+        return jsonValues.computeIfAbsent(cls, this::findValueType);
+    }
+
+    private Optional<Class<?>> findValueType(Class<?> cls) {
+        // First of all, we check that the `@JsonValue` annotation is
+        // used on a method of the class.
+        var jsonValue = Arrays.stream(cls.getMethods())
+                .filter(method -> Arrays.stream(method.getAnnotations())
+                        .map(ann -> ann.annotationType().getName()).anyMatch(
+                                name -> name.equals(JsonValue.class.getName())))
```
taefi, line 74: Couldn't the `method.isAnnotationPresent(JsonValue.class)` be used here instead of streaming the method annotations and checking the name equality?

packages/java/parser-jvm-plugin-backbone/src/test/java/com/vaadin/hilla/parser/plugins/backbone/jsonvaluenojsoncreator/JsonValueNoJsonCreatorTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class JsonValueNoJsonCreatorTest {
+    private final TestHelper helper = new TestHelper(getClass());
+
+    @Test
+    public void should_CorrectlyMapJsonValue() {
```
taefi, line 17: The test name should change to reflect that it throws when json value is used without the json creator.

taefi (changes requested):

packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/JsonValuePlugin.java
cromoteca: Good catch! Damn copy-paste.

Just for information, I haven't added the opposite test (only `JsonCreator`), since such an endpoint crashes the generator anyway.

cromoteca: Definitely. This is a leftover of an initial different implementation. Thanks for the suggestion.

taefi (approved):
LGTM!

PR #2383, opened 2024-05-03
chore: Make hilla endpoint optional for plugin

mshabarov:
This is an alternative for https://github.com/vaadin/platform/pull/5264 that prevents `hilla-dev` to include Hilla artifacts (endpoint artifact) into projects that don't use Spring and Hilla.

Fixes https://github.com/vaadin/platform/issues/5260


platosha (approved):
Makes sense to me.

PR #2384, opened 2024-05-06
chore: bump snapshot version to 24.5

platosha:


.github/workflows/validation.yml
```diff
@@ -6,7 +6,7 @@ defaults:
 
 on:
   push:
-    branches: [ main, '2.5', '2.4', '2.3', '2.2', '1.4', '1.3' ]
+    branches: [ main, '24.4', '2.5', '1.4', '1.3' ]
```
taefi, line 9: I wonder if we should add `1.5` here and remove `1.3` already.

packages/java/tests/pom.xml
```diff
@@ -15,7 +15,7 @@
     <packaging>pom</packaging>
 
     <properties>
-        <testbench.element.version>24.4-SNAPSHOT</testbench.element.version>
+        <testbench.element.version>24.5-SNAPSHOT</testbench.element.version>
```
taefi, line 18: Could we postpone upgrading `testbench.element` to the near future when it's released? Now, it is breaking the build.

pom.xml
```diff
@@ -75,7 +75,7 @@
     <jackson.version>2.17.0</jackson.version>
     <junit.version>5.10.1</junit.version>
     <mockito.version>4.5.0</mockito.version>
-    <flow.version>24.4-SNAPSHOT</flow.version>
+    <flow.version>24.5-SNAPSHOT</flow.version>
```
taefi, line 78: Could we postpone upgrading `Flow` to the near future when it's released? Now, it is breaking the build.

taefi (changes requested):

pom.xml
platosha: Seems to work now.

taefi (approved):

PR #2401, opened 2024-05-08
chore: change ClientViewMenuConfig order type

tltv:
Changing ClientViewMenuConfig order field type from Long to Double to match with Flow's RouteData and Menu annotation.

Related-to: vaadin/flow/issues/19332


taefi (approved):
LGTM!

PR #2424, opened 2024-05-14
chore: Delete polling based hotswap listener

Artur-:
It currently is enabled for Flow applications and logs a lot for no good reason. It, and Hotswapper, will be replaced by a common solution for Flow+Hilla


taefi (approved):

taefi:
@Artur- This isn't going to be picked to 24.4, right?

PR #2432, opened 2024-05-15
feat!(frontend): reload the page on login / logout success

### LINKED ISSUE #2235: Hilla doesn't redirect to a given URL when logout with Hilla's useAuth API
### Description of the bug

Hilla doesn't redirect to a `logoutSuccessfulUrl` being set with `VaadinWebSecurity::setLoginView(http, "/login", "/")` if the logout method uses Hilla's provided `useAuth::logout` API.

It performs the logout successfully, but the page isn't redirected and the page content is visible.

If I logout from a Hilla view, it redirects me to login page, which also a bit unexpected, but better than keeping the content of the current page.

Here is how I configure my logout:

**MainLayout.tsx:**
```
import {useAuth} from "../auth";
...
const { state, logout } = useAuth();
...
<Button onClick={async () => logout()}>Sign out</Button>
```
**auth.tsx**
```
import { UserInfoService } from 'Frontend/generated/endpoints';
import {configureAuth} from "@vaadin/hilla-react-auth";

const auth = configureAuth(UserInfoService.getUserInfo, {
    getRoles: (userInfo) => userInfo.authorities,
});

export const useAuth = auth.useAuth;
export const AuthProvider = auth.AuthProvider;
```

Browser sends POST "/logout" request to server and it's caught by `UidlRedirectStrategy`. Then it falls into `super.sendRedirect(request, response, url);` and redirect doesn't happen for some reason.

### Expected behavior

Flow should redirect to "/" after logout.

1. Checkout branch https://github.com/vaadin/flow-hilla-hybrid-example/tree/stateful-auth
2. mvn
3. Log in with admin/admin
4. Navigate to Flow view
5. Click on Logout button
6. Browser the Flow content is still there.

### Versions

- Vaadin / Flow version: 24.4.0.alpha15

Lodin:
Fixes #2235

In addition, introduces `onSuccess` and  `navigate` callbacks for custom authentication workflows, such as retrieving user info and offline storage.


mcollovati (approved):
Can't say much about the implementation, but the issues that caused tests failures due to unexpected redirect to login page, seem to be fixed by using the newly introduced  `onSuccess` callback to store authentication details.
Good job!

packages/ts/react-auth/src/useAuth.tsx
Lodin, line 1: So, we cannot make a logout a simple document reload?

platosha: This wouldn't respect the logout success URL specified in Spring Security. Most of the time reloading a private view after logging out would redirect to the login form instead of the root other target location.

packages/ts/frontend/src/Authentication.ts
```diff
@@ -59,12 +61,73 @@
...
+}
+
+/**
+ * Navigates to the provided path using page reload.
+ *
+ * @param to - navigation target path
+ */
+function navigateWithPageReload(to: string) {
+  // Consider absolute path to be within application context
+  const url = to.startsWith('/') ? new URL(`.${to}`, document.baseURI).toString() : to;
```
Lodin, line 129: You don't need to convert URL to string, `replace` supports URL. I would say, the URL is better here

packages/ts/frontend/src/Authentication.ts
```diff
@@ -59,12 +61,73 @@ export interface LoginResult {
...
 
+export type SuccessCallback = () => Promise<void> | void;
+
+export type NavigateFunction = (path: string) => void;
+
 export interface LoginOptions {
+  /**
+   * The URL for login request, defaults to `/login`.
+   */
   loginProcessingUrl?: string;
```
Lodin, line 72: Maybe, we should support `URL` here or change the naming? It is a bit confusing that `...Url` does not support real `URL`.

packages/ts/frontend/src/Authentication.ts
```diff
@@ -59,12 +61,73 @@ export interface LoginResult {
...
+   * reloads the page.
+   */
+  navigate?: NavigateFunction;
 }
 
 export interface LogoutOptions {
+  /**
+   * The URL for logout request, defaults to `/logout`.
+   */
   logoutUrl?: string;
```
Lodin, line 90: Should we also support `URL` here?

Lodin (commented):
LGTM
(I cannot approve my own PR)

platosha (approved):

PR #2392, opened 2024-05-06
refactor: disallow parametrized items in auto menu

Lodin:
Fixes #2378.

PR #2447, opened 2024-05-21
fix: support wildcard route parameter in menu

tltv:
Fixes server menu item url for routes with route parameter with a wildcard like (route/:param*).

RelatedTo: vaadin/flow/issues/19392


platosha:
Would it make sense to cherry-pick https://github.com/vaadin/hilla/pull/2392 instead?

tltv:
> Would it make sense to cherry-pick #2392 instead?

24.4 uses RouteData which don't have RouteParameterType and 24.5(main) uses AvailableViewInfo. Makes no sense in cherry-picking in practice as some parts regarding menu is moved to Flow in 24.5.
Also in 24.5 ([PR](https://github.com/vaadin/flow/pull/19406)), Flow will clean up optional (and wildcard which is also optional) parameters from the url and exclude menu items that have required parameter by setting menu item exclude=true. Flow has to do that in the end anyway to keep data same for client and server menu.

platosha (approved):

PR #2448, opened 2024-05-21
refactor(file-router): tune server routes addition in case of variable routes presented

### LINKED ISSUE #2404: Optional route parameters are broken when using withFallback
### Describe the bug

A route with an optional parameter (i.e. a file like `views/foo/{{param}.tsx`) will not rended without a parameter value if `.withFallback(Flow)` is in use. Instead, the 404 page from Flow is shown, along with "supports optional parameter" for the view in question.

### Expected-behavior

Can navigate to a view with an optional parameter without providing a parameter.

### Reproduction

1. Add `views/foo/{{param}.tsx` to an application from start.vaadin.com
2. Navigate to `/foo`
3. Boom

The problem disappears if you create a copy of `generated/routes.tsx` and comment out the `.withFallback(Flow)` line.

### System Info

Vaadin 24.4.0.beta1

Lodin:
Fixes #2404.


ZheSun88:
```
  NX   Running target test:coverage for 16 projects failed

   Failed tasks:
   
   - @vaadin/hilla-file-router:test:coverage
```

Lodin:
@ZheSun88, yeah, it looks like there is some kind of a FS handling on different systems. I cannot reproduce that locally but @taefi managed to do it

platosha (approved):

PR #2449, opened 2024-05-21
feat: use flow client route collection and filtering

caalador:
requires vaadin/flow#19389


packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
```diff
@@ -128,28 +128,26 @@ public void modifyIndexHtmlResponse(IndexHtmlResponse response) {
...
-                        config.isAutoRegistered(), config.menu(),
-                        Collections.emptyList(), config.getRouteParameters()))
-                .collect(Collectors.toMap(AvailableViewInfo::route,
-                        Function.identity()));
+        return MenuRegistry
+                .collectClientMenuItems(true, deploymentConfiguration, request)
+                .entrySet().stream()
+                .filter(view -> !hasRequiredParameter(
+                        view.getValue().routeParameters()))
+                .collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
```
taefi, line 143: I think it worth checking that the `MenuRegistry#validateViewAccessible` (used in vaadin/flow#19389) is actually checking A&A in the same way that `filter(config -> routeUtil.isRouteAllowed(isUserInRole, isUserAuthenticated, config))` was doing recursively for the views and the possible parent layout(s) configurations.

taefi (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
```diff
@@ -128,28 +128,26 @@ public void modifyIndexHtmlResponse(IndexHtmlResponse response) {
...
-                .filter(config -> routeUtil.isRouteAllowed(isUserInRole,
-                        isUserAuthenticated, config))
-                .map(config -> new AvailableViewInfo(config.getTitle(),
-                        config.getRolesAllowed(), config.isLoginRequired(),
-                        config.getRoute(), config.isLazy(),
-                        config.isAutoRegistered(), config.menu(),
-                        Collections.emptyList(), config.getRouteParameters()))
-                .collect(Collectors.toMap(AvailableViewInfo::route,
-                        Function.identity()));
+        return MenuRegistry
```
taefi, line 138: This part was trying to optimize loading of the latest changes in the client-side menus (in dev mode), and now that they are being collected by MenuRegistry, it seems a bit pointless. `ClientRouteRegistry` was the only source of truth about the routes and was preventing concurrency by having all methods `synchronized` due to having some  states: 

* `registeredRoutes` (originally for optimizations)
* `hasMainLayout` (which can be determined only when traversing the json hierarchy of menus). 
 
The optimizations during the Dev mode is not very critical, but the `hasMainLayout` is used for handling this logic, which I'm not sure how is it handled anymore, maybe on Flow side?

taefi (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
caalador: The MenuRegistry removes all child routes for a route that is not valid, which is the same as done here, but from parent to child instead of child to parent.

The output should be exactly the same with no changes in the result.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
caalador: The MenuRegistry always recreates the objects from the actual file. We could have caching there or here if it's necessary, but at the moment it is always generated and thus only a change to the file while reading could mess things up.

hasMainLayout does not have a method in flow as there is no mainLayout there.
Looking at it now it is populated for non prod mode running the `loadLatestDevModeFileRoutesJsonIfNeeded` method. So this is something that is not fully correct.

 What is the requirement for a mainLayout? Is it just no route but has childern and is not a child of anyone? As that we could probably check from the results?

taefi: > it is populated for non prod mode

The `hasMainLayout` is initialized in `registerAndRecurseChildren` which is called in `public synchronized void registerClientRoutes`, so it is not only for Dev mode, the `RouteUnifyingServiceInitListener` also calls it in the Prod mode.

> What is the requirement for a mainLayout?

For generating an automatic menu, the requirement was defined as: Server side routes should only be populated to the `window.Vaadin.views` if a main layout exists (supposedly the menu is rendered there). That logic was checked [here](https://github.com/vaadin/hilla/blob/main/packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java#L169-L175).

> Is it just no route but has childern and is not a child of anyone?

A layout is defined as an entry in the `file-routes.json` that has:
- has no parent (for distinguishing it from intermediary layouts), and
- `"route": ""`, and
- `"children": []` or `"children": [some entries]` (`"children"` should not be undefined)

"having no parents" was basically one of the motivations for not loading and traversing the `file-routes.json` and trying to infer it while reading that file only when it is necessary.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
Lodin, line 143:
```suggestion
                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListenerTest.java
```diff
@@ -36,10 +42,15 @@
...
+import com.vaadin.flow.server.menu.MenuRegistry;
 import com.vaadin.flow.server.menu.RouteParamType;
-import com.vaadin.hilla.route.records.ClientViewConfig;
 
+import static com.vaadin.flow.server.menu.MenuRegistry.FILE_ROUTES_JSON_NAME;
+import static com.vaadin.flow.server.menu.MenuRegistry.FILE_ROUTES_JSON_PROD_PATH;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.CALLS_REAL_METHODS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
```
Lodin, line 53: Let's follow a single code style for all the Mockito imports. If we use `when` as static import, then there should be no `Mockito.when`. And vice versa.

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListenerTest.java
```diff
@@ -414,9 +490,39 @@ private void mockDevelopmentMode() throws IOException {
...
+        File clientFiles = new File(frontendGeneratedDir,
+                FILE_ROUTES_JSON_NAME);
+        Files.writeString(clientFiles.toPath(), clientRoutes);
     }
 
     @PageTitle("RouteTarget")
     private static class RouteTarget extends Component {
     }
+
+    String clientRoutes = """
```
Lodin, line 503: Could we have it as a JSON resource?

Lodin (changes requested):
Some code style notes

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
caalador: done

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListenerTest.java
caalador: done.

packages/java/endpoint/src/test/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListenerTest.java
caalador: done.

Lodin (dismissed):
LGTM

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
taefi: @Lodin 
According to the changes in this PR, what is sent the browser is collected by `MenuRegistry`, but the other parts of Hilla will still be relying on the loaded cache of `ClientRouteRegistry` (it means codes calling `clientRouteRegistry::getAllRoutes` are getting whatever is already collected in the `menuRegistry` map). I still believe there should be only one source of truth since these two can easily diverge in the near future and get out of sync.

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
```diff
@@ -190,6 +169,38 @@ protected Map<String, AvailableViewInfo> collectServerViews() {
                         Function.identity()));
     }
 
+    private boolean hasMainMenu(DeploymentConfiguration deploymentConfiguration,
+            VaadinRequest request) {
+        Map<String, AvailableViewInfo> clientItems = MenuRegistry
+                .collectClientMenuItems(true, deploymentConfiguration, request);
```
taefi, line 175: Probably the result of calling the `MenuRegistry.collectClientMenuItems` could be passed to this method to avoid a second call and loading/reading the json file again.

taefi (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUnifyingIndexHtmlRequestListener.java
caalador: done

packages/java/endpoint/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
```diff
@@ -4,6 +4,4 @@ com.vaadin.hilla.ApplicationContextProvider
 com.vaadin.hilla.crud.CrudConfiguration
 com.vaadin.hilla.startup.EndpointRegistryInitializer
 com.vaadin.hilla.startup.RouteUnifyingServiceInitListener
-com.vaadin.hilla.route.ClientRouteRegistry
-com.vaadin.hilla.route.RouteUtil
```
taefi, line 8: Not sure if RouteUtil should be removed from the AutoConfiguration imports, IIRC, the some Hilla security code samples related to spring security configuration (user codes) is instructing to utilize this.

packages/java/endpoint/src/main/java/com/vaadin/hilla/startup/RouteUnifyingServiceInitListener.java
```diff
@@ -99,11 +85,12 @@ public void serviceInit(ServiceInitEvent event) {
                     "{} mode: Registered RouteUnifyingIndexHtmlRequestListener.",
                     deploymentMode);
 
-            clientRouteRegistry.registerClientRoutes(deploymentConfiguration,
-                    LocalDateTime.now());
+            Map<String, AvailableViewInfo> stringAvailableViewInfoMap = MenuRegistry
+                    .collectClientMenuItems(false, deploymentConfiguration,
+                            null);
+
+            boolean hasHillaFsRoute = !stringAvailableViewInfoMap.isEmpty();
```
taefi, line 92:
```suggestion
            Map<String, AvailableViewInfo> clientMenus = MenuRegistry
                    .collectClientMenuItems(false, deploymentConfiguration,
                            null);

            boolean hasHillaFsRoute = !clientMenus.isEmpty();
```

taefi (changes requested):

packages/java/endpoint/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
caalador: Route util requires ClientRouteRegistry and as that doesn't exist we do not have a RouteUtil either anymore.
What parts of the functionality is referenced for use? The class itself was even marked as `For internal use only`

packages/java/endpoint/src/main/java/com/vaadin/hilla/springnative/HillaHintsRegistrar.java
```diff
@@ -40,8 +39,6 @@ public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
         registerEndpointTypes(hints);
 
         hints.resources().registerPattern("file-routes.json");
-        hints.reflection().registerType(ClientViewConfig.class,
-                MemberCategory.values());
```
taefi, line 44: Reminder: This also needed to be verified separately/manually by running a native build for the `[hilla-project]/tests/spring/native` project, as it is not part of the validation build workflow ATM.

packages/java/endpoint/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports
taefi: At least this repo uses RouteUtil:
https://github.com/vaadin/flow-hilla-hybrid-example/blob/v24.4/src/main/java/org/vaadin/example/security/SecurityConfig.java#L30
Let's ping @cromoteca as there's a high chance this have changed very recently and I'm not up-to-date.

packages/java/endpoint/src/main/java/com/vaadin/hilla/springnative/HillaHintsRegistrar.java
caalador: So what is required for a native build execution?
Has it been tested on windows? I just keep getting `cannot find module` issues.

caalador: Didn't have any issues with the native apart from the old node version.

taefi: If the native build and running the image was successful, this can be considered as resolved.

taefi (approved):
LGTM!

PR #2480, opened 2024-05-27
chore: switch from Karma to Web Test Runner + Vite

Lodin:
Karma is [deprecated](https://github.com/karma-runner/karma?tab=readme-ov-file#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes), and they recommend using WTR as a replacement. We had it as a main test runner once but it was broken because of modernweb-dev/web#1700. The issue is still here, so until now I didn't consider reimplementing it.

However, I managed to set it up using the [web-test-runner-vite-plugin](https://github.com/remcovaes/web-test-runner-vite-plugin). It allows using Vite as an intermediate server between WTR and the code, and it respects all the settings and plugins we've created for `karma-vite`.


platosha:
Closing in favour of Vitest, which seems to be a better alternative.

PR #2493, opened 2024-05-30
refactor(file-router): make route with non-null children detected as a layout

### LINKED ISSUE #2379: @layout.tsx is shown in the menu if there are no views
### Describe the bug

If you go to start.vaadin.com -> download -> Hilla main layout

then you get an application that has a `@layout.tsx` and no views, but it looks like this
image

### Expected-behavior

It should look like this
image

### Reproduction

As above

### System Info

Vaadin 24.4 beta1

Lodin:
Fixes #2379.

### DIFF
--- a/packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
+++ b/packages/java/endpoint/src/main/java/com/vaadin/hilla/route/ClientRouteRegistry.java
@@ -15,18 +15,6 @@
  */
 package com.vaadin.hilla.route;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.vaadin.flow.function.DeploymentConfiguration;
-import com.vaadin.flow.server.frontend.FrontendUtils;
-import com.vaadin.hilla.route.records.ClientViewConfig;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.stereotype.Component;
-import org.springframework.util.AntPathMatcher;
-import com.vaadin.flow.router.internal.ClientRoutesProvider;
-
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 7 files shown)


taefi (commented):
There is this method called `isMainLayout` in the `ClientRouteRegistry` which is dependent on the children to be not null (could be empty array, but not null) for deciding about an entry being a MainLayout. The correct detection of a MainLayout is critical for the Flow views to be added to the menu. A case comes to mind that an application has a MainLayout in Hilla, but no Hilla views, and only some Flow views. Is this changes currently supporting that?

Lodin:
Nope. I guess, I can just exclude it from filtering and add as a route

taefi:
> Nope. I guess, I can just exclude it from filtering and add as a route

Probably, adding an exception case for MainLayout is the best choice for the short term. TBH, this part is a bit messy regarding what should be on Hilla side, and what should be on Flow side, and there are some related changes introduced in the #2449. Anyway, IMO it's best not to break the currently agreed functionality.

Lodin:
Hmm, maybe then it would be better to filter in the Hilla menu instead of a server?

platosha (dismissed):

platosha:
> There is this method called isMainLayout in the ClientRouteRegistry which is dependent on the children to be not null (could be empty array, but not null) for deciding about an entry being a MainLayout. The correct detection of a MainLayout is critical for the Flow views to be added to the menu. A case comes to mind that an application has a MainLayout in Hilla, but no Hilla views, and only some Flow views. Is this changes currently supporting that?

Seems so, as this line suggests: https://github.com/vaadin/hilla/pull/2493/files#diff-fd76ca650f85b66ff69dfdb7b1ff3248840101fef661307b71bb79bced728c10R90

- layouts have empty children by default, the main layout is not an exception
- leaf views otherwise have undefined children (parsed as null in Java)

platosha:
So let me approve again, as according to my understanding, this change makes it more consistent, and the expectation for the main layout is fulfilled.

platosha (approved):

taefi (approved):

Issue #2060, opened 2024-02-12
[lit-form] make model API not depend on form binder


platosha:
In few cases, the current models' API implementation requires direct access to the respective binder node:
- `AbstractModel::valueOf()` - alternative: `binderNode.value` 
- `AbstractModel::toString()` - alternative: `String(binderNode.value)`
- `ArrayModel::[Symbol.iterator]()` - no alternative yet

As part of introducing [Object-based models](https://github.com/vaadin/hilla/issues/1903), and extracting models into a separate package for wider use, it would be great to replace these APIs with some alternatives that do not require invoking binder nodes from models.

Let's deprecate these APIs and provide alternative API for iterating arrays in the binder node.

Documentation needs to be updated accordingly.

PR #2497, opened 2024-06-03
fix(binder): update form when array values change

### LINKED ISSUE #1196: React form array APIs
Let's discuss and figure out the API of using arrays in React forms. Consider the following use cases:

- rendering the array items, for example, rows in an order form
- binding the item fields using the model
- accessing the item state (`value`, `errors`, and such)
- providing the item `key` for rendering in React
- mutating the array (appending and prepending new empty (or given) items, removing an existing item)

One simple way items is to add the `.map` method directly to the `ArrayModel` that maps the item models using the provided callback, similar to `Array.prototype.map`. However, since we share models between Lit and React binders, the model API should preferably stay free of Lit and React opinionated helpers. This would mean that, as a baseline, the mapping callback provides the only the `model` for the rendering:

```tsx
function OrderForm() {
  const {model, field} = useForm(OrderFormModel);
  
  return <ul>
    {
      model.rows.map((rowModel) => (
        <li>
          <TextField label="Product" {...field(rowModel.product)}/>
        </li>
      ))
    }
  </ul>
}
```

This alone is clearly not sufficient for the use cases. Based on the `access` idea from https://github.com/vaadin/hilla/issues/1186, we could potentially add the `key` and the array mutation helpers to the `access` result:

```tsx
function OrderForm() {
  const {model, field, access} = useForm(OrderFormModel);
  
  return <ul>
    {
      model.rows.map((rowModel) => (
        <li key={access(rowModel).key}>
          <TextField label="Product" {...field(rowModel.product)}/>
          <button onClick={access(rowModel).removeSelf}>Remove</button>
        </li>
      ))
    }
    <button onClick={access(model).appendItem}>Add</button>
  </ul>
}
```

Or, alternatively, we could add another hook for the array use case:

```tsx
function OrderForm() {
  const {model, field, access} = useForm(OrderFormModel);
  const {append, key, remove} = useFormArray(model.rows);
  
  return <ul>
    {
      model.rows.map((rowModel) => (
        <li key={key(rowModel)}>
          <TextField label="Product" {...field(rowModel.product)}/>
          <button onClick={remove(rowModel)}>Remove</button>
        </li>
      ))
    }
    <button onClick={append}>Add</button>
  </ul>
}
```

cromoteca:
When accessing an array in a form, fields do not show their values. This PR adds `update` calls to inform React about value changes.

With this change, it is possible to use arrays in forms without any new API. Nevertheless, #1196 stays open as discussion for a better API.


Lodin:
I have a proposal for the API here. The code that is considered here is the following:
```ts
const TeamForm: React.FC = () => {
    const { field, model } = useForm(TeamModel);
    const { map } = useFormArray(model.players);
    const name = useFormPart(model.name);

    return (
      <>
        <input data-testid="team.name" type="text" {...field(model.name)} />
        <output data-testid="validation.team.name">
          {name.invalid ? name.ownErrors.map((e) => e.message).join(', ') : 'OK'}
        </output>
        {map(({ model: player, invalid, ownErrors }) => (
          <>
            <input
              key={`${player.valueOf().id}`}
              data-testid={`lastName.${player.valueOf().id}`}
              type="text"
              {...field(player.lastName)}
            />
            <output data-testid={`validation.lastName.${player.valueOf().id}`}>
              {invalid ? ownErrors.map((e) => e.message).join(', ') : 'OK'}
            </output>
          </>
        ))}
      </>
    );
  };
```

However, we may probably not need an additional API created for that specific case. We already have all the elements to do the correct array iteration.

```ts
function TeamFormPlayer({ model }: { model: PlayerModel }) {
    const { field, value, invalid, ownErrors } = useFormPart(model);

    return (
      <div>
        <input data-testid={`lastName.${value!.id}`} type="text" {...field(model.lastName)} />
        <output data-testid={`validation.lastName.${value!.id}`}>
          {invalid ? ownErrors.map((e) => e.message).join(', ') : 'OK'}
        </output>
      </div>
    );
  }

  function TeamForm() {
    const { field, model } = useForm(TeamModel);
    const name = useFormPart(model.name);

    return (
      <>
        <input data-testid="team.name" type="text" {...field(model.name)} />
        <output data-testid="validation.team.name">
          {name.invalid ? name.ownErrors.map((e) => e.message).join(', ') : 'OK'}
        </output>
        {Array.from(model.players, (player) => (
          <TeamFormPlayer key={`${player.value!.id}`} model={player.model} />
        ))}
      </>
    );
  }
```
The pros of my approach is that we don't need to support additional API, so we need less tests and push less code to the user's machine. It's also a more React-ish approach since it uses a separate component for the array item.
The cons is that to make it shine, the user would need to create an additional component for the array item because it requires using `useFormPart`.

What do you think?
cc @Legioth, @platosha

cromoteca:
The API should not only provide `map`, but also `append` and `remove`. Would direct array manipulation be supported without additional API?

Lodin:
@cromoteca Yes, you can use a combination of `part.value` and `part.setValue`:
```ts
const { setValue, value } = useFormPart(model);

setValue([...value, newItem]);
setValue(value.filter((item) => item.lastName === 'Doe'));
```

packages/ts/react-form/src/index.ts
```diff
@@ -72,6 +75,12 @@ export type UseFormResult<M extends AbstractModel> = Omit<UseFormPartResult<M>,
     update(): void;
   }>;
 
+export type UseFormArrayResult<M extends ArrayModel> = Readonly<{
+  append(item: Value<ArrayItemModel<M>>): void;
+  list: ReadonlyArray<UseFormPartResult<ArrayItemModel<M>>>;
+  remove(index: number): void;
+}>;
+
```
Lodin, line 83: I guess, this one remains unused since we've deleted all the code. We can wait until the discussion is finished and then either remove it or leave it here.

Lodin (commented):
LGTM but let's wait for the outcome of the discussion

cromoteca:
I'm testing the API you proposed and it seems to cover the required use cases.

Interestingly enough, as such this PR becomes a bugfix and IMHO should be backported.

platosha:
What I'm missing from the `Array.from` approach:
- being able to remove an item by model. Could be just a matter of exposing `binderNode.removeSelf()` in example though.
- this seems to expose BinderNode directly to the React user. While this seems to work well for providing a `.model`, in other methods it won't be as clean as with a hook, because they presume that the user works with it as with a class instance. For instance, calling `binderNode.removeSelf()` passes `this`, but would it still work with destructuring `({removeSelf}) => (<button onClick={removeSelf}>Remove</button>)`?
- a way of providing a `key` for new items generated with, e. g., `append`. Using `id` from the object as a key is a good practice, but the id won't be available for the items added on the client.

platosha:
Also, for the future, I was proposing to deprecate the model methods that use / expose binder nodes, so that we could finally resolve the dependency circle, see https://github.com/vaadin/hilla/issues/2060. The `Array.from(arrayModel)` and `model.valueOf()` examples would utilise the things that were causing some trouble so far — maybe not a good idea...

platosha:
With #2060 in mind, something a bit better would be adding a function that iterates through model instances, not binder nodes:

```typescript
export class ArrayModel<MItem extends AbstractModel = AbstractModel> extends AbstractModel<Array<Value<MItem>>> {
  //...
  
  /**
   * Iterates the current array value and yields a model for every item.
   */
  *items(): Generator<MItem> {
    const array = this.valueOf();

    if (array.length !== this.#items.length) {
      this.#items.length = array.length;
    }

    for (let i = 0; i < array.length; i++) {
      let item: MItem | undefined = this.#items[i];

      if (!item) {
        item = this.#createItem(this, i);
        this.#items[i] = item;
      }

      yield item;
    }
  }
}
```

Unlike `Array.from(arrayModel)`, using `Array.from(arrayModel.items())` would not leak binder nodes in React context.

Lodin (approved):

platosha:
We discussed the options and concluded that we'd introduce a separate `useFormArrayPart` hook:

```tsx
  function TeamFormPlayer({ model }: { model: PlayerModel }) {
    const { field, value, invalid, ownErrors } = useFormPart(model);

    return (
      <div>
        <input data-testid={`lastName.${value!.id}`} type="text" {...field(model.lastName)} />
        <output data-testid={`validation.lastName.${value!.id}`}>
          {invalid ? ownErrors.map((e) => e.message).join(', ') : 'OK'}
        </output>
      </div>
    );
  }

  function TeamForm() {
    const { field, model } = useForm(TeamModel);
    const { items, setValue: setPlayers, value: playersValue } = useFormArrayPart(model.players);
    const name = useFormPart(model.name);

    return (
      <>
        <input data-testid="team.name" type="text" {...field(model.name)} />
        <output data-testid="validation.team.name">
          {name.invalid ? name.ownErrors.map((e) => e.message).join(', ') : 'OK'}
        </output>
        {items.map((playerModel, index) => (
          <TeamFormPlayer key={`${playersValue[index].id ?? -index}`} model={playerModel} />
        ))}
      </>
    );
  }
```

PR #2502, opened 2024-06-04
feat: add NumberSignal basic support

### LINKED ISSUE #2429: [Full-stack Signals] Add basic support for NumberSignals
### Describe your motivation

There should be support for exposing a `@BrowserCallable` service containing public methods returning a `NumberSignal` subscription type as follows:
```java
@BrowserCallable
@AnonymousAllowed
public class CounterService {
  private final NumberSignal counter = new NumberSignal(0);

  public NumberSignal counter() {
    return counter;
  }
}
```
in a way that the necessary TS codes are generated so that the user code can contain usages like this:
```tsx
import { CounterService } from "Frontend/generated/endpoints";

const counter = CounterService.counter();

export default function CounterView() {
  return <Button onClick={() => counter.increment(1)}>Click count: {counter}</Button>
}
```

N/A

This very first step can ignore as many complexities as possible such as:
- disconnecting/reconnecting the Websocket connection when moving away from the view of having connection problems
- having an event queue on the server to resolve the concurrency issues and etc.
- support for latency compensation

### Additional context

The following modules are targeted for adding this feature:
- TS: `react-signals`
- Java: `endpoint` (under the `com.vaadin.hilla.signals`)

taefi:
## Description
Add basic support for full-stack `NumberSignal` type to be returned by endpoints as a shared signal instance that supports number value operations. The client-side user code (for now) need to do manual wrapping of the needed metadata about the endpoint method(s) that return shared signal instance(s).

Assuming a simple endpoint returning a `NumberSignal` instance like this:
```java
import com.vaadin.hilla.signals.NumberSignal;
import com.vaadin.hilla.BrowserCallable;

@AnonymousAllowed
@BrowserCallable
public class SharedCounterService {
    private final NumberSignal counter = new NumberSignal(0);

    public NumberSignal counter() {
        return counter;
    }
}

```
The needed manually written user code is like this (this will be generated later):
```ts
import { NumberSignalQueue } from "@vaadin/hilla-react-signals";
import client_1 from "Frontend/generated/connect-client.default";

export class CounterServiceWrapper {
  static counter() {
    const signalChannel = new NumberSignalChannel('SharedCounterService.counter', client_1);
    return signalChannel.signal;
  }
}
```
Then then usage of this wrapping service in views could be similar to what is described in the related issue:
```tsx
const counter = CounterServiceWrapper.counter();

export default function NumberSignalView() {
  return (
    <Button onClick={() => counter.value++}>
       Click count: { counter }
    </Button>
  );
}
``` 
NOTE: Since the Full-stack signals are experimental for now, please don't forget to enable the `fullstackSignals` feature flag in your applications (by adding the following in `src/main/resources/vaadin-featureflags.properties`):
```
com.vaadin.experimental.fullstackSignals=true
```

Fixes #2429 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


Legioth:
Why is `package-lock.json` changed without any corresponding change to `package.json`?

packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/core/EventQueue.java
```diff
@@ -0,0 +1,124 @@
...
+import reactor.core.publisher.Sinks.Many;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.locks.ReentrantLock;
+
+public abstract class EventQueue<T extends StateEvent> {
```
Legioth, line 17: Why is an event queue needed for tracking a single number value? A queue will eventually be needed for efficient resynch of more complex data structures but the most efficient way of catching up for a single value is always to just load the latest value.

taefi:
> Why is `package-lock.json` changed without any corresponding change to `package.json`?

Good catch. That was unintentionally committed/pushed. Reverted.

packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/core/EventQueue.java
taefi: Looking at all types of Signals (no matter a single value holding or more complex ones such as list signals) simplified the early implementation of how we register signal instances to the SignalRegistry and how we receive updates (from client-side code) and dispatch them to the server-side instance of the shared signal. 
But, your point makes complete sense when it comes to the unnecessary overhead of backing single value holding signals with event queues, even if we initialize the size of the backing queue to one.

taefi: It is easy enough to refactor and take out the event queue and the signal queue from this PR.

packages/ts/react-signals/src/Signals.ts
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * A signal that holds a value. The underlying
+ * value of this signal is stored and updated as a
+ * shared value on the server.
+ *
+ * @internal
+ */
+export abstract class ValueSignal<T> extends Signal<T> {
+  private readonly publish: (event: StateEvent) => Promise<boolean>;
```
Lodin, line 12: We shouldn't rely on TS privacy when we have JS privacy
```suggestion
  readonly #publish: (event: StateEvent) => Promise<boolean>;
```

packages/ts/react-signals/src/Signals.ts
```diff
@@ -0,0 +1,78 @@
...
+   * Note that this method is not setting
+   * the signal's value.
+   *
+   * @param value The new value of the signal
+   * to be published to the server.
+   */
+  override set value(value: T) {
+    const id = crypto.randomUUID();
+    const event: SetEvent = { id, type: 'set', value };
+    this.publish(event).then(r => undefined);
```
Lodin, line 44: Why `r => undefined`? Why `then` is necessary here at all?

Lodin, line 44: I'd suggest merging it together as it doesn't bring any value to split them up.
```suggestion
    this.publish({ id, type: 'set', value }).then(r => undefined);
```

packages/ts/react-signals/src/EventChannel.ts
```diff
@@ -0,0 +1,92 @@
...
+      }
+    } else if (snapshotEvent.value !== undefined) {
+      this.internalSignal = this.createInternalSignal((event: StateEvent) => this.publish(event), snapshotEvent.value);
+    }
+  }
+
+  public async publish(event: StateEvent): Promise<boolean> {
+    return this.channelDescriptor.publish(this.id, JSON.stringify(event))
+          .then((_) => { return true; })
+          .catch((error) => { throw Error(error); });
```
Lodin, line 70: Since it's a `async` function, let's use async stuff here.

```suggestion
    try {
      await this.channelDescriptor.publish(this.id, JSON.stringify(event));
      return true;
    } catch (e: unknown) {
      throw Error(e)
    }
```

packages/ts/react-signals/src/EventChannel.ts
```diff
@@ -0,0 +1,92 @@
...
+      subscribe: (signalProviderEndpointMethod: string, signalId: string) => this.signalsHandler.subscribe(signalProviderEndpointMethod, signalId),
+      publish: (signalId: string, event: string) => this.signalsHandler.update(signalId, event),
+    }
+
+    this.internalSignal = this.createInternalSignal((event: StateEvent) => this.publish(event));
+
+    this.connect();
+  }
+
+  private connect() {
```
Lodin, line 47:
```suggestion
  #connect() {
```

packages/ts/react-signals/src/EventChannel.ts
```diff
@@ -0,0 +1,92 @@
...
+ * The signal channel is responsible for subscribing to
+ * the server-side signal and updating the local signal
+ * based on the received events.
+ */
+abstract class SignalChannel<T, S extends Signal = Signal> {
+  private readonly channelDescriptor: SignalChannelDescriptor<string>;
+  private readonly signalsHandler: SignalsHandler;
+  private readonly id: string;
+
+  private internalSignal: ValueSignal<T> | null = null;
```
Lodin, line 31:
```suggestion
  readonly #channelDescriptor: SignalChannelDescriptor<string>;
  readonly #signalsHandler: SignalsHandler;
  readonly #id: string;

  #internalSignal: ValueSignal<T> | null = null;
```

packages/ts/react-signals/src/EventChannel.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  public async publish(event: StateEvent): Promise<boolean> {
+    return this.channelDescriptor.publish(this.id, JSON.stringify(event))
+          .then((_) => { return true; })
+          .catch((error) => { throw Error(error); });
+  }
+
+  getSignal(): S {
+    return this.internalSignal as Signal as S;
```
Lodin, line 74: It looks weird. Shouldn't we have an `S` in the definition from the very start?

Also, I would suggest a getter here: `get signal(): S`

packages/ts/react-signals/src/SignalsHandler.ts
```diff
@@ -0,0 +1,23 @@
+import { ConnectClient, type EndpointRequestInit, type Subscription } from '@vaadin/hilla-frontend';
+
+/**
+ * SignalsHandler is a helper class for handling the
+ * communication of the full-stack signal instances
+ * and their server-side counterparts they are
+ * subscribed and publish their updates to.
+ */
+export default class SignalsHandler {
+  private readonly client: ConnectClient;
```
Lodin, line 10:
```suggestion
  readonly #client: ConnectClient;
```

Lodin (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/core/SignalsRegistry.java
```diff
@@ -0,0 +1,206 @@
...
+     * @param signalId
+     *            the signal id, must not be null
+     * @throws NullPointerException
+     *             if {@code signalId} is null
+     */
+    public synchronized void unregister(UUID signalId) {
+        Objects.requireNonNull(signalId,
+                "Signal id to remove must not be null");
+        signals.remove(signalId);
+        clientSignalToSignalMapping.values().removeIf(signalId::equals);
```
cromoteca, line 135: This is not a complete review, but I feel that checking all values is a potential performance problem.
I propose this alternative:

- remove `clientSignalToSignalMapping` completely and store all mapping in the existing weak map, no matter if the UUIDs come from the client or the server;
- add a `Set` of related known UUIDs to the `Signal` itself;
- each time a mapping is added, also add the UUID to the corresponding `Set` in the `Signal`;
- when a client UUID is removed from the map, also remove it from the corresponding `Set`;
- when a server UUID is removed from the map, clear the `Signal`'s `Set` completely.

So, all keys related to a `Signal` will be preserved as they are referenced in its `Set`, but the server-side removal still allows full garbage collection, as the client UUIDs will only remain as weak keys.

I think that this implementation is far simpler if made by subclassing `Map`.

cromoteca (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/core/SignalsRegistry.java
cromoteca: If you want the server UUID to stay "weak", a second `put` method is also needed, as for removal, to avoid adding that UUID to the `Signal`'s `Set`.

taefi: Thanks for the review! 
TBH, while I was implementing the SignalsRegistry this way, I wasn't thinking about the performance that much as suggested by the [issue's description](https://github.com/vaadin/hilla/issues/2429). But, I had two things in mind: 
1. To easily find out whether a signal instance returned by an endpoint has already been registered to avoid duplicate registration (the weak map used for this purpose).
2. To easily access the instance to which a client is subscribed (`clientSignalToSignalMapping` is used for this purpose). This access is happening every time a client changes the signal's value, so it needs to be fast.

In a subsequent PR, we will enable clients to unsubscribe from a signal instance, e.g. when the user moves away from the view containing a usage (accessing/manipulating) for that signal instance. This will be supported by the registry's `removeClientSignalToSignalMapping` method (which is fast as well). The `unregister` method would not be called so often, for example, as of now it is not even called anywhere, and it is only there for the sake of feature completeness of the registry itself, so this will not be a performance issue for now. In fact, unregistering a server signal instance can/should happen when there are no client is subscribing to them, which means no mappings in the `removeClientSignalToSignalMapping`, and this will let the garbage collection to remove it from the weak map automatically / eventually.

Your suggested implementation is also supporting both concerns mentioned above, however, I find it easier to debug with the this way, where all the mappings are stored in a single class (maybe it is a matter of taste). Maybe we can even remove the `unregister` method, and address that functionality and its possible performance issues later if happens. WDYT?

cromoteca (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/NumberSignal.java
```diff
@@ -0,0 +1,163 @@
...
+        this(0.0);
+    }
+
+    /**
+     * Subscribes to the signal.
+     *
+     * @return a Flux of JSON events
+     */
+    public Flux<ObjectNode> subscribe() {
+        Sinks.Many<ObjectNode> sink = Sinks.many().multicast()
```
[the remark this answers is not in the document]
taefi: Good point, you're right. 
Basically, mulitcast is for having one Flux with multiple subscribers, but we cannot use here, since the behaviour of the Flux when an error happens is to propagate the error to all subscribers, and then to terminate the Flux (no more propagations can happen through that). 
We need a separate Sink (and Flux) for each client, so that when an error happens in propagating the updates to one client, it doesn't terminate the flow of updates to other clients. So, each sink has only one subscriber in practice, and unicast the correct way to do it.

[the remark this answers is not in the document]
taefi: Done.

cromoteca (dismissed):

Lodin (approved):

cromoteca (approved):

PR #2505, opened 2024-06-05
feat(binder): implement `useFormArrayPart`

### LINKED ISSUE #1196: React form array APIs

cromoteca:
Closes #1196.

### DIFF
--- a/packages/ts/react-form/src/index.ts
+++ b/packages/ts/react-form/src/index.ts
@@ -16,6 +16,7 @@ import {
   type Validator,
   type Value,
   type ValueError,
+  type ArrayModel,
 } from '@vaadin/hilla-lit-form';
 import { useEffect, useMemo, useReducer, useRef } from 'react';
 import type { Writable } from 'type-fest';
@@ -71,6 +72,13 @@ export type UseFormResult<M extends AbstractModel> = Omit<UseFormPartResult<M>, 
     read(value: Value<M> | null | undefined): void;
     update(): void;
   }>;
+
+export type UseFormArrayPartResult<M extends ArrayModel, M1 extends AbstractModel = AbstractModel> = Omit<
+  UseFormPartResult<M>,
+  'field'
+> & {
+  items: readonly M1[];
+};
 
 type FieldState<T = unknown> = {
   required: boolean;
@@ -258,4 +266,21 @@ export function useFormPart<M extends AbstractModel>(model: M): UseFormPartResul
     ...getFormPart(binderNode),
     field,
   };
+}
+
+/**
+ * Hook to access an array model part of a form. It provides the same API as `useFormPart`,
+ * but adds an `items` property that allows to iterate over the items in form of an array of models.
+ *
+ * @param model - The array model to access
+ * @returns The array model part of the form
+ */
+export function useFormArrayPart<M extends ArrayModel>(
+  model: M,
+): M extends ArrayModel<infer M1> ? UseFormArrayPartResult<M, M1> : never {
+  const binderNode = getBinderNode(model);
+  return {
+    ...getFormPart(binderNode),
+    items: [...model].map((item) => item.model),
+  } satisfies UseFormArrayPartResult<M> as any;
 }
--- a/packages/ts/react-form/test/useForm.spec.tsx
+++ b/packages/ts/react-form/test/useForm.spec.tsx
@@ -6,7 +6,7 @@ import chaiDom from 'chai-dom';
 import { useEffect, useState } from 'react';
 import sinon from 'sinon';
 import sinonChai from 'sinon-chai';
-import { useForm as _useForm, useFormPart } from '../src/index.js';
+import { useForm as _useForm, useFormArrayPart, useFormPart } from '../src/index.js';
 import {
   type Contract,
   EntityModel,
@@ -401,7 +401,7 @@ describe('@vaadin/hilla-react-form', () => {
   }) {
     const { field, model, read } = useForm(TeamModel);
     const name = useFormPart(model.name);
-    const { value, setValue } = useFormPart(model.players);
+    const { items, value, setValue } = useFormArrayPart(model.players);
 
     useEffect(() => {
       testAction({ read, value, setValue });
@@ -413,8 +413,8 @@ describe('@vaadin/hilla-react-form', () => {
         <output data-testid="validation.team.name">
           {name.invalid ? name.ownErrors.map((e) => e.message).join(', ') : 'OK'}
         </output>
-        {Array.from(model.players, (player) => (
-          <TeamFormPlayer key={`${player.value!.id}`} model={player.model} />
+        {items.map((player, index) => (
+          <TeamFormPlayer key={`${value![index].id || -index}`} model={player} />
         ))}
       </>
     );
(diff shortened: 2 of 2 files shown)


packages/ts/react-form/src/index.ts
```diff
@@ -259,3 +267,20 @@ export function useFormPart<M extends AbstractModel>(model: M): UseFormPartResul
...
+ * @returns The array model part of the form
+ */
+export function useFormArrayPart<M extends ArrayModel>(
+  model: M,
+): M extends ArrayModel<infer M1> ? UseFormArrayPartResult<M, M1> : never {
+  const binderNode = getBinderNode(model);
+  return {
+    ...getFormPart(binderNode),
+    items: [...model].map((item) => item.model),
+  } satisfies UseFormArrayPartResult<M> as any;
```
Lodin, line 285:
```suggestion
  const part = useFormPart(model);
return {
  ...part,
  items: Array.from(model, (item) => item.model),
}
```

packages/ts/react-form/src/index.ts
```diff
@@ -72,6 +73,13 @@ export type UseFormResult<M extends AbstractModel> = Omit<UseFormPartResult<M>,
     update(): void;
   }>;
 
+export type UseFormArrayPartResult<M extends ArrayModel, M1 extends AbstractModel = AbstractModel> = Omit<
+  UseFormPartResult<M>,
+  'field'
+> & {
+  items: readonly M1[];
```
Lodin, line 80: We have an `ArrayItemModel<M>` type that extracts the type from `ArrayModel`:
```suggestion
  items: ReadonlyArray<ArrayItemModel<M>>;
```

packages/ts/react-form/src/index.ts
```diff
@@ -259,3 +267,20 @@ export function useFormPart<M extends AbstractModel>(model: M): UseFormPartResul
...
+/**
+ * Hook to access an array model part of a form. It provides the same API as `useFormPart`,
+ * but adds an `items` property that allows to iterate over the items in form of an array of models.
+ *
+ * @param model - The array model to access
+ * @returns The array model part of the form
+ */
+export function useFormArrayPart<M extends ArrayModel>(
+  model: M,
+): M extends ArrayModel<infer M1> ? UseFormArrayPartResult<M, M1> : never {
```
Lodin, line 280: With the suggestion above, you don't need to use `extends` here:
```suggestion
): UseFormArrayPartResult<M> {
```

Lodin (changes requested):

packages/ts/react-form/test/useForm.spec.tsx
```diff
@@ -442,19 +442,23 @@ describe('@vaadin/hilla-react-form', () => {
     });
 
     it('should add item to empty array', async () => {
+      let val: Player[] | undefined;
+      let setVal: ((value: Player[] | undefined) => void) | undefined;
       const { findByTestId } = render(
         <TeamForm
           testAction={({ read, value, setValue }) => {
```
Lodin, line 449: BTW, instead of `read` here you could probably use `setValue` for the array only. I guess, it'll make SonarQube happy

cromoteca: I try to do something vaguely similar to what would happen in a real application. Honestly, I don't care that much about code repetition in tests: in some cases, it's even a benefit as it makes a single test easier to read.

Lodin: Ok, but don't forget to go to the SonarCloud and mark it as an intentional change (it allows doing it if you login).

cromoteca: I changed the method I set up the form in the test so that code repetition is no longer an issue. Now I have a weird Codecov report.

packages/ts/react-form/src/index.ts
```diff
@@ -259,3 +265,18 @@ export function useFormPart<M extends AbstractModel>(model: M): UseFormPartResul
...
+ *
+ * @param model - The array model to access
+ * @returns The array model part of the form
+ */
+export function useFormArrayPart<M extends ArrayModel>(model: M): UseFormArrayPartResult<M> {
+  const binderNode = getBinderNode(model);
+  return {
+    ...getFormPart(binderNode),
+    items: Array.from(model, (item) => item.model as ArrayItemModel<M>),
+  } satisfies UseFormArrayPartResult<M> as any;
```
Lodin, line 281: I guess it's possible to remove this cast now since `extends` in the return type is gone
```suggestion
  }
```

packages/ts/react-form/test/useForm.spec.tsx
```diff
@@ -373,131 +374,84 @@ describe('@vaadin/hilla-react-form', () => {
...
-    }): void;
-  }) {
-    const { field, model, read } = useForm(TeamModel);
-    const name = useFormPart(model.name);
-    const { value, setValue } = useFormPart(model.players);
-
-    useEffect(() => {
-      testAction({ read, value, setValue });
-    }, []);
+      players = useSignal(initialPlayers);
```
Lodin, line 398: Do we really need a signal here? 🤔

Lodin (changes requested):
LGTM overall

platosha (approved):

packages/ts/react-form/test/useForm.spec.tsx
cromoteca: No, it was just a way to simplify code, but I'm open to alternatives. It allows me to react to further changes to the player array.

Lodin (approved):

PR #2508, opened 2024-06-06
fix(form): initialize value / defaultValue keys when rendering optional fields

### LINKED ISSUE #2339: [Binder] Form validation problem with optional fields and custom validator on object level
### Describe the bug

Using binder with a model having two optional fields and a custom validator on the object level (instead of field level) e.g.

```
this.binder.addValidator({
     message: 'Error',
    validate(value, binder) {
        const result: Array<ValidationResult> = [];
        if (value.startTime && value.stopTime && (value.startTime >= value.stopTime)) {
            result.push({
                message: 'Must be after start time',
                property: model.stopTime,
            });
        }
        return result;
    }
});
```

the validator is only called for the first time the field gets blurred and the invalid state of the field doesn't get updated on future modifications.

The bug can be "fixed" if you modify the binders _defaultValue_ and add a value for the field which can get invalid, which causes BinderNode.getChildBinderNodes() to emit this child node in case of validation update. But you have to remember to adjust binders _defaultValue_ every time after the defaultValues gets modified e.g. binder.read()

### Expected-behavior

I expect that the invalid status (and error message) gets updated every time the field is changed by the user

### Reproduction

Here is a modified starter project which demonstrates the bug. Just follow the instructions in the browser.
The interesting code changes are only in frontend/views/main-view.ts

[skeleton-starter.zip](https://github.com/vaadin/hilla/files/15023473/skeleton-starter.zip)

### System Info

Hilla 2.5.7

platosha:
Fixes #2339

### DIFF
--- a/packages/ts/form/src/BinderNode.ts
+++ b/packages/ts/form/src/BinderNode.ts
@@ -257,18 +257,13 @@ export class BinderNode<M extends AbstractModel = AbstractModel> extends EventTa
       return undefined;
     }
 
-    let { value } = this.parent;
-
-    if (value === undefined) {
-      this.parent.initializeValue(true);
-      ({ value } = this.parent);
-    }
+    this.initializeValue();
 
     const key = this.model[_key];
 
     // The value of parent in unknown, so we need to cast it.
     type ParentValue = { readonly [K in typeof key]: Value<M> };
-    return (value as ParentValue)[key];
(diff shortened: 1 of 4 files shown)


Lodin (approved):

Lodin:
The base branch does not allow updates? 😦

platosha:
> The base branch does not allow updates? 😦

Fixed.

PR #2522, opened 2024-06-07
refactor(lit-form): deprecate AbstractModel methods that require BinderNode

### LINKED ISSUE #2060: [lit-form] make model API not depend on form binder

Lodin:
Resolves #2060.


platosha (approved):

platosha (changes requested):

packages/ts/lit-form/src/Models.ts
```diff
@@ -213,40 +230,54 @@ export class ArrayModel<MItem extends AbstractModel = AbstractModel> extends Abs
...
+   * @deprecated Use the {@link getIterator} function instead.
+   */
+  *[Symbol.iterator](): IterableIterator<BinderNode<MItem>> {
+    for (const item of this[_items]()) {
       yield getBinderNode(item);
     }
   }
 }
+
+export function getIterator<M extends ArrayModel>(model: M): Generator<ArrayItemModel<M>, void, void> {
```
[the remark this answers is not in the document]
Lodin: Hmm, I usually expect to have an array by the function named this way. Could also be confusing... Maybe, `getItemsIterator` / `getItemsIter`? Or `iterateItems`?

packages/ts/lit-form/src/Models.ts
```diff
@@ -213,39 +230,53 @@ export class ArrayModel<MItem extends AbstractModel = AbstractModel> extends Abs
...
+  }
+
+  /**
+   * Iterates over the current model and yields a binder node for every item
+   * model.
+   *
+   * @deprecated Use the {@link m.items} function instead. E.g.:
+   * ```ts
+   * const itemValues = Array.from(m.items(model), (itemModel) => m.value(itemModel));
+   * ```
```
platosha, line 276: `m.value` is not available yet :-)

Maybe it's fine to use some use-case specific examples for Lit and React:

```suggestion
   * @deprecated Use the {@link m.items} function instead. For example, in React:
   * ```tsx
   * const {model, field} = useForm(GroupModel);
   * return Array.from(m.items(model.people), (personModel) => 
   *   <TextField label="Full name" {...field(personModel.fullName)} />
   * );
   * ```
   * In Lit:
   * ```ts
   * return html`${repeat(m.items(this.binder.model.people), personModel => html`
   *   <vaadin-text-field label="Full name" ${field(personModel.fullName)}></vaadin-text-field>
   *  `)}
   * `;
```

packages/ts/lit-form/src/index.ts
```diff
@@ -7,6 +9,20 @@ export * from './Validation.js';
...
+   *
+   * @param model - The array model to iterate over.
+   * @returns An iterator over item models.
+   */
+  items<M extends ArrayModel>(model: M): Generator<ArrayItemModel<M>, void, void> {
+    return model[_items]() as Generator<ArrayItemModel<M>, void, void>;
+  },
+};
+
+export default m;
```
platosha, line 24: Let's not use default export from the package index here — it's the Form library, not the models library. Does not feel semantically correct to export the model utils as the default asset here.

platosha (changes requested):

packages/ts/lit-form/src/index.ts
Lodin: Ah, right, didn't think about that. Thanks

cromoteca (approved):

PR #2532, opened 2024-06-11
chore: add parent project for hilla-bom

ZheSun88:
this seems has been missing from the beginning of our project.  without it, there is missing deploy settings to maven central


packages/java/hilla-bom/pom.xml
```diff
@@ -4,6 +4,13 @@
...
     <modelVersion>4.0.0</modelVersion>
 
+    <parent>
+        <groupId>com.vaadin</groupId>
+        <artifactId>hilla-project</artifactId>
+        <version>24.5-SNAPSHOT</version>
+        <relativePath>../../../pom.xml</relativePath>
+    </parent>
+
     <groupId>com.vaadin</groupId>
```
Artur-, line 14:
```suggestion
```

ZheSun88: i am going to check if there are other modules has the duplicated version.. 
for example https://github.com/vaadin/hilla/blob/main/packages/java/hilla/pom.xml#L18

but they are not a blocker, i think

PR #2537, opened 2024-06-13
feat: add support for JsonNode subclasses

cromoteca:
Support for `JsonNode` was added recently. This PR adds the same treatment for its subclasses.


platosha:
Just wondering: should we rather represent JsonObject as `Record<never, never>` and JsonArray as `unknown[]`?

cromoteca:
I decided to only support the 3 non-raw-value classes, which also removes the need to use ClassGraph for discovery.

Lodin (approved):

PR #2541, opened 2024-06-13
chore: use vaadin-parent 2.2.1 as parent for hilla-bom

ZheSun88:
using hilla-project brings the `<dependencyManangement>` from parent, which is not the one want

### DIFF
--- a/packages/java/hilla-bom/pom.xml
+++ b/packages/java/hilla-bom/pom.xml
@@ -13,6 +13,7 @@
     <artifactId>hilla-bom</artifactId>
     <packaging>pom</packaging>
     <name>Hilla Platform (Bill of Materials)</name>
+    <version>24.5-SNAPSHOT</version>
     <description>Hilla Platform (Bill of Materials)</description>
     <url>https://hilla.dev</url>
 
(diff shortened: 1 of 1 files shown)


Artur- (approved):

packages/java/hilla-bom/pom.xml
```diff
@@ -6,9 +6,8 @@
 
     <parent>
         <groupId>com.vaadin</groupId>
-        <artifactId>hilla-project</artifactId>
-        <version>24.5-SNAPSHOT</version>
-        <relativePath>../../../pom.xml</relativePath>
+        <artifactId>vaadin-parent</artifactId>
+        <version>2.2.1</version>
     </parent>
```
Artur-, line 11: You need to add the Hilla version though

Artur- (changes requested):
Version cannot be 2.2.1

packages/java/hilla-bom/pom.xml
ZheSun88: thanks, added

ZheSun88:
same fix has been added to 24.4 branch via #2542

PR #2576, opened 2024-06-25
refactor: Do not use installAutoSignalTracking

Artur-:


Artur-:
Depends on https://github.com/vaadin/flow/pull/19639

Artur-:
With the Flow PR, at least `react-i18n` and `react-signals` pass locally

Artur-:
Maybe there is a better way with Karma to add
```
window.$RefreshReg$ = () => {}
window.$RefreshSig$ = () => (type) => type
window.__vite_plugin_react_preamble_installed__ = true
```

The "correct" way would be to pass the HTML files through Vite but I don't see how that would be possible

platosha (commented):
Using custom Karma HTML is OK to me. @Lodin WDYT?

Lodin (approved):
Looks a bit hacky to me but let's have it this way. I hope we will be able to solve it easier in #2480

PR #2597, opened 2024-07-03
fix: Set routes as a global so Flow can pick them up

Artur-:


Artur-:
What is needed here to make `window` work in tests?

PR #2610, opened 2024-07-17
feat(maven-plugin) add m2e metadata

cromoteca:
Adds `lifecycle-mapping-metadata.xml` like in Flow plugin, but using `generate` as incremental goal.

Should fix both https://github.com/vaadin/platform/issues/6562 and https://github.com/vaadin/platform/issues/6590.


mcollovati (approved):
Tested with Eclipse, and it behaves like the older `vaadin-maven-plugin` versions

PR #2630, opened 2024-07-26
chore: fix prettier error in autogrid-columns.tsx

taefi:


Lodin (approved):

Lodin:
The issue was in the dependency update. So, this PR is no longer necessary

PR #2632, opened 2024-07-26
test: add NumberSignal IT

### LINKED ISSUE #2629: [Full-stack Signals] add integration tests
### Describe your motivation

As a follow-up for #2502, there should be integration-tests to check Full-stack signals are properly working end-to-end, and updates made by one client is automatically received on the other clients.

_No response_

_No response_

### Additional context

_No response_

taefi:
Fixes #2629


packages/java/tests/spring/react-signals/frontend/helper/NumberSignalServiceWrapper.ts
```diff
@@ -0,0 +1,14 @@
+import { NumberSignalChannel } from '@vaadin/hilla-react-signals';
+import client_1 from 'Frontend/generated/connect-client.default.js';
+
+export class NumberSignalServiceWrapper {
```
Lodin, line 4: I would suggest converting this code from a class with static methods to exported functions, so you can import them using `*`:

```ts
import * as NumberSignalServiceWrapper from 'Frontend/helper/NumberSignalServiceWrapper.ts';
```

packages/java/tests/spring/react-signals/frontend/routes.tsx
```diff
@@ -1,6 +1,7 @@
 import { createBrowserRouter } from 'react-router-dom';
 import MainLayout from './MainLayout.js';
 import BasicSignalView from './_views/BasicSignalView.js';
+import SharedNumberSignal from './_views/SharedNumberSignal.js';
```
Lodin, line 4: I guess, we can already rename it: `_views` -> `views`. The problem why we had to rename it has already gone

Lodin (changes requested):
LGTM but I would like to transform the class with static methods to a list of functions since it's gonna be the final implementation

packages/java/tests/spring/react-signals/frontend/helper/NumberSignalServiceWrapper.ts
taefi: > ... class with static methods to a list of functions since it's gonna be the final implementation

I'm a bit confused about the _"it's gonna be the final implementation"_, since I was picturing the generator will generate client-side services in a way that this helper service wrapper will not be needed at all, so this is temporary until we add the generator plugin for endpoints methods returning a signal instance.

Lodin (approved):

PR #2633, opened 2024-07-29
feat: allow setting signal's value before server confirmation

### LINKED ISSUE #2624: [Full-stack Signals] allow client-side signal's value to be set before confirmed by server
### Describe your motivation

It is important to let the value of the client-side instance of a signal to be set upon any change made on the client-side before waiting for the confirmed value comes back from the server as an update. This is needed for the future improvements such as supporting latency compensation and offline support.

_No response_

_No response_

### Additional context

_No response_

taefi:
## Description
This allows the value of the signal to be set
without waiting for update of the confirmed
value from server. When an update of the
confirmed value comes from the server, the
value will be updated accordingly as before.

Fixes #2624

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


packages/ts/react-signals/test/Signals.spec.tsx
```diff
@@ -59,5 +59,16 @@ describe('@vaadin/hilla-react-signals', () => {
       await nextFrame();
       expect(result.container.textContent).to.equal('Value is 42');
     });
+
+    it('should set local value when signal value is updated', () => {
+      const numberSignal = new NumberSignal(publishSpy);
+      expect(numberSignal.value).to.equal(undefined);
+      numberSignal.value = 42;
+      expect(numberSignal.value).to.equal(42);
```
Legioth, line 67: Should maybe also test that an `effect(() => doSomething(numberSignal.value))` is run already before the server has confirmed the change

packages/ts/react-signals/src/Signals.ts
```diff
@@ -47,6 +47,9 @@ export abstract class ValueSignal<T> extends Signal<T> {
    */
   override set value(value: T) {
     const id = crypto.randomUUID();
+    // first set the local value:
```
Lodin, line 50: Could you describe in this comment why do we need it?

packages/ts/react-signals/test/Signals.spec.tsx
```diff
@@ -59,5 +60,27 @@ describe('@vaadin/hilla-react-signals', () => {
...
+      effect(() => {
+        if (counter === 0) {
+          expect(anotherNumberSignal.value).to.be.undefined;
+        } else if (counter === 1) {
+          expect(anotherNumberSignal.value).to.equal(42);
+        } else {
+          expect(anotherNumberSignal.value).to.equal(43);
+        }
+        counter += 1;
+      });
```
Lodin, line 81: I'm not sure if this code is evaluated because the test is probably over before the effect starts working. You have to add an expectation in the end of the test and outside of any callback. You could use an array or a map to collect data, and in the end of the test you can assert it.

Lodin (changes requested):

packages/ts/react-signals/src/Signals.ts
taefi: This was more of reminder about the order of the operations, so it is not changed to first publish the value to the server and then set the local value by accident (in the future), as I couldn't think of a clear way of verifying that through a test.

packages/ts/react-signals/test/Signals.spec.tsx
taefi: I know that the effect is called since if you change the expected values it will fail, but I see your point ;) 
I'll simply expect the counter's value to be 2 by the end of test, that should be enough I guess.

packages/ts/react-signals/src/Signals.ts
taefi: Added the reason in the comment.

PR #2639, opened 2024-08-01
feat: add generator plugin for endpoints returning signals

### LINKED ISSUE #2628: [Full-stack Signals] hilla-generator should generate TS codes for endpoints exposing signals to eliminate the need for manual service wrappers
### Describe your motivation

After the first implementation of Full-stack Signals, a manually written TS code in user projects is needed for obtaining the Signal instance exposed by an endpoint. For an endpoint as follows:
```java
import com.vaadin.hilla.signals.NumberSignal;
import com.vaadin.hilla.BrowserCallable;

@AnonymousAllowed
@BrowserCallable
public class SharedCounterService {
    private final NumberSignal counter = new NumberSignal(0);

    public NumberSignal counter() {
        return counter;
    }
}
```
The needed manually written code is like this:
```ts
import { NumberSignalQueue } from "@vaadin/hilla-react-signals";
import client_1 from "Frontend/generated/connect-client.default";

export class CounterServiceWrapper {
  static counter() {
    const signalChannel = new NumberSignalChannel('SharedCounterService.counter', client_1);
    return signalChannel.signal;
  }
}
```

Hilla generator should be able to generate the `SharedCounterService.ts` in a way that it doesn't need a separate wrapper for using signals. As the result. the user codes that uses a full-stack signal should look like this:
```ts
const counter = SharedCounterService.counter();

export default function NumberSignalView() {
  return (
    <Button onClick={() => counter.value++}>
       Click count: { counter }
    </Button>
  );
}
```

_No response_

### Additional context

_No response_

taefi:
## Description

The generator-plugin-signals enables the generator 
to transform implementation of generated endpoints 
containing methods that return a signal instances in 
a way to create and return the correct signal instance
instead of Promise.

The added plugin ensures that an endpoint containing
a mixture of methods returning signal instances and 
other supported types has correct implementation for
each method. It also ensures the implementation of
endpoints without any methods returning a signal 
remains unchanged.

Fixes #2628

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


packages/ts/generator-plugin-signals/src/SignalProcessor.ts
```diff
@@ -0,0 +1,132 @@
...
+        imports.named.add(HILLA_REACT_SIGNALS, id.text, true, id);
+      }
+    });
+  }
+
+  #removeUnusedRequestInitImports(file: SourceFile) {
+    const transformedFileText = ts.createPrinter().printFile(file);
+
+    const hasNormalEndpointCalls =
+      transformedFileText.includes('init?: EndpointRequestInit') && transformedFileText.includes('.call(');
```
Lodin, line 126: Do I understand correctly that you're checking if the `init` is used here?

taefi: Yeah, that's the idea, since our generated services with calls to normal endpoints will have `init?: EndpointRequestInit_1` (or some other _N), but services without normal endpoint, will not have methods with that parameter after transformation.

Lodin: Maybe, it'd be better to check the function's argument to be used in the function body? Something like

```ts
transform((node) => {
  if (
    ts.isFunctionDeclaration(node) &&
    node.parameters.some(({ name }) => ts.isIdentifier(name) && name.text === 'init') &&
    node.body
  ) {
    let hasInit = false;

    traverse(node.body, (n) => {
      if (ts.isIdentifier(n) && n.text === 'init') {
        hasInit = true;
      }
    });

    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
    return hasInit
      ? node
      : ts.factory.createFunctionDeclaration(
          node.modifiers,
          node.asteriskToken,
          node.name,
          node.typeParameters,
          node.parameters.filter(({ name }) => !(ts.isIdentifier(name) && name.text === 'init')),
          node.type,
          node.body,
        );
  }

  return node;
}),
```
Where `traverse` function should be defined in `generator-utils/ast.ts` as:
```ts
export function traverse(root: Node, visitor: (node: Node) => void): void {
  const _visitor = (node: Node) => {
    visitor(node);
    ts.forEachChild(node, _visitor);
  };

  ts.forEachChild(root, _visitor);
}
```

taefi: Well, if I understand this correctly, this is a nice improvement for the part I've just passed `[]` (an empty array) to the `createFunctionDeclaration` as the `parameters`, correct? Ofc, I can use it with another variable scoped outside of this transformer to detect whether the `init` is used in any of the functions or not.

Lodin: ah, if you don't need any parameters, then sure. I just focused on removing the `init`one.

taefi: Since the methods that are not returning signal instances are not transformed by this plugin, it's easy just to exclude the `init` parameter for the methods returning the signal instance, as we are sure that is not used in those method bodies.

packages/ts/generator-plugin-signals/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+export type PathSignalType = Readonly<{
+  path: string;
+  signalType: string;
+}>;
+
+const SIGNAL_CLASSES = ['#/components/schemas/com.vaadin.hilla.signals.NumberSignal'];
+
+function extractEndpointMethodsWithSignalsAsReturnType(storage: SharedStorage): PathSignalType[] {
+  const pathSignalTypes: PathSignalType[] = [];
+  Object.entries(storage.api.paths).forEach(([path, pathObject]) => {
```
Lodin, line 14: I guess, you could replace it with `map` with `filter` approach or `reduce` to simplify it

packages/ts/generator-plugin-signals/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+      }
+    }
+  });
+  return pathSignalTypes;
+}
+
+function groupByService(signals: PathSignalType[]): Map<string, MethodInfo[]> {
+  const serviceMap = new Map<string, MethodInfo[]>();
+
+  signals.forEach((signal) => {
```
Lodin, line 37: `reduce` approach would be the best here

packages/ts/generator-utils/src/dependencies/ImportManager.ts
```diff
@@ -138,6 +150,21 @@ export class DefaultImportManager extends StatementRecordManager<ImportDeclarati
     return this.#map.get(path)?.id;
   }
 
+  remove(path: string): void {
+    if (this.#map.has(path)) {
+      this.#map.delete(path);
+    }
+  }
+
+  find(predicate: (path: string) => boolean): [string, boolean, Identifier] | undefined {
```
Lodin, line 159: The import managers are designed the same in terms of API, so we should either add `find` method to every manager or use `identifiers` generator instead (it looks kinda broken right now, but we can fix it I guess)

packages/ts/generator-plugin-signals/src/SignalProcessor.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    const [file] = ts.transform<SourceFile>(this.#sourceFile, [
+      ...this.#methods.map((method) =>
+        transform<SourceFile>((tsNode) => {
+          if (ts.isFunctionDeclaration(tsNode) && tsNode.name?.text === method.name) {
+            const endpointAndMethodName = ts.factory.createStringLiteral(`${this.#service}.${method.name}`);
+            const body = template(
+              `
+function dummy() {
+  return new ${NUMBER_SIGNAL_CHANNEL}(${ENDPOINT_AND_METHOD_NAME}, ${CONNECT_CLIENT}).signal;
```
Lodin, line 54: Let's use the text wherever we can. E.g.:
```suggestion
return new ${NUMBER_SIGNAL_CHANNEL}('${this.#service}.${method.name}', ${CONNECT_CLIENT}).signal;
```

packages/ts/generator-plugin-signals/src/SignalProcessor.ts
```diff
@@ -0,0 +1,132 @@
...
+              ],
+            );
+
+            let returnType = tsNode.type;
+            if (
+              returnType &&
+              ts.isTypeReferenceNode(returnType) &&
+              'text' in returnType.typeName &&
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
+              returnType.typeName.escapedText === 'Promise'
```
Lodin, line 74: This variant does not throw a eslint error
```suggestion
              returnType.typeName.text === 'Promise'
```

packages/ts/generator-plugin-signals/src/SignalProcessor.ts
```diff
@@ -0,0 +1,132 @@
...
+    return createSourceFile(
+      [
+        ...this.#dependencyManager.imports.toCode(),
+        ...file.statements.filter((statement) => !ts.isImportDeclaration(statement)),
+      ],
+      file.fileName,
+    );
+  }
+
+  #processSignalImports(signalImports: string[]) {
```
Lodin, line 108:
```suggestion
  #processSignalImports(signalImports: readonly string[]) {
```

Lodin (changes requested):
LGTM overall but there is some places to fix

packages/ts/generator-utils/src/dependencies/ImportManager.ts
taefi: Added the find and remove methods the managers that were missing it.

packages/ts/generator-plugin-signals/src/SignalProcessor.ts
```diff
@@ -0,0 +1,132 @@
...
+              ts.factory.createBlock(body ?? [], false),
+            );
+          }
+
+          return tsNode;
+        }),
+      ),
+    ]).transformed;
+
+    this.#removeUnusedRequestInitImports(file);
```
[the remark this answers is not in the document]
taefi: Done.

packages/ts/generator-utils/src/dependencies/ImportManager.ts
```diff
@@ -93,6 +116,21 @@ export class NamespaceImportManager extends StatementRecordManager<ImportDeclara
     return id;
   }
 
+  remove(path: string): void {
+    if (this.#map.has(path)) {
+      this.#map.delete(path);
+    }
+  }
+
+  find(predicate: (id: Identifier) => boolean): string | undefined {
```
Lodin, line 125: Hmm, after a second look, I'm not sure if it's a good idea to have this method at all. I mean, it's pretty hard to say what exactly we are looking for using this method. E.g., here you get an `Identifier` in a search function, and in `NamedImportManager` you get a `path` and `specifier`.

So, what if we remove these methods and use an `identifier` generator to look for a specific thing right in the code?

Lodin (changes requested):

packages/ts/generator-plugin-signals/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+        ? responseSchema.anyOf
+            ?.filter((c) => '$ref' in c && c.$ref && SIGNAL_CLASSES.includes(c.$ref))
+            .map((c: OpenAPIV3.ReferenceObject | OpenAPIV3.SchemaObject) => ({
+              path,
+              signalType: '$ref' in c ? c.$ref : '',
+            }))
+        : [];
+    })
+    .filter((signalArray) => signalArray !== undefined)
+    .reduce<PathSignalType[]>((acc, current) => acc.concat(current), []);
```
Lodin, line 33: You already have an array, `reduce` is unnecessary here

Lodin (changes requested):

Lodin (approved):

PR #2646, opened 2024-08-08
refactor: use nanoid instead of crypto.randomUUID

taefi:
We do not need security-grade random ids neither
for the events being exchanged between client and
server, nor for the client-side signal instances.
Since `crypto.randomUUID()` was creating problems
for executing tests on hub (with no https connection), 
it is replaced with the nanoid npm package.

This also refactor the server-side APIs to replace UUID
type for client-side signal ids to be string, and keeps 
using UUID only for the server-side signal ids.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18913,7 +18913,8 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@preact/signals-react": "^2.0.0",
-        "@vaadin/hilla-frontend": "24.5.0-alpha7"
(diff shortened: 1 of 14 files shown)


Lodin:
Do we really need to replace `UUID` with `String` on the Java side?

taefi:
> Do we really need to replace `UUID` with `String` on the Java side?

Yes, since ids generated by the nanoid are not compatible with the UUID format, so java api that are dealing with those need to change.

Lodin (approved):

PR #2649, opened 2024-08-09
feat: add support for Daily Active User feature

mcollovati:
Adds support for Daily Active User feature for DAU enabled production build artifacts.


mcollovati:
This is a draft PR, just to share a potential solution for DAU integration in Hilla and start a discussion.
Currently, it won't compile because it requires vaadin/flow#19771

packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointController.java
```diff
@@ -196,13 +214,22 @@ public ResponseEntity<String> serveEndpoint(
...
-            // Put a VaadinRequest in the instances object so as the request is
-            // available in the endpoint method
-            VaadinServletService service = (VaadinServletService) VaadinService
-                    .getCurrent();
-            CurrentInstance.set(VaadinRequest.class,
-                    new VaadinServletRequest(request, service));
+            if (service != null) {
+                DAUUtils.TrackableOperation.INSTANCE.execute(() -> {
+                    service.requestStart(vaadinRequest, vaadinResponse);
+                    DAUUtils.applyEnforcement(vaadinRequest, unused -> true);
```
mcollovati, line 230: Once the Flow PR will be merged, this should be changed int this way
```java
    if (DAUUtils.isDauEnabled(service)) {
        DAUUtils.applyEnforcement(vaadinRequest, unused -> true)
    }
```

mshabarov: Done.

packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointController.java
```diff
@@ -196,13 +241,13 @@ public ResponseEntity<String> serveEndpoint(
...
-            // available in the endpoint method
-            VaadinServletService service = (VaadinServletService) VaadinService
-                    .getCurrent();
-            CurrentInstance.set(VaadinRequest.class,
-                    new VaadinServletRequest(request, service));
+            enforcementResult = DAUUtils.trackDAU(this.vaadinService, request,
+                    response);
+            if (enforcementResult.isEnforcementNeeded()) {
+                return buildEnforcementResponseEntity(enforcementResult);
+            }
```
mshabarov, line 250: Does it make sense to run this block only if `DAUUtils.isDauEnabled(service)` is true?

mshabarov (dismissed):

packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointController.java
tltv: `VaadinRequest` wouldn't be set to `CurrentInstance` then and cleared afterwards. `DAUUtils.trackDAU` does that now. It was set there originally, so I would assume something to break in Hilla without it.

platosha (approved):

PR #2651, opened 2024-08-11
fix: enable security for SignalsHandler

### LINKED ISSUE #2642: [Full-stack Signals] subscribe and update calls to SignalsHandler should be secured
### Describe the bug

When acquiring an instance of a full-stack signal in the client-side code, the call made to the endpoint that exposes a shared value as a signal instance can have the same method security similar to any other endpoint methods in Hilla framework. However, the subsequent calls to a signal instance is not secured. These subsequent calls are handled by an automatically registered endpoint called SignalsHandler that doesn’t apply any security checks on the request (as of now).

### Expected-behavior

Developers should be able to restrict access to shared signal instance (on the server). In other words, It should be possible to restrict access to a shared signal instance, for instance, the same shared signal instance:
- might be unavailable at all for anonymous users
- could have read-only access for anonymous users or users with insufficient privileges
- could be read and manipulated by users with sufficient privileges

### Reproduction

Currently, the developers can define these fine-grained rules on an endpoint method that exposes a (shared) signal instance. But, the `SignalsHandler` is not applying them when `subscribe` or `update` methods are getting called from client-side.

taefi:
## Description

All the requests for subscribing to a
signal instance returned by an endpoint
as well as sending updates to a signal
instance is being checked for having the
correct permissions by the user requesting it.

Fixes #2642

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


packages/java/tests/spring/react-signals/src/test/java/com/vaadin/hilla/test/SignalBasicSecurityIT.java
```diff
@@ -0,0 +1,162 @@
...
+        loginAs("user");
+
+        Assert.assertEquals(20, fetchCounterValue("User")); // shows server and
+                                                            // related endpoint
+                                                            // are working
+        Assert.assertEquals(20, getCounterSignalValue("userCounter")); // user
+                                                                       // counter
+                                                                       // should
+                                                                       // be
+                                                                       // accessible
```
platosha, line 63: Seems like the end-line comments aren't playing nicely with autoformatter:

```suggestion
        // Shows server and related endpoint are working:
        Assert.assertEquals(20, fetchCounterValue("User"));

        // User counter should be accessible:
        Assert.assertEquals(20, getCounterSignalValue("userCounter")); 
```

platosha: Also below in the same file.

platosha (approved):
Looks good, the formatting aside.

packages/java/tests/spring/react-signals/src/test/java/com/vaadin/hilla/test/SignalBasicSecurityIT.java
taefi: To avoid extra builds and re-review requests, I will merge this as it is, and will fix it on the next PR.

PR #2657, opened 2024-08-15
refactor(react-signals): simplify the implementation

Lodin:


taefi (commented):
Apart from some points I didn't fully understand (mentioned down there), In general, I like the way things got simpler. However, there are things I don't like:
In the currently existing implementation (main branch), the signal was isolated, while now it know about an external thing called channel and how it connects to the server and etc. SignalChannel also knows about the signal (ValueSignal), so now they are very coupled together.

packages/ts/react-signals/src/SignalChannel.ts
```diff
@@ -0,0 +1,100 @@
...
+  #dependencies: readonly ValueSignal[] = [];
+  #subscription?: Subscription<StateEvent>;
+
+  /**
+   * @param client - The client instance to be used for communication.
+   * @param signalProviderEndpointMethod - The method name of the signal provider
+   * service.
+   * @returns The signal channel instance.
+   */
+  constructor(client: ConnectClient, signalProviderEndpointMethod: string) {
```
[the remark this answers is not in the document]
Lodin: Ok, that sounds nice. I have updated it, could you please check?

packages/ts/react-signals/src/FullStackSignal.ts
```diff
@@ -0,0 +1,167 @@
...
+    this.#subscription ??= client.subscribe(ENDPOINT, 'subscribe', {
+      providerEndpoint: endpoint,
+      providerMethod: method,
+      clientSignalId: this.#id,
+    });
+
+    return this.#subscription;
+  }
+
+  async call(event: SubscriptionEvent<T>): Promise<void> {
```
taefi, line 77: I suggest renaming `call` to `update`: 
```suggestion
  async update(event: SubscriptionEvent<T>): Promise<void> {
```

packages/ts/react-signals/src/FullStackSignal.ts
```diff
@@ -0,0 +1,167 @@
...
+import { nanoid } from 'nanoid';
+import { Signal } from './core.js';
+
+const ENDPOINT = 'SignalsHandler';
+
+type SubscriptionEvent<T> = Readonly<{
+  id: string;
+  type: StateEventType;
+  value: T;
+}>;
```
taefi, line 12: I don't understand what's the use case for this type. 
It seems to be an unnecessary duplicate of the `StateEvent` that can be dropped.

packages/ts/react-signals/src/FullStackSignal.ts
taefi: Also, the `SubscriptionEvent` should change to `StateEvent`, here we not actually subscribing anything, but sending `SET` events.

packages/ts/react-signals/src/FullStackSignal.ts
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * @internal
+ */
+export abstract class FullStackSignal<T> extends Signal<T> {
+  readonly id = nanoid();
+  readonly #pending = signal(false);
+  readonly #error = signal<Error | undefined>(undefined);
+  readonly #manager: SubscriptionManager<T>;
+
+  constructor(value: T | undefined, data: ConnectionData) {
```
taefi, line 104: Naming of the second parameter variable and type (`data: ConnectionData`) seems to be a bit unfortunate, specially when it is used alongside with `value` (first param). I suggest changing it to something that announces its purpose, such as `connectionConfig` or something else.

packages/ts/react-signals/src/FullStackSignal.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  /**
+   * Defines whether the signal has an error.
+   */
+  get error(): Error | undefined {
+    return this.#error.value;
+  }
+
+  get updating(): Promise<void> {
```
taefi, line 150: Missing TS Docs.

taefi (changes requested):
The followings seems to be missing:
* Tests for `SubscriptionManager`'s `cancel` and `subscription` public methods.
* Tests for `FullStackSignal`'s `pending`, `error`, `updating` public methods.

And, a misbehaviour is now happening at runtime, that is an update is sent to the server before even any call for subscription is made to the server. It is trying to set a `null` value on server-side instance, but since no subscription exists yet for the targeted signal instance, it simply fails with `IllegalStateException` causing unnecessary buzz in the console.

Other details are commented on the code:

packages/ts/react-signals/src/FullStackSignal.ts
taefi: TBH, I didn't fully understand the purpose of having both `updating` and `pending`. Shouldn't the users write the same code (written in `updating` function) for using the `updating` itself?

Legioth: And please read the API document and suggest suggestions rather than just making up things in the implementation. The API document doesn't define any `updating` property and the `pending` property is typed as `ReadonlySignal<boolean>` so that any dependency on it will have an explicit `.value` in application code to show that a signal value is being read.

Lodin: Fixed all the issues here, now it should be in order with the document

taefi (approved):

PR #2662, opened 2024-08-19
feat: Flow main layout

caalador:
add `withLayouts` method to RouterConfigurationBuilder
for using Flow serverside view as main layout.
To enable  call `.withLayouts(Flow)` in `routes.tsx`
and add `"layout": true` into the views `ViewConfig`.


packages/ts/file-router/src/types.d.ts
```diff
@@ -25,6 +25,12 @@ export type ViewConfig = Readonly<{
    */
   route?: string;
 
+  /**
+   * Allows giving marking view as using server side parent layout
```
taefi, line 29:
```suggestion
   * Set to true to indicate that the view as using server side parent layout
```

caalador: That is already possible, but the linking is done on the server so it is not needed as information on the client.

taefi (changes requested):
Generally, looks fine, added a couple of nit picking comments ;)

packages/ts/file-router/src/types.d.ts
```diff
@@ -25,6 +25,12 @@ export type ViewConfig = Readonly<{
    */
   route?: string;
 
+  /**
+   * Set to true to indicate that the view as using server side parent layout
```
taefi, line 29:
```suggestion
   * Set to true to indicate that the view is using server side parent layout
```

taefi (changes requested):

packages/ts/file-router/src/runtime/RouterConfigurationBuilder.ts
```diff
@@ -143,6 +143,51 @@ export class RouterConfigurationBuilder {
     return this;
   }
 
+  /**
+   * Adds the layoutComponent as the parent layout to views with the flowLayouts ViewConfiguration set.
+   *
+   * @param layoutComponent - layout component to use, usually Flow
+   */
+  withLayouts(layoutComponent: ComponentType): this {
```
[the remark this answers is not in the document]
caalador: done

taefi (approved):
LGTM!

PR #2663, opened 2024-08-19
feat: add automatic subscribe/unsubscribe to signals

### LINKED ISSUE #2621: [Full-stack Signals] add automatic unsubscription from the signal instance
### Describe your motivation

When a user is moving away from a view accessing/manipulating a Full-stack signal, the subscription for that instance should be terminated automatically to prevent memory leakage.

_No response_

_No response_

### Additional context

_No response_

taefi:
## Description

Fixes #2621

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

### DIFF
--- a/packages/ts/react-signals/src/FullStackSignal.ts
+++ b/packages/ts/react-signals/src/FullStackSignal.ts
@@ -1,7 +1,6 @@
-import { computed, signal } from '@preact/signals-react';
 import type { ConnectClient, Subscription } from '@vaadin/hilla-frontend';
 import { nanoid } from 'nanoid';
-import { Signal } from './core.js';
+import { computed, signal, Signal } from './core.js';
 
 const ENDPOINT = 'SignalsHandler';
 
--- a/packages/ts/react-signals/src/FullStackSignal.ts
+++ b/packages/ts/react-signals/src/FullStackSignal.ts
@@ -22,6 +22,41 @@ export type StateEvent<T> = Readonly<{
   value: T;
 }>;
 
+abstract class DependencyTrackingSignal<T> extends Signal<T> {
+  readonly #onSubscribe: () => void;
+  readonly #onUnsubscribe: () => void;
+
+  // -1 means to ignore the first subscription that is created internally in the
+  // FullStackSignal constructor.
+  #subscribeCount = -1;
+
+  protected constructor(value: T | undefined, onSubscribe: () => void, onUnsubscribe: () => void) {
+    super(value);
+    this.#onSubscribe = onSubscribe;
+    this.#onUnsubscribe = onUnsubscribe;
+  }
+
+  protected S(node: unknown): void {
+    // @ts-expect-error: We use the protected method from the base class.
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    super.S(node);
+    if (this.#subscribeCount === 0) {
+      this.#onSubscribe();
+    }
+    this.#subscribeCount += 1;
+  }
+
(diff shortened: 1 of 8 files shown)


packages/ts/react-signals/test/FullStackSignal.spec.tsx
```diff
@@ -4,13 +4,50 @@ import { render } from '@testing-library/react';
...
+      }
+    }
+
+    let onSubscribe: () => void;
+    let onUnsubscribe: () => void;
+    let signal: TestSignal;
+
+    beforeEach(() => {
+      onSubscribe = sinon.spy();
+      onUnsubscribe = sinon.spy();
```
Lodin, line 28: Just a small improvement; I believe, the conception of `stub` fits here better.
```suggestion
      onSubscribe = sinon.stub();
      onUnsubscribe = sinon.stub();
```

packages/ts/react-signals/test/FullStackSignal.spec.tsx
```diff
@@ -4,13 +4,50 @@ import { render } from '@testing-library/react';
...
+  describe('DependencyTrackingSignal', () => {
+    class TestSignal<T = unknown> extends DependencyTrackingSignal<T> {
+      constructor(value: T | undefined, onSubscribe: () => void, onUnsubscribe: () => void) {
+        super(value, onSubscribe, onUnsubscribe);
+        this.subscribe(() => {}); // Ignores the internal subscription.
+      }
+    }
+
+    let onSubscribe: () => void;
+    let onUnsubscribe: () => void;
```
Lodin, line 23: Better to have a specific Sinon type here:
```suggestion
    let onSubscribe: sinon.SinonStub;
    let onUnsubscribe: sinon.SinonStub;
```

packages/ts/react-signals/src/FullStackSignal.ts
```diff
@@ -153,6 +189,26 @@ export abstract class FullStackSignal<T> extends Signal<T> {
       }
     });
 
-    paused = false;
+    this.#paused = false;
+  }
+
+  #connect() {
+    if (this.server.subscription !== undefined) {
```
Lodin, line 196: Why do we need this condition here?

Lodin (changes requested):

packages/ts/react-signals/src/FullStackSignal.ts
taefi: Yeah, good point. This probably cannot/should not happen, atm.
Later, when these connect/disconnect methods are going to be called every time the flux connection recovers (apart from the a first dependency subscribing to the signal), and in case of having an active subscription it should not create another one.

taefi: I'll remove it for now.

packages/ts/react-signals/test/FullStackSignal.spec.tsx
taefi: Thanks for the comment, just out of curiosity can I ask why?
To my understanding a spy is for monitoring and tracking function calls, while stub is more versatile, and I mostly use it e.g. when I want to replace/mock the implementation of a function to return something and etc.
Here, I'm only tracking the function calls and nothing more.

Lodin: The idea here is that `spy` wraps an existing function while `stub` creates a new one.

taefi: Changed to stub.

PR #2666, opened 2024-08-20
Use the correct service name

Legioth:


taefi:
Funny that since we moved to servlet 5, no one complained about this, so either:

* No one is deploying a war package on a servlet container, or
* The `com.vaadin.hilla.startup.EndpointsValidator` is loaded by Flow automatically since it is implementing `com.vaadin.flow.server.startup.ClassLoaderAwareServletContainerInitializer`. 

Then, the question is do we actually need it?
Anyway, this change doesn't harm anyone and should have been applied sooner.

taefi (approved):

Legioth:
Should maybe also be cherry picked to 24.4.x?

taefi:
> Should maybe also be cherry picked to 24.4.x?

At least to 24.4.
2.5 is also using the jakarta namespace.

cromoteca:
This PR fixes the `EndpointValidator` registration in the servlet container, making it work for the first time. The purpose of this validator is to verify that Hilla endpoints are used in a Spring Boot application. The CSRF tests thus fail because they are not Spring Boot applications, but they do have an endpoint, which is `SignalsHandler`.

There are some possible solutions:
- exclude `SignalsHandler` from the check (this is related to https://github.com/vaadin/hilla/pull/2938 and the need of correctly identify the internal endpoints, as in https://github.com/vaadin/hilla/blob/0638cb8df9b8425de9c73a1388e7138dba35d1e0/packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointController.java#L134-L137).
- log a warning instead of throwing in `EndpointValidator`
- make `EndpointValidator` optional by configuration

cromoteca (changes requested):
This PR fixes the `EndpointValidator` registration in the servlet container, making it work for the first time. The purpose of this validator is to verify that Hilla endpoints are used in a Spring Boot application. The CSRF tests thus fail because they are not Spring Boot applications, but they do have an endpoint, which is `SignalsHandler`.

There are some possible solutions:
- exclude `SignalsHandler` from the check (this is related to https://github.com/vaadin/hilla/pull/2938 and the need of correctly identify the internal endpoints, as in https://github.com/vaadin/hilla/blob/0638cb8df9b8425de9c73a1388e7138dba35d1e0/packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointController.java#L134-L137).
- log a warning instead of throwing in `EndpointValidator`
- make `EndpointValidator` optional by configuration

cromoteca:
Some of those changes could not propagate well to some of the cherry-pick target that have been selected for this PR.

cromoteca:
Also, given that it uses a `ClassFinder` from Flow, it seems to collect *all* classes annotated with `BrowserCallable/Endpoint`, but not all classes will be instantiated and used, so it seems a bit too strong as a check.

taefi:
I'm in favor of logging a warning instead of throwing in EndpointValidator. A visible enough warning + documentation should be enough.

cromoteca:
I still see this as a sort of duplicate of https://github.com/vaadin/hilla/pull/2938. While they serve different purposes, the check for the presence of Spring could be added at the same place instead of having to declare a dedicated listener.

cromoteca:
No longer needed after https://github.com/vaadin/hilla/pull/3167

PR #2672, opened 2024-08-22
feat: add generic single valued signal (ValueSignal)

### LINKED ISSUE #2623: [Full-stack Signals] add support for having single valued signals other than numbers
### Describe your motivation

Full-stack signals library should offer signals that can hold a single shared value of types other than numbers, e.g. JSON-based types `string` and `boolean`. 

A generic type called `ValueSignal<T>` can be used for representing these types as of now. In the future, if there was a need for introducing atomic operations for these types as well, we can introduce more specific types.

The fact that `NumberSignal` is the designated signal type to handle atomic arithmetic operations should not prevent users from having `ValueSignal<Integer>` instances. For instance:

```java
@AnonymousAllowed
@BrowserCallable
public class ShoppingCartService {
    private final ValueSignal<Integer> cartSize = new ValueSignal<>(0);

    public ValueSignal<Integer> size() {
        return cartSize;
    }
}
```

```ts
const cartSize = ShoppingCartService.size();
```
(Disclaimer: The above is just an example to show the possibility of using `ValueSignal` with numbers as well, not a recommendation for implementation.) 

The suggested `ValueSignal<T>` supports three different write operations that all replace the current value with a new value with different approaches for dealing with concurrent changes (using the above example for brevity):
1. Last-write-wins (overwrite):
```ts
cartSize.value++; // this is the same as followings:
// cartSize.value = cartSize.value + 1;
// cartSize.set(cartSize.value + 1);
```

2. Conditional replace (also known as "compare and set") - the change is applied only if the expected value matches the current value when the change is applied on the server:
```ts
const expected = cartSize.value;
cartSize.replace(expected, expected + 1);
```

3. Runs the callback and tries to set the returned value as the signal value. In case of a concurrent change, the callback is run again with an updated input value. This is repeated until the result can be applied without concurrent changes:
```ts
const updateOperation = cartSize.update(lastSeenValue => lastSeenValue + 1);
// Abort if contention has prevented success after 1000 ms
setTimeout(() => updateOperation.cancel(), 1000);
```
Note that there's no guarantee that cancel() will be effective since a succeeding operation might already be on its way to the server. 

_No response_

### Additional context

_No response_

taefi:
## Description

Fixes #2623

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [ ] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.

### DIFF
--- a/packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/NumberSignal.java
+++ b/packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/NumberSignal.java
@@ -1,44 +1,20 @@
 package com.vaadin.hilla.signals;
 
-import com.fasterxml.jackson.databind.node.ObjectNode;
-
-import com.vaadin.hilla.signals.core.StateEvent;
-import jakarta.annotation.Nullable;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import reactor.core.publisher.Flux;
-import reactor.core.publisher.Sinks;
-
-import java.util.HashSet;
-import java.util.Objects;
-import java.util.Set;
-import java.util.UUID;
-import java.util.concurrent.locks.ReentrantLock;
-
 /**
  * A signal that holds a number value.
  */
-public class NumberSignal {
-
-    private static final Logger LOGGER = LoggerFactory
-            .getLogger(NumberSignal.class);
-
-    private final ReentrantLock lock = new ReentrantLock();
-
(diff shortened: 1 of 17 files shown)


platosha (commented):
LGTM overall. Please mind the TypeScript error in the tests, makes the build red.

taefi:
> LGTM overall. Please mind the TypeScript error in the tests, makes the build red.

Done.

platosha (approved):

PR #2704, opened 2024-09-05
fix: adjust increment api and remove add from NumberSignal

taefi:
## Description

This will introduce increment event for NumberSignal so that it is processed atomically on server based on the last seen value of NumberSignal without any retries.

Also, this refactors the increment method to incrementBy, and also removes the add methods that were introduced in #2694


taefi:
I wonder what's the problem with `codecove/project` check failure. Maybe, it is confused with the file `Signals.ts` being renamed to `NumberSignal.ts`?

platosha (approved):

packages/java/endpoint/src/test/java/com/vaadin/hilla/signals/core/event/StateEventTest.java
```diff
@@ -134,7 +134,7 @@ public void constructor_withJsonInvalidEventType_shouldThrowInvalidEventTypeExce
                 StateEvent.InvalidEventTypeException.class,
                 () -> new StateEvent<>(json, String.class));
 
-        String expectedMessage = "Invalid event type invalidType. Type should be either of: [SNAPSHOT, SET, REPLACE, REJECT]";
+        String expectedMessage = "Invalid event type invalidType. Type should be either of: [SNAPSHOT, SET, REPLACE, REJECT, INCREMENT]";
```
cromoteca, line 137: "either of" is rather used with two options. Should be "one of".

packages/java/endpoint/src/test/java/com/vaadin/hilla/signals/NumberSignalTest.java
```diff
@@ -96,6 +96,23 @@ public void submit_eventWithUnknownCommand_throws() {
...
+        signal.submit(createIncrementEvent("2"));
+        assertEquals(44.0, signal.getValue(), 0.0);
+
+        signal.submit(createIncrementEvent("-5.5"));
+        assertEquals(38.5, signal.getValue(), 0.0);
+    }
+
+    private ObjectNode createIncrementEvent(String value) {
+        var setEvent = new StateEvent<>(UUID.randomUUID().toString(),
+                StateEvent.EventType.INCREMENT, Double.parseDouble(value));
```
cromoteca, line 112: `valueOf` is slightly better than `parseDouble` here (IDE suggestion)

cromoteca (approved):
LGTM! Just two minor things to report, feel free to ignore.

PR #2706, opened 2024-09-06
fix: Nested child with flowLayout and autolayout withFallback

caalador:
Having a flowLayout marked view
in folder should add flow layout.

Setting component for withFallback
will automatically register withLayout
for that component as layoutComponent.


packages/ts/file-router/src/runtime/RouterConfigurationBuilder.ts
```diff
@@ -150,16 +152,7 @@ export class RouterConfigurationBuilder {
...
-        if (route.children === undefined) {
-          return route;
-        }
-
-        return {
-          ...route,
-          children: applyLayouts(route.children),
-        } as RouteObject;
-      });
+      const nestedRoutes = routes.map((route) => route);
```
platosha, line 155: The map call seems to do nothing. Could we remove it then?

platosha (approved):
LGTM but one nitpick, which be addressed in a follow-up PR.

packages/ts/file-router/src/runtime/RouterConfigurationBuilder.ts
caalador: I think it was left just because routes was readonly and it "broke" something.

caalador: Yes so I get a `TS4104: The type readonly RouteObject[] is readonly and cannot be assigned to the mutable type RouteObject[]` when trying to use routes directly.

Issue #2643, opened 2024-08-06
In Dev-Mode, AutoCrud throws error when selecting an entity of which one field has the value null


sahrmann:
### Describe the bug

If I use an `<AutoCrud>`-Component and I select an entity of which one field is `null`, the following error is shown:

```
Unexpected Application Error!
baseQueue is undefined

updateReducer@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/chunk-E4UNORPT.js?v=65bb8eeb:11752:25
useReducer@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/chunk-E4UNORPT.js?v=65bb8eeb:12737:24
useReducer@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/chunk-IF5A3MEL.js?v=65bb8eeb:1070:29
useUpdate@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/@vaadin_hilla-react-crud.js?v=65bb8eeb:1211:52
useForm@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/@vaadin_hilla-react-crud.js?v=65bb8eeb:1328:18
AutoForm@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/@vaadin_hilla-react-crud.js?v=65bb8eeb:1492:23
renderWithHooks@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/chunk-E4UNORPT.js?v=65bb8eeb:11548:35
updateFunctionComponent@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/chunk-E4UNORPT.js?v=65bb8eeb:14582:28
beginWork@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/chunk-E4UNORPT.js?v=65bb8eeb:15924:22
beginWork$1@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/chunk-E4UNORPT.js?v=65bb8eeb:19753:22
performUnitOfWork@http://localhost:8080/VAADIN/@fs/../my-app/node_modules/.vite/deps/chunk-E4UNORPT.js?v=65bb8eeb:19198:20
[...]

PR #2707, opened 2024-09-06
fix: send updates only when not rendering

krissvaa:
React doesn't allow updates to state while rendering.
[Reference](https://github.com/facebook/react/issues/18178)

Fixes: #2643


packages/ts/react-form/test/useForm.spec.tsx
```diff
@@ -6,7 +6,7 @@ import chaiDom from 'chai-dom';
 import { useEffect, useState } from 'react';
 import sinon from 'sinon';
 import sinonChai from 'sinon-chai';
-import { useForm as _useForm, useFormArrayPart, useFormPart } from '../src/index.js';
+import { useForm as _useForm, useFormArrayPart, useFormPart } from '../src';
```
taefi, line 9: I think we better not to use `../src` in test imports, but the exact file, in this case the `/index.js`:
```suggestion
import { useForm as _useForm, useFormArrayPart, useFormPart } from '../src/index.js';
```

taefi (approved):
Overal, LGTM!
Just added one minor comment.

taefi (approved):

platosha (approved):

PR #2708, opened 2024-09-06
feat: support bidirectional communication on push channels

cromoteca:
This is a first working version of `EndpointSubscription` offering bidirectional communication. This is just the minimal code needed to have something working, many improvements are needed to have a robust feature.

Note that the flux must be open to allow to handle messages. At the moment, if no more messages, the flux is closed and received messages are ignored.

Example of an endpoint method:
```java
public EndpointSubscription<String> subscribeToHello(String name) {
    return EndpointSubscription.of(Flux.interval(Duration.ofSeconds(1)).map(l -> name), (String msg) -> {
        System.out.println(msg);
    });
}
```

Usage example:
```typescript
const subscription = useSignal(HelloEndpoint.subscribeToHello('some text'));
subscription.value.send('other text');
// ...
```

Some obvious improvements:
- subclass `EndpointSubscription` to define different types for the two message directions
- handle security, errors and lost messages


Legioth:
Is it a valid assumption that outgoing and incoming messages are of the same type? Also, the API might need explicitly passing a class literal for the received type so that Jackson can know what type of Java instance to create if receiving anything else than a string in the `content` property.

> `useSignal(HelloEndpoint.subscribeToHello('some text'));`

This might be slightly problematic since it means that `subscribeToHello` will be called every time the render function is called, even though the returned value will be ignored in all cases except the first one.


> `subscription.value.send('other text');``

Isn't a regular Flux endpoint subscription passive until `onNext` has been called? Should `send` also open the subscription if it's not yet open, or should the function be moved so that you need to run `onNext` to get access to the send function?

Issue #2670, opened 2024-08-22
useFormPart of nested optional form fields causes an error


peholmst:
### Describe the bug

If a form model contains nested object models that are optional, `useFormPart` throws an error when any of the nested properties are accessed. Given these data structures:

```typescript
interface StreetAddress {
    number?: string
    name?: string
}

interface PostalAddress {
    streetAddress?: StreetAddress
}
```

The following code will throw a `TypeError: Cannot read properties of undefined (reading 'number')`:

```typescript
const streetNumberField = useFormPart(form.model.streetAddress.number)
```

### Expected-behavior

If the data types (and corresponding models) are changed as follows (`streetAddress` is no longer optional), the code works as expected (no errors):
```typescript
```typescript
interface StreetAddress {
    number?: string
    name?: string
}

interface PostalAddress {
    streetAddress: StreetAddress
}
```

The behavior should be the same regardless of whether `streetAddress` is optional or not.

### Reproduction

Consider this code:

```tsx
import {
    _getPropertyModel,
    makeObjectEmptyValueCreator,
    NotEmpty,
    ObjectModel,
    StringModel
} from "@vaadin/hilla-lit-form";
import {useForm, useFormPart} from "@vaadin/hilla-react-form";
import {useEffect} from "react";
import {TextField} from "@vaadin/react-components";

interface StreetAddress {
    number?: string
    name?: string
}

interface PostalAddress {
    streetAddress?: StreetAddress
}

class StreetAddressModel<T extends StreetAddress = StreetAddress> extends ObjectModel<T> {
    static override createEmptyValue = makeObjectEmptyValueCreator(StreetAddressModel)

    get number(): StringModel {
        return this[_getPropertyModel]("number", (parent, key) => new StringModel(parent, key, true, {meta: {javaType: "java.lang.String"}}))
    }

    get name(): StringModel {
        return this[_getPropertyModel]("name", (parent, key) => new StringModel(parent, key, true, {meta: {javaType: "java.lang.String"}}))
    }
}

class PostalAddressModel<T extends PostalAddress = PostalAddress> extends ObjectModel<T> {
    static override createEmptyValue = makeObjectEmptyValueCreator(PostalAddressModel)

    get streetAddress(): StreetAddressModel {
        return this[_getPropertyModel]("streetAddress", (parent, key) => new StreetAddressModel(parent, key, true))
    }
}

export default function DebugView() {
    const form = useForm(PostalAddressModel)

[...]

Issue #2658, opened 2024-08-16
Form Binding (FirstNameComponent example) renders "undefined", crashes with nullable fields


Pascalmh:
### Describe the bug

I’m trying to create my own Form Components (Text-Input, Select-Boxes, etc.)

I copied the “FullNameComponent” from the docs - when using it as it is it renders the String “undefined” as the content of the <input>-Element.

When I'm adding another property to the PersonModel that's nullable:

```
@Nullable
private Location location;
``` 

that takes a `private String streetName` and I'm passing that to the FullNameComponent via the fullNameModel-prop I'm getting this error:

> Error: Should have a queue. This is likely a bug in React. Please file an issue.

I've pushed a reproduction repo here (Branch: custom-text-input-component) with just the "initial commit" with the code from the "skeleton-starter-hilla-react-24.zip" and [a second commit that reproduces the issues](https://github.com/Pascalmh/vaadin-storybook/commit/550a1289830ebfff2f83d76c2497996df7707e89) I'm facing.

https://github.com/Pascalmh/vaadin-storybook/tree/custom-text-input-component

### Expected-behavior

Using the FullNameComponent copied from the docs works the same as using the Hilla TextField component.

### Reproduction

Repo: https://github.com/Pascalmh/vaadin-storybook/tree/custom-text-input-component
Commit: https://github.com/Pascalmh/vaadin-storybook/commit/550a1289830ebfff2f83d76c2497996df7707e89

### System Info

macOS 14.5, Vaadin 24.4.8, ARC 1.56.0 (Chromium Engine Version 127.0.6533.120)

taefi:
It seems to work as intended, explained here: https://vaadin.com/forum/t/form-binding-firstnamecomponent-example-renders-undefined-as-input-value/166913/4

sahrmann:
Could be related to https://github.com/vaadin/hilla/issues/2643

platosha:
One idea to fix this bug is to initialise all the primitive value keys to their respective defaults.

taefi:
#2713 fixed the crash issue, but still the problem of rendering the literal string `'undefined'` remains open.

PR #2724, opened 2024-09-11
fix: binder values initialized only when needed

krissvaa:
Fixed that binder initialized value and default value preemptively. 

Fixes: #2670


packages/ts/react-form/test/useForm.spec.tsx
```diff
@@ -370,6 +371,19 @@ describe('@vaadin/hilla-react-form', () => {
...
+      function StreetAddressView() {
+        const { model, field } = useForm(PostalAddressModel);
+        useFormPart(model.streetAddress.name);
+
+        return <input data-testid="street-name" {...field(model.streetAddress.name)} />;
+      }
+
+      const { container } = render(<StreetAddressView />);
+      const streetNameField = await waitFor(() => container.querySelector('input')!);
+      expect(streetNameField.value).to.have.rendered.eq('undefined');
```
taefi, line 385: I think this line could be simply changed to:
```suggestion
      expect(streetNameField.value).to.be.equal('undefined');
```
since this is actually the current behaviour when binding to a plain `<input />` element.
Related issue: #2658

taefi: For making sure it remains compliant with the React Components, we can also render another component that renders a `<TextField />` and check that the `'undefined'` is taken care of in our library, by expecting the value of the `<TextField />` element is equal to `''`, not `'undefined'`.

taefi (changes requested):

packages/ts/react-form/test/useForm.spec.tsx
krissvaa: `TextField` test would add "react-components" dependency that would need have its own version as it is done in react-crud. 
I think the best action would be to fix https://github.com/vaadin/hilla/issues/2658

taefi: OK, let's keep it as it is with `<input />`. I believe the `to.have.rendered.eq('undefined');` could still be simplified to `to.be.equal('undefined');`.

Issue #2709, opened 2024-09-09
Make Flow layout usage opt-out


caalador:
### Describe your motivation

When having available server side Layout and no react layout is defined we should automatically get the server side layout as long as the developer doesn't mark the view as no flow layout.

This will make the Layout feature work the same way as it does with Flow routes.

### Describe the solution you'd like

When file-routes are generated we check routes without layout against available server layouts and add the `"flowLayout": true` flag into the file-routes.json so that the server side layout is used.

caalador:
Will close vaadin/flow#19904

PR #2725, opened 2024-09-11
feat: automtic flowLayout if match exists

caalador:
Automatically enable flowLayout
for views if a server layout
exists that matches, the flag
is not explicitly defined.

Closes https://github.com/vaadin/hilla/issues/2709


platosha (approved):
I did some minor cleanup in TypeScript code. Looks good overall.

platosha (approved):

PR #2732, opened 2024-09-16
fix(signals): require a default value for non-nullable signals

cromoteca:
This PR amends the code generation to require a default value when the Java signal is not nullable.

Fixes # 2719

### DIFF
--- a/packages/ts/generator-plugin-signals/src/SignalProcessor.ts
+++ b/packages/ts/generator-plugin-signals/src/SignalProcessor.ts
@@ -1,5 +1,6 @@
 import type Plugin from '@vaadin/hilla-generator-core/Plugin.js';
 import { template, transform, traverse } from '@vaadin/hilla-generator-utils/ast.js';
+import createFullyUniqueIdentifier from '@vaadin/hilla-generator-utils/createFullyUniqueIdentifier.js';
 import createSourceFile from '@vaadin/hilla-generator-utils/createSourceFile.js';
 import DependencyManager from '@vaadin/hilla-generator-utils/dependencies/DependencyManager.js';
 import PathManager from '@vaadin/hilla-generator-utils/dependencies/PathManager.js';
@@ -11,6 +12,7 @@ const CONNECT_CLIENT = '$CONNECT_CLIENT$';
 const METHOD_NAME = '$METHOD_NAME$';
 const SIGNAL = '$SIGNAL$';
 const RETURN_TYPE = '$RETURN_TYPE$';
+const INITIAL_VALUE = '$INITIAL_VALUE$';
 
 const signals = ['NumberSignal', 'ValueSignal'];
 const genericSignals = ['ValueSignal'];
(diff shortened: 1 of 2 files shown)


packages/ts/generator-plugin-signals/src/SignalProcessor.ts
```diff
@@ -44,30 +46,48 @@ export default class SignalProcessor {
...
+          // Searching for a non-nullable signal like ValueSignal<SomeType>
+          if (
+            ts.isUnionTypeNode(returnType) &&
+            returnType.types.length &&
+            ts.isTypeReferenceNode(returnType.types[0]) &&
+            returnType.types[0].typeArguments?.length === 1 &&
+            ts.isUnionTypeNode(returnType.types[0].typeArguments[0]) &&
+            returnType.types[0].typeArguments[0].types.length === 1
+          ) {
+            const requiredDefaultType = returnType.types[0].typeArguments[0].types[0];
```
taefi, line 72: I'd suggest to wrap this logic in a function called something like `getRequiredDefaultValueType` for the sake of better readability.

packages/ts/generator-plugin-signals/test/fixtures/ValueSignalServiceMix.snap.ts
```diff
@@ -6,7 +6,9 @@ async function getPerson_1(init?: EndpointRequestInit_1): Promise<Person_1 | und
 function personSignal_1(): ValueSignal_1<Person_1 | undefined> {
     return new ValueSignal_1(undefined, { client: client_1, endpoint: "PersonService", method: "personSignal" });
 }
-function personSignalNonNull_1(): ValueSignal_1<Person_1> {
-    return new ValueSignal_1(undefined, { client: client_1, endpoint: "PersonService", method: "personSignalNonNull" });
+function personSignalNonNull_1({ defaultValue: defaultValue_1 }: {
```
taefi, line 9: Could we also have a test function that validates the combination of a default client-side value and the service params work together?

taefi (changes requested):

packages/ts/generator-plugin-signals/test/fixtures/ValueSignalServiceMix.snap.ts
cromoteca: Good catch! It needs a fix indeed.

packages/ts/generator-plugin-signals/test/fixtures/ValueSignalServiceMix.snap.ts
```diff
@@ -3,10 +3,24 @@ import { ValueSignal as ValueSignal_1 } from "@vaadin/hilla-react-signals";
...
+function personSignalNonNull_1({ defaultValue: defaultValue_3 }: {
+    defaultValue: Person_1;
+}): ValueSignal_1<Person_1> {
+    return new ValueSignal_1(defaultValue_3, { client: client_1, endpoint: "PersonService", method: "personSignalNonNull" });
+}
+function personSignalNonNullWithParams_1(dummyBoolean: boolean, dummyString: string | undefined, { defaultValue: defaultValue_4 }: {
+    defaultValue: Person_1;
+}): ValueSignal_1<Person_1> {
+    return new ValueSignal_1(defaultValue_4, { client: client_1, endpoint: "PersonService", method: "personSignalNonNullWithParams", params: { dummyBoolean, dummyString } });
+}
```
taefi, line 25: Just thinking out loud (feel free to ignore, as it is not breaking anything):
Is the alias for `defaultValue` needed to be unique in the whole module? Isn't a function scoped uniqueness enough?

taefi (approved):
LGTM!

PR #2733, opened 2024-09-16
fix: Do not wrap with layout for empty array

caalador:
Empty array should just return
and not wrap the empty array.

Fixes vaadin/flow#19956


mshabarov (dismissed):
This patch works for me 👍 
Just tested with `react-crm-tutorial` example project by patching `RouterConfigurationBuilder.ts` in `hilla-file-router` accordingly.

Needs a test in Hilla though.

ZheSun88:
the failure here seems not relate to this PR, as main branch snapshot has the same error

platosha (approved):

PR #2734, opened 2024-09-17
test: avoid top level await

caalador:
Do not use top level await.


platosha:
Well, it's OK, but I'd rather prefer changing the test module away from using top-level await for now, something like:

```tsx
i18n.configure().then(render);
```

caalador:
You can update this PR with that as I don't know how that would be set up.

caalador (commented):
:+1:

platosha (approved):

PR #2737, opened 2024-09-17
feat(push): add callback on subscription lost event

### LINKED ISSUE #2716: [push] client tries to reconnect to a Flux that no longer exists
### Describe the bug

https://github.com/vaadin/hilla/pull/2684 introduced support for reconnection, but if the underlying Flux is no longer active, that reconnection keeps trying indefinitely.

That can happen, for example, after server timeout (`ApplicationConfig.WEBSOCKET_IDLETIME` in the Atmosphere servlet configuration).

### Expected-behavior

When the Flux is no longer there, the retry should stop at the first attempt and the API should give the developer an option to reconnect. There's no universal logic for that: only the developer knows what to do in that case.

### Reproduction

Configuring a short idle time (it's in milliseconds) allows to test that case easily.

### System Info

Hilla 24.5.alpha13

cromoteca:
This PR allows to define a behavior for when the `reopened` event happens on the client.

The new `onSubscriptionLost` callback can be used to perform an action when the websocket is reopened, but there's no longer a Flux connected to the subscription. Based on the return value of that callback, automatic resubscription can be performed.

Fixes #2716


taefi (approved):
LGTM and can be merged when the tests are passed.

packages/ts/frontend/src/FluxConnection.ts
```diff
@@ -187,7 +231,11 @@ export class FluxConnection extends EventTarget {
         }
         this.#removeSubscription(id);
         if (!callback) {
-          throw new Error(`Error in ${endpointInfo}: ${message.message}`);
+          throw new Error(
+            endpointInfo
+              ? `Error in ${endpointInfo.endpointName}.${endpointInfo.methodName}(${JSON.stringify(endpointInfo.params)}): ${message.message}`
+              : `Error in unknown subscription: ${message.message}`,
+          );
```
taefi, line 238: Minor: A test can be added to cover this change.

platosha (approved):

PR #2745, opened 2024-09-19
fix: fix layout matching

caalador:
The automatic layout matching
is mistakenly made as equals
when it should be starts with
path parts.

Fixes vaadin/flow#20072


packages/ts/file-router/src/vite-plugin/applyLayouts.ts
```diff
@@ -26,6 +26,36 @@ function enableFlowLayout(route: RouteMeta): RouteMeta {
...
+      const splitLayout: string[] = layout.split('/');
+      if (splitLayout.length > splitPath.length) {
+        return false;
+      }
+      for (let i = 0; i < splitLayout.length; i++) {
+        if (!(splitPath[i] === splitLayout[i])) {
+          return false;
+        }
+      }
+      return true;
```
taefi, line 54: Isn't this the same as `path.startWith(layout)`?
Another thing, can layouts have patterns e.g. `/home/*`? If so, I believe `/home/*` should match a path like `/home/contacts` but the current logic will return false, if I understand correctly. The trailing `/` and `possible `/*` should also be stripped before this logic is checked.

caalador: The path parts are split with / so there should be not / in any part. Except if javascript split works in mysterious ways.

No patterns as you only give the path it matches to so all given layout paths are seen as ending with /*

caalador: Also it is not the same as startWith as if the layout path is `/home/owner` it should not match `/home/owner2/house`, but should match `/home/owner/house`

packages/ts/file-router/src/vite-plugin/applyLayouts.ts
```diff
@@ -26,6 +26,36 @@ function enableFlowLayout(route: RouteMeta): RouteMeta {
...
+ * and '/home' matches '/home' anything with the start path '/home/*'
+ *
+ * @param layoutPaths - available layout paths
+ * @param path - to check for layout
+ */
+function layoutExists(layoutPaths: string[], path: string) {
+  const splitPath: string[] = path.split('/');
+
+  return (
+    layoutPaths.filter((layout) => {
```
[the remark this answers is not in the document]
caalador: done

packages/ts/file-router/src/vite-plugin/applyLayouts.ts
```diff
@@ -26,6 +26,34 @@ function enableFlowLayout(route: RouteMeta): RouteMeta {
...
+  return layoutPaths.some((layout) => {
+    if (layout.length === 0) {
+      return true;
+    }
+    const splitLayout: string[] = layout.split('/');
+    if (splitLayout.length > splitPath.length) {
+      return false;
+    }
+    for (let i = 0; i < splitLayout.length; i++) {
+      if (!(splitPath[i] === splitLayout[i])) {
```
taefi, line 49: nit:
```suggestion
      if (splitPath[i] !== splitLayout[i]) {
```

taefi (approved):

taefi (approved):
LGTM

Issue #2769, opened 2024-09-26
Introduce onConnectionStateChange callback for Flux subscriptions-


krissvaa:
### Describe your motivation

Add `onConnectionStateChange` callback and also change the previous `onDisconnect' callback to receive the current connection state. A subscription can have three different connection states.

- `connecting `in two cases: a) when initially setting up the subscription and waiting for the WebSocket to be established and while waiting for confirmation that the subscription has been set up and b) after what currently triggers onDisconnect and until either getting back to receiving live updates or giving up.
- `connected` while receiving live updates
- `closed` when explicitly closed or when implicitly closed after lost server-side state and automatic resubscribe is not enabled.
Mapping to UI states, connecting corresponds to a loading indicator, connected corresponds to a green light, and closed corresponds to a "retry" button.

Different states should be added as an enum, and the state should be passed as an event object that wraps the state enum value.

PR #2782, opened 2024-10-01
feat: add onConnectionStateChange callback for flux subscription

krissvaa:
Fixes: #2769


platosha (changes requested):
There're some errors in the CI that seem relevant with the changes in the PR. Would it be possible to address them?

platosha (approved):

Issue #2587, opened 2024-06-28
protected setter/getter is in the generated typescript


jcgueriaud1:
### Describe the bug

When I have this Java class:
```

public class Person {

    @NotNull
    private String fullName = "fullName";

    @NotNull
    private String descriptionPublic = "descriptionPublic";
    @NotNull
    private String descriptionProtected = "descriptionProtected";

    public String getFullName() {
        return fullName;
    }

    public void setFullName(String fullName) {
        this.fullName = fullName;
    }

    protected String getDescriptionProtected() {
        return descriptionPublic;
    }

    protected void setDescriptionProtected(String descriptionProtected) {
        this.descriptionProtected = descriptionProtected;
    }

    public String getDescriptionPublic() {
        return descriptionPublic;
    }
}

```

The model generated in Typescript is containing `descriptionProtected`:
```
interface Person {
    fullName?: string;
    descriptionPublic?: string;
    descriptionProtected?: string;
}
export default Person;

```

The data sent doesn't contain the `descriptionProtected`

### Expected-behavior

The model should be: 
```
interface Person {
    fullName?: string;
    descriptionPublic?: string;
}
export default Person;

```

Note: If you remove the Setter then the model doesn't contain `descriptionProtected`.

### Reproduction

https://github.com/jcgueriaud1/hilla-example-form/tree/bug-protected

Go to the home page and see the data in the copilot log (the data is ok).

Go to the `Person.ts` and see the `descriptionProtected`

### System Info

Vaadin 24.4.4

PR #2783, opened 2024-10-02
fix: generator use only public getters and setters

krissvaa:
Fixes: #2587


platosha (changes requested):
Looks good so far, but let's also add an IT test for this use case to make sure that the runtime JSON is in sync with the generated TypeScript.

cromoteca:
> Looks good so far, but let's also add an IT test for this use case to make sure that the runtime JSON is in sync with the generated TypeScript.

Isn't a unit test enough?

platosha (approved):
Looking good, thanks!

Issue #2390, opened 2024-05-06
Add support for `OffsetDateTime` and `OffsetTime`


cromoteca:
### Describe your motivation

The generator supports the `java.time.Local*` date types, but not the `Offset*` counterparts. This is unfortunate, as they are the preferred classes for data transfer. From the [java.time JavaDoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/package-summary.html):

> The offset-based date-time types `OffsetTime` and `OffsetDateTime`, are intended primarily for use with network protocols and database access.

### Describe the solution you'd like

Those types are mapped to `string` instead of `unknown`.

### Describe alternatives you've considered

There are easy workarounds like casting on the client or using a DTO that converts to `String`, but they are quite ugly.

### Additional context

There's a [related forum thread](https://vaadin.com/forum/t/offsetdatetime-support-for-the-endpointgenerator/166175) from @deniz-frick that illustrates the issue.

platosha:
Note that we need to update the CRUD components to support `OffsetTime`/`OffsetDateTime`.

PR #2785, opened 2024-10-02
feat: add support for java.time offset and zoned date-time

krissvaa:
Fixes: #2390


cromoteca:
Values seem to be converted correctly, so once the check errors are adressed, it should be easy to approve this one.

cromoteca (approved):

PR #2795, opened 2024-10-04
feat: HMR for translations

### LINKED ISSUE #2175: [i18n] Add support for HMR when changing translations
When the developer changes translations in a properties file, or Copilot writes a new translation string into a property file, the frontend should update immediately and render the updated translation without a page reload.

The current approach is to serve translation strings from the server at runtime, where the server looks up translations from a resource bundle packaged into the Java application. That means just changing a properties file would have no effect on the frontend. The developer would also have to recompile the Java app which updates the resource bundle and triggers a restart of the Spring app, which would then also trigger a reload of the page in the browser. Ideally our solution would notice changes to property files and then update translations in the frontend automatically, without requiring a page reload.

Two options come to mind:
1. In development mode, the server request handler does not use the compiled resource bundle, but loads the contents of the property files from the file system. We would then add a watcher for the property files on the file system, somehow use the Vite dev server to push a notice to the browser that those files have changed, which in turn causes the I18n instance to reload translations from the server.
2. We don't load translations from the server at all, but instead compile the property files into Javascript modules as part of the Vite build. The I18n instance would use dynamic imports to load these JS modules. By using JS imports we can reuse the Vite HMR mechanism that automatically updates these modules when they are recompiled.

Artur-:
Fixes #2175


packages/ts/react-i18n/src/index.ts
```diff
@@ -182,6 +191,32 @@ export class I18n {
...
+   */
+  private async reloadTranslations() {
+    const currentLanguage = this.#language.value;
+    if (!currentLanguage) {
+      return;
+    }
+
+    let translationsResult: TranslationsResult;
+    try {
+      translationsResult = await this.#backend.loadTranslations(currentLanguage);
```
Artur-, line 206: Should the cache for other languages be cleared here also? If I change german translations and are looking at English ones, I want to see the updated german strings when switching the language back to german

sissbruecker: There is no cache, when you change the language it always loads all translations that have been loaded so far again.

platosha (approved):

Issue #2750, opened 2024-09-20
hillaI18n Feature flags for experimental features is ignored in production mode


jcgueriaud1:
### Describe the bug

This issue is still happening in Vaadin 24.4.12 and Vaadin 24.5.0.beta1.

https://github.com/vaadin/hilla/issues/2637

> I have a file vaadin-featureflags.properties which contains com.vaadin.experimental.hillaI18n=true. In my local environment, the i18n-instance can be used without problems. However, in my production build, the frontend crashes with the error The Hilla I18n API is currently considered experimental and may change in the future. To use it you need to explicitly enable it in Copilot or by adding com.vaadin.experimental.hillaI18n=true to vaadin-featureflags.properties.

### Reproduction

https://github.com/jcgueriaud1/spring-petclinic-vaadin-flow/tree/hilla

Build and run the application in production mode

### System Info

Vaadin 24.4.12 and Vaadin 24.5.0.beta1

platosha:
Last time this was caused by using a wrong class loader when generating feature flags TypeScript values in production build.

cromoteca:
Fixed in `24.6` by https://github.com/vaadin/platform/pull/6897, cherry-picked to `24.5` and `24.4`. It will be included in the next patch releases.

jcgueriaud1:
Fixed in 24.5.1

sveinnetnordic:
Works for me :-)

cromoteca:
Released in `24.5.1` and `24.4.16`.

PR #2798, opened 2024-10-07
fix: check full-stack signals feature-flag in client-side

taefi:
Check whether the experimental feature flag is set when using FullStack Signal in a project client-side code.
This prevents users from getting confused with "not a flux..." error in their browser when subscribing to a FS signal.


cromoteca:
Does https://github.com/vaadin/hilla/issues/2750 block this one also?

taefi:
> Does #2750 block this one also?

Well, until #2750 is fixed, this check will not probably work in production, but at least it prevents the confusion for the users in DEV mode.

cromoteca (approved):

PR #2799, opened 2024-10-07
feat(full-stack signals): make operations thenable

### LINKED ISSUE #2712: [Full-stack Signals] Make signal operations thenable
All client-side operations that may modify the value of a signal should return a thenable object that will be resolved once it has been confirmed that the operation has been processed on the server.

Note that this is a custom object with a `then` method rather than a full-blown `Promise` since the object will also have other properties. Most of the other properties are specific to the operation but there's also a universal `cancel()` function that will prevent the operation from being sent to the server if it hasn't already been sent (no guarantee that the cancellation will actually succeed). The thenable type can be `void` for now but will eventually be customized for various operations to reflect what the operation actually did.

cromoteca:
This pull request introduces a feature to make operations "thenable", which means it implements the ability for operations to return promise-like objects that can be chained. Those objects are in the `result` properties to avoid the return type themselves to be treated as promises by lint checkers.

Closes #2712


packages/ts/react-signals/test/ValueSignal.spec.tsx
```diff
@@ -137,6 +137,32 @@ describe('@vaadin/hilla-react-signals', () => {
...
+    it('should accept a callback after replace', (done) => {
+      const valueSignal = new ValueSignal<string>('a', config);
+      subscribeToSignalViaEffect(valueSignal);
+      valueSignal.replace('a', 'b').result.then(done);
+      const [, , params] = client.call.firstCall.args;
+      simulateReceivedChange(subscription, {
+        id: (params!.event as { id: string }).id,
+        type: 'replace',
+        value: 'b',
+        accepted: true,
```
taefi, line 149: Do we need to add couple of tests to cover the case that the `accepted = false` in which the then callback should not be called? (same for the test in `NumberSignal.spec.ts`)

packages/ts/react-signals/test/ValueSignal.spec.tsx
```diff
@@ -137,6 +137,32 @@ describe('@vaadin/hilla-react-signals', () => {
       expect(valueSignal.value).to.equal('baz');
     });
 
+    it('should accept a callback after replace', (done) => {
```
taefi, line 140: Just asking due to my lack of knowledge:
How passing the `done` to `.result.then(done)` asserts that it is called?

packages/ts/react-signals/src/ValueSignal.ts
```diff
@@ -86,20 +129,25 @@ export class ValueSignal<T> extends FullStackSignal<T> {
...
-        record.waiter.resolve();
-      }
       this.#applyAcceptedEvent(event);
+      // `then` callbacks can be associated to the record or the event
+      // it depends on the operation that was performed
+      [record?.id, event.id].filter(Boolean).forEach((id) => this.runThenCallback(id!));
+    }
+  }
+
+  protected runThenCallback(eventId: string): void {
```
taefi, line 146: Could the function signature be modified to resemble other `protected` methods in [FullStackSignal.ts](https://github.com/vaadin/hilla/blob/main/packages/ts/react-signals/src/FullStackSignal.ts#L212)?

packages/ts/react-signals/src/ValueSignal.ts
```diff
@@ -7,17 +8,41 @@ import {
...
+/**
+ * An operation where all callbacks are predefined to be no-ops.
+ */
+export const noOperation: Operation = Object.freeze({
+  result: {
+    then(_callback: ThenCallback): Operation['result'] {
+      return noOperation.result;
+    },
+  },
+});
```
taefi, line 35: Could all this common definitions be moved to the `FullStackSignals.ts` so that other signals (e.g. ListSignal) can also leverage from the same mechanisms for making their operations `thenable`?

taefi (changes requested):

packages/ts/react-signals/test/ValueSignal.spec.tsx
cromoteca: Mocha itself expects `done` to be called if declared in the function. So, if it's not called, the test waits for some timeout and then fails.

packages/ts/react-signals/src/ValueSignal.ts
cromoteca: Moved them, although I think that those kind of changed can be done when needed.

packages/ts/react-signals/src/NumberSignal.ts
```diff
@@ -38,24 +38,28 @@ export class NumberSignal extends ValueSignal<number> {
    *
    * @param delta - The delta to increment the value by. The delta can be
    * negative.
+   * @returns An operation object that allows to perform additional actions.
    */
-  incrementBy(delta: number): void {
+  incrementBy(delta: number): Operation {
     if (delta === 0) {
-      return;
+      return noOperation;
```
taefi, line 45: I'm a bit confused why a `noOperation` should be returned in this case?
If I understand the `noOperation` correctly, the then callback that is passed to a `noOperation` is not executed. 
I would suggest to change this to an implementation of `Operation` interface, that calls the passed-in callback immediately without storing anything in the `thenCallbacks` map, so that something like the following example manages to call the `notifySubscribers` function after the `incrementBy` is done, even in case the result of `calculateAmount()` happened to be `0`:
```
let amount = 1;
if (some condition is satisfied) {
    amount = calculateAmount();
}
numberSignal.incrementBy(amount).result.then(() => notifySubscribers());
```

taefi (changes requested):

packages/ts/react-signals/src/NumberSignal.ts
taefi: Or maybe I'm not understanding the `noOperation` implementation correctly.

cromoteca: It depends on what should happen when 0 is passed as value, or any other "void" value for the other signal operations. It was my understanding that the callback must be run when an update happens, so as nothing changed on the server (it returns immediately without processing), I see no reason to process any callback and, in your example, notify the subscribers.

taefi: Well the user might not know about all the possible outcomes of `calculateAmount` that can happen at runtime, so it will be smoother for them to just chain the `notifySubscribers()` to be done after the `incrementBy`. To me, an `incrementBy(0)` is a successful operation without the need to make the roundtrip to the server.

taefi (dismissed):

packages/ts/react-signals/src/FullStackSignal.ts
```diff
@@ -5,6 +5,21 @@ import { createSetStateEvent, type StateEvent } from './events.js';
...
 const ENDPOINT = 'SignalsHandler';
 
+/**
+ * The type of function that is used to define a callback that is called when
+ * the `then` method is called on an operation object.
+ */
+export type ThenCallback = () => void;
+
+/**
+ * A return type for signal operations.
```
taefi, line 15: Nit:
This can be improved a bit to express the fact that the operation enables the user to chain then callbacks.

taefi:
Also ListSignal's operations need another follow-up PR to enable the thenables, right?

packages/ts/react-signals/src/ValueSignal.ts
```diff
@@ -65,19 +74,18 @@ export class ValueSignal<T> extends FullStackSignal<T> {
    *
    * @param callback - The function that is applied on the current value to
    *                   produce the new value.
-   * @returns An operation subscription that can be canceled.
+   * @returns An operation object that allows to perform additional actions, including cancellation.
```
taefi, line 77:
```suggestion
   * @returns An operation object that allows to perform additional actions,
   *                   including cancellation.
```

taefi (approved):

PR #2801, opened 2024-10-08
chore: Use components 24.5 rc2

Artur-:


taefi:
It seems that with auto-created PR #2805 merged, this PR is not needed anymore.

PR #2807, opened 2024-10-09
fix(signals): subscribe on the fly if needed on updates

### LINKED ISSUE #2701: [Full-stack Signals] Cannot update a signal that you're not subscribed to
### Describe the bug

Trying to manipulate a signal for which you don't have an open subscription will lead to an exception like `java.lang.IllegalStateException: Signal not found for client signal: MXbPB5IqFiJzkzI1v7W0m`.

Aside from offline cases that are described in https://github.com/vaadin/hilla/issues/2653, this does also happen if you want to e.g. increment a counter already before it has been shown. This might be the case if you want to count how many times a view has been opened or if you build a poll where you only see the results after you have submitted your own choice.

### Expected-behavior

Expected that signal operations can be submitted even without an active subscription as long as the access control check passes for the server-side service method. This means that the service method would have to be invoked again for each operation when there's no subscription but that should be fine since we are anyways expecting such methods to be deterministic.

### Reproduction

```tsx
const counter = StatsService.counter();

export default function Counter() {
    const subscribe = useSignal(false);
    
    return <VerticalLayout>
        <Button onClick={() => counter.increment()}>Increment</Button>
        <Checkbox checked={subscribe.value} onCheckedChanged={(event) => subscribe.value = event.detail.value} label="Subscribed" />
        {subscribe.value ? <>Live value: {counter}</> : <>Peeked value: {counter.peek()} </>}
    </VerticalLayout>
}
```

### System Info

Hilla 24.5.0.alpha13

cromoteca:
Fixes #2701


taefi (dismissed):
> Didn't merge as I didn't remember the relation of this fix with the `thenable` feature. See [this](https://github.com/vaadin/hilla/issues/2701#issuecomment-2355441222) comment.

I don't remember the reason of that comment.

taefi (approved):

Issue #2341, opened 2024-04-18
[Binder] Permanent invalid state with optional (text based) fields


Feuerwerk:
### Describe the bug

When validating an optional field which is bound to a text-field based widget (e.g. vaadin-text-field, vaadin-number-field.....) and having a validator registered to that field, once the field gets invalid, this state is permanent.
The field stay red but the error message can go away.

```
protected firstUpdated(_changedProperties: PropertyValues) {
    super.firstUpdated(_changedProperties);
    const model = this.binder.model;
    this.binder.for(model.text).addValidator(new Size({
        message: 'max 5 characters',
        min: 0,
        max: 5,
    }));
}

render() {
    const model = this.binder.model;
    return html`<vaadin-text-field ${field(model.text)}></vaadin-text-field>`;
}
```

As a workaround you can add for example the _required_ attribute to the textfield or override _checkValidity_ with an implementation which always return true to get it working again
`<vaadin-text-field ${field(model.text)} . checkValidity="${() => true}"></vaadin-text-field>`

This seems due to _InputConstraintsMixin.checkValidity_ behaving differently depending on constraints (like _required_ for example) present on the input field. When there is no constraint on the input field _checkValidity_ simply remembers the previous invalid state and keep that.

### Expected-behavior

I expect that the invalid status gets updated every time the field is changed by the user

### Reproduction

Here is a modified starter project which demonstrates the bug. Just follow the instructions in the browser.
The interesting code changes are only in frontend/views/main-view.ts
[skeleton-starter.zip](https://github.com/vaadin/hilla/files/15023949/skeleton-starter.zip)

### System Info

Hilla: 2.5.7

PR #2815, opened 2024-10-11
fix: ignore vaadin field old invalid state

krissvaa:
Fixes: #2341


packages/ts/lit-form/src/Field.ts
```diff
@@ -201,6 +205,17 @@ export class VaadinFieldStrategy<T = any, E extends FieldElement<T> = FieldEleme
...
+  override elementValidity(): boolean {
+    // Ignore the `invalid` property of the Vaadin component to avoid
+    // reading the component's internal old validation state and validate
+    // the element based on the current state.
+    const isElementInvalid = this.element.invalid;
+    this.element.invalid = false;
+    const valid = this.element.checkValidity?.() ?? true;
+    this.element.invalid = isElementInvalid;
+    return valid;
+  }
```
platosha, line 218: Would it be possible to override `checkValidity` here to avoid introducing public API method? 

```suggestion
  override checkValidity(): boolean {
    // Ignore the `invalid` property of the Vaadin component to avoid
    // reading the component's internal old validation state and validate
    // the element based on the current state.
    const isElementInvalid = this.element.invalid;
    this.element.invalid = false;
    const valid = super.checkValidity();
    this.element.invalid = isElementInvalid;
    return valid;
  }
```

packages/ts/lit-form/src/Field.ts
```diff
@@ -120,7 +120,7 @@ export abstract class AbstractFieldStrategy<T = any, E extends FieldElement<T> =
       return true;
     }
 
-    const valid = this.#element.checkValidity();
+    const valid = this.elementValidity();
```
platosha, line 123: Let us try to avoid introducing a method, see suggestions below

```suggestion
    const valid = this.#element.checkValidity();
```

packages/ts/lit-form/src/Field.ts
```diff
@@ -129,6 +129,10 @@ export abstract class AbstractFieldStrategy<T = any, E extends FieldElement<T> =
     return valid;
   }
 
+  elementValidity(): boolean {
+    return this.#element.checkValidity?.() ?? true;
+  }
+
```
platosha, line 135: Let us try to avoid introducing a method, see suggestions below


```suggestion

```

platosha (changes requested):

platosha (approved):

Issue #2809, opened 2024-10-09
Handling of @Id and @Version fields in AutoCrud could be improved


rbrki07:
### Describe the bug

When you have a JPA entity with a field annotated with `@Id` or `@Version` and you use this entity in the AutoCrud component, you likely won't be able to submit the form of the AutoCrud component. You can fill the form and press submit, but nothing will happen, and you won't see any errors in the frontend or backend.

Hilla validates these fields already on the client, but the error is never shown because AutoCrud / AutoForm hide fields for id and version automatically. So there are client-side validation errors, but they are not visible. The values for id and version fields are generated only on the server after saving an entity.

A workaround is to add the `@com.vaadin.hilla.Nullable` annotation to the id and/or version field of the JPA entity, but this feels strange. From a developer perspective, I want this fields to be not empty nor null, and I know, that JPA will take care of this.

### Expected-behavior

Using the annotations `@Id` and `@Version` is quite common. As a developer, I want to use entities with these annotations in the AutoCrud component with a good developer experience. This could mean:
* Getting an error/warning when the validation of the hidden id and/or version field in the AutoCrud components fails. This could be a `console.error` or a notification or a popup/popover.
* AutoCrud could skip validation for fields that are hidden per default. 

### Reproduction

* Create a new Hilla project using start.vaadin.com or Vaadin CLI
* Add an entity with id and/or version field and `@Id` and/or `@Version` annotations
* Create a JpaRepository and CrudRepositoryService for that entity
* Create a view, add the AutoCrud component to that view and configure the component to use the generated model and service of that entity.
* Try to add a new item with the form component of the AutoCrud component

### System Info

Vaadin 24.5.0.beta6

rbrki07:
Origin: https://vaadin.com/forum/t/hilla-autocruds-submmit-button-wont-persist-data-on-database/166588

platosha:
See #2217 specifically about the errors from hidden fields. Besides that, we could assume `@Nullable` for such `@Id` and `@Version` fields, allowing the value to be undefined on the client side.

rbrki07:
I just realized that I was able to use entities with id and/or version fields together with AutoCrud and AutoForm in a previous app using Vaadin 24.4. Did this behaviour change on purpose?

rbrki07:
I think I need to add some more context. I detected a typo in my example code
```java
@Id
@GeneratedValue(strategy = GenerationType.IDENTITY)
private long id;
```

[...]

Issue #2217, opened 2024-03-13
Errors related to non-rendered fields are ignored by AutoForm


Artur-:
### Describe the bug

If you have an id field like
```
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;
```

then the id field is not rendered in an AutoForm but trying to submit the form results in a validation error:
```
{message: 'must be a number', property: 'id', validator: IsNumber, value: NaN}
```

then that error message is ignored because it has a `property` set, so the user can never see it.

### Expected-behavior

Error message related to non-rendered fields (if they ever occur) would be rendered as error messages that are not related to a field

### Reproduction

```
<AutoCrud service={ProductService} model={ProductModel}></AutoCrud>;
``

with 
```
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;
```

in Product

### System Info

Vaadin 24.4 alpha20

platosha:
As a workaround, you can customize the form and render errors for specific fields in the custom renderer.

PR #2833, opened 2024-10-15
fix: show autoform non-rendered property errors

krissvaa:
Show autoform non-rendered property errors
Make @Id and @Version annotations resolve as nullable fields

Fixes: #2217
Fixes: #2809

### DIFF
--- a/packages/java/parser-jvm-plugin-nonnull/src/test/java/com/vaadin/hilla/parser/plugins/nonnull/nullable/NullableEndpoint.java
+++ b/packages/java/parser-jvm-plugin-nonnull/src/test/java/com/vaadin/hilla/parser/plugins/nonnull/nullable/NullableEndpoint.java
@@ -1,0 +1,20 @@
+package com.vaadin.hilla.parser.plugins.nonnull.nullable;
+
+
+import jakarta.persistence.Id;
+import jakarta.persistence.Version;
+
+@Endpoint
+public class NullableEndpoint {
+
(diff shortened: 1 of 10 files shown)


packages/java/parser-jvm-plugin-nonnull/src/test/java/com/vaadin/hilla/parser/plugins/nonnull/nullable/NullableEndpoint.java
```diff
@@ -0,0 +1,20 @@
...
+@Endpoint
+public class NullableEndpoint {
+
+    public NullableFieldModel nullableFieldModel(
+            NullableFieldModel nullableFieldModel) {
+        return nullableFieldModel;
+    }
+
+    public static class NullableFieldModel {
+        @Id
```
cromoteca, line 15: Those fields look nullable by default, unless I'm missing something. You should probably test them under a `@NonNullApi` annotation.

cromoteca (changes requested):

packages/java/parser-jvm-plugin-nonnull/src/test/java/com/vaadin/hilla/parser/plugins/nonnull/nullable/NullableEndpoint.java
krissvaa: I should, but I should also check that they are indeed nullable without any other annotation. 
I will add another test.

platosha (approved):

Issue #2825, opened 2024-10-14
Menu entries for Flow routes are not populated to dynamic menu


mshabarov:
### Describe the bug

`createMenuItems` does not contain entries to Flow routes, which leads to absent menu items in Hilla's main layout, when it uses dynamic menu API.

### Expected-behavior

A link for Flow route is available.

### Reproduction

Download and start the project from https://github.com/vaadin/flow-hilla-hybrid-example/tree/v24.5.
Login as admin/admin.
Notice that the menu doesn't have a Flow menu item.

### System Info

Vaadin 24.5.0.rc1.

PR #2836, opened 2024-10-15
fix: Check main layout presence separately (#2831)(CP: main)

mshabarov:
## Description

MenuRegistry::collectClientViews erases/overrides the layout entry by the view entry, if they have the same mapping, e.g. when you add @layout.tsx and @index.tsx in the same folder.

Thus, the ServerAndClientViewsProvider::hasMainMenu method may not get all the layouts in the input collection.

That leads to hasMainMenu=false and to skipping menu entries for Flow routes.

This patch uses https://github.com/vaadin/flow/pull/20245 new method to detect top-level main layout that fulfils the same criteria: empty route, non-null children, no other root elements.

Related-to https://github.com/vaadin/hilla/issues/2825

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


taefi (approved):
LGTM.
There is only a consideration about using the already existing `deploymentConfiguration` instance in `ServerAndClientViewsProvider` instead of getting it from the VaadinService (from request). This is discussed with @mshabarov already, and since the original PR has been merged, this can also be merged, so that the subsequent PR fixing this can also be CPed automatically.

PR #2842, opened 2024-10-16
fix(push): track `onClose` Atmosphere events to properly reconnect and resubscribe

### LINKED ISSUE #2834: Flux connection is not properly resumed in some cases
### Describe the bug

A Flux subscription is not resumed after disconnecting using the "Offline" preset in Chrome dev tools.

It seems like the underlying WebSocket is reopened but `onSubscriptionLost` is never called which in turn means that the subscription is never reopened. I did also observe similar symptoms in real-world scenarios with other browsers but there's no easy deterministic reproduction for those since Firefox keeps WebSocket connections open when enabling "offline" mode and Safari doesn't even have an offline emulation mode.

The `@Push` functionality in Flow does automatically handle a corresponding situation so we're probably just missing some Atmosphere event to listen to.

### Expected-behavior

Expected that the subscription could be restarted after being offline.

### Reproduction

* Server-side endpoint with `return EndpointSubscription.of(Flux.interval(Duration.ofSeconds(1)), () -> System.out.println("Unsubscribe"))` 
* Client-side subscription with `onSubscriptionLost(() => { console.log("Lost"); return ActionOnLostSubscription.RESUBSCRIBE})`
* Enable the "offline" preset in Chrome dev tools. Wait for server-side log. Disable the "offline" preset. Observe that no client-side message is logged

### System Info

Originally observed with Vaadin 24.5.0.beta4

### LINKED ISSUE #2841: reconnection callback is called twice
### Describe the bug

https://github.com/vaadin/hilla/pull/2782 introduced a new connection state management, but `onReconnect` will inadvertently run the user-defined reconnect callback to get its return type.
The same callback is already called from `onReopen`.

### Expected-behavior

_No response_

### Reproduction

Look at https://github.com/vaadin/hilla/commit/9bafcb7a5c15810c522d085321dbdacf3b2f971f#diff-a041bf6ab4a2971d34fa718ca88626d7100fa4ffec35a9260bf65669deeb0a59R230

### System Info

24.6.0.alpha1

cromoteca:
Adds a new status to track if the channel was closed and trigger reconnection callback only once, but on different events, as the event chain is different according to how the connection disruption happens.

Detected chains are:

```
onClose => onReconnect => onReopen
onClose => onOpen
```

Fixes #2834.
Fixes #2841.


krissvaa:
Everything looks good
The only thing I am wondering is if this scenario is correct:
if the subscription `reconnect` is set to `REMOVE` or empty, and then the connection closes. The Atmosphere will try to reconnect so the subscription state will be `CONNECTING` until `reOpen` when the state will be set to `CLOSED`.  
(onClose => onReconnect => onReopen scenario)
As a developer, I wouldn't understand why subscription connection is not just closing but connecting and closing.

cromoteca:
> As a developer, I wouldn't understand why subscription connection is not just closing but connecting and closing.

Good point, but I don't know how to solve it: the reconnection callback is a function and we haven't the right to call it just to get the return value. So we can't know its value before running it, and to run it we need to reconnect first.

krissvaa (approved):

PR #2844, opened 2024-10-17
fix: add useViewConfigTitle utility

tltv:
Adds useViewConfigTitle function to read title from current route's ViewConfig or window.Vaadin.documentTitleSignal or given string.

RelatedTo: vaadin/flow#20207


platosha (changes requested):
Let's consider this before merging: https://github.com/vaadin/flow/issues/20207#issuecomment-2434938751

PR #2848, opened 2024-10-18
fix(parser): add position information to type signature node

### LINKED ISSUE #2847: Parser is unable to deal with identical type arguments
### Describe the bug

When two or more type arguments are identical, the parser throws an exception indicating that the number of items in the schema is different from the number of type parameters: `java.lang.IllegalStateException: Number of parameters mismatch for Root(ScanResult)/...`

This happens because, when adding child nodes, `PluginExecutor` doesn't enqueue identical paths, and the path corresponding to a type argument has no information about its position, so they are identical if the type is the same.

See https://github.com/vaadin/hilla/blob/f88c7e1546cea27a17462aa420de781d6d454d3d/packages/java/parser-jvm-core/src/main/java/com/vaadin/hilla/parser/core/PluginExecutor.java#L42-L44

### Expected-behavior

TypeScript is generated correctly.

To make this happen, the path for a type argument needs to contain information about its position.

### Reproduction

Consider this endpoint:
```java
@BrowserCallable
public class DummyService {
    public record Pair<T1, T2>(T1 value1, T2 value2) {}
    public void two(Pair<String, String> values) {}
}
```

### System Info

Hilla 24.5.0 ... 24.6.0.alpha1

cromoteca:
Adds position information to `TypeSignatureNode` so that identical type arguments are retained thanks to their different position.

Fixes #2847

### DIFF
--- a/packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/TypeSignaturePlugin.java
+++ b/packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/TypeSignaturePlugin.java
@@ -338,17 +338,15 @@ public final class TypeSignaturePlugin
     private NodeDependencies scanTypeSignature(TypeSignatureNode node,
             NodeDependencies nodeDependencies) {
         var signature = node.getSource();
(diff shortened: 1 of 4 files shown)


taefi (approved):
LGTM!

PR #2853, opened 2024-10-21
fix: handle child layouts

caalador:
Handle child layouts depending
on if they have flowLayout true or false.
Fixes all layouts having flowLayout
if the parent layout or one child has
flowLayout true.

Fixes vaadin/flow#20261


mshabarov:
@platosha could you or someone else from Hilla team please review this pull request before we lost the memories what the topic was about 😄

Lodin (approved):

PR #2882, opened 2024-10-31
fix(form): support non-UTC timezone date time in VaadinDateTimePicker

platosha:
Existing for VaadinDateTimePickerStrategy failed when running in non-UTC timezone environment. This change addresses the bug and makes the form library tests always use a non-UTC timezone.


krissvaa (approved):
LGTM

PR #2894, opened 2024-11-05
fix(crud): use 3-params overload for dates instead of separate assignments

### LINKED ISSUE #2887: Date picker AutoCrud, AutoGrid  and AutoForm broken when the date is 31st of October
### Description

No months can be used that only have 30 days on the 31 of October, setting the date of the computer to the 1st of November resolves the issue.

The time of existing items being edited also jumps back 2 hours.

Also reported this on the forum: https://vaadin.com/forum/t/autocrud-grid-form-date-time-picker-doesnt-allow-to-select-september-2024/167670/6

### Expected outcome

I get the date I selected.

All 3 Auto doc pages contain a date time picker where the bug can be reproduced.

https://vaadin.com/docs/latest/components/auto-crud
https://vaadin.com/docs/latest/components/auto-grid
https://vaadin.com/docs/latest/components/auto-form

### Steps to reproduce

1. Set date of computer to 31 November 2024
2. Try to select a month with only 30 days
3. Result = another month

### Environment

Vaadin version(s): 24.5.2
OS: Windows

### Browsers

Chrome, Firefox, Edge

cromoteca:
Assigning year, month, and day, in this order, is problematic when the chosen month has less days than the current day of month.

So, for example, on Oct. 31st 2024, choosing Sept. 15th 2024 leads to 2024 -> Sept. 31 -> Oct. 1 -> Oct. 15.

Passing the 3 parameters at once avoids this problem.

Fixes #2887

### DIFF
--- a/packages/ts/react-crud/src/locale.ts
+++ b/packages/ts/react-crud/src/locale.ts
@@ -66,7 +66,8 @@ export class LocaleFormatter {
   formatDate(value?: DatePickerDate | string): string {
     if (typeof value === 'object') {
       const { year, month, day } = value;
-      const date = new Date(year, month, day);
+      const date = new Date();
+      date.setFullYear(year, month, day);
       return this.#date.format(date);
     }
 
@@ -96,7 +97,8 @@ export class LocaleFormatter {
     const day = Number(match?.groups?.day);
 
     // Verify that the parsed date is valid
-    const dateInstance = new Date(year, month, day);
+    const dateInstance = new Date();
+    dateInstance.setFullYear(year, month, day);
 
     if (dateInstance.getFullYear() !== year || dateInstance.getMonth() !== month || dateInstance.getDate() !== day) {
       return undefined;
--- a/packages/ts/react-crud/src/locale.ts
+++ b/packages/ts/react-crud/src/locale.ts
@@ -66,10 +66,7 @@ export class LocaleFormatter {
   formatDate(value?: DatePickerDate | string): string {
     if (typeof value === 'object') {
       const { year, month, day } = value;
-      const date = new Date();
-      date.setFullYear(year);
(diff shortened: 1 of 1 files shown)


packages/ts/react-crud/src/locale.ts
```diff
@@ -67,9 +67,7 @@ export class LocaleFormatter {
     if (typeof value === 'object') {
       const { year, month, day } = value;
       const date = new Date();
-      date.setFullYear(year);
-      date.setMonth(month);
-      date.setDate(day);
+      date.setFullYear(year, month, day);
```
platosha, line 70: Could be shortened down to `new Date(year, month, day);` I guess?

```suggestion
      const date = new Date(year, month, day);
```

packages/ts/react-crud/src/locale.ts
```diff
@@ -100,9 +98,7 @@ export class LocaleFormatter {
 
     // Verify that the parsed date is valid
     const dateInstance = new Date();
-    dateInstance.setFullYear(year);
-    dateInstance.setMonth(month);
-    dateInstance.setDate(day);
+    dateInstance.setFullYear(year, month, day);
```
platosha, line 101: Could be shortened down to `new Date(year, month, day);` I guess?

```suggestion
      const dateInstance = new Date(year, month, day);
```

platosha (changes requested):

packages/ts/react-crud/src/locale.ts
cromoteca: No, it can't. That's what I did at the beginning, but year XX is then converted to 19XX and doesn't pass a test.

DuncanCasteleyn (approved):

platosha (approved):

PR #2901, opened 2024-11-11
fix: allow dynamically import atmosphere.js for SW context

### LINKED ISSUE #2867: PWA endpoint errors while called from SW
### Describe the bug

Some issues stop endpoints from being called from the Service Worker. 
Most of the issues are that `window` or `document` references are not available in the worker context. 
Which should be quite easy to solve: 

- https://github.com/vaadin/hilla/blob/c7d4707995f2235574f48d84a8619ab038beb503/packages/ts/frontend/src/Connect.ts#L21
- https://github.com/vaadin/hilla/blob/c7d4707995f2235574f48d84a8619ab038beb503/packages/ts/frontend/src/Connect.ts#L303
- https://github.com/vaadin/hilla/blob/c7d4707995f2235574f48d84a8619ab038beb503/packages/ts/frontend/src/CookieManager.ts#L8

Then there are related to the same issue but more complicated issues:
1. `__REGISTER__()` includes references to `window` - but is defined in a different place    
https://github.com/vaadin/hilla/blob/c7d4707995f2235574f48d84a8619ab038beb503/packages/ts/frontend/src/index.ts#L8C1-L8C16
2. `ConnectionState` and `ConnectionIndicator `from `@vaadin/common-frontend` also have references to `window` and need to be either fixed in their repo or clone+fixed in hilla separately
https://github.com/vaadin/hilla/blob/c7d4707995f2235574f48d84a8619ab038beb503/packages/ts/frontend/src/Connect.ts#L2

3. Lastly there is an issue with atmosphere.js and default exports to Vite. Atmosphere dependency could be excluded by SW Vite plugin or by finding a better fix for exclusion as I am not sure about this approach. 

> RollupError: node_modules/@vaadin/hilla-frontend/FluxConnection.js (1:7): "default" is not exported by "node_modules/atmosphere.js/lib/browser.js", imported by "node_modules/@vaadin/hilla-frontend/FluxConnection.js".
>     at getRollupError (file:///app/node_modules/rollup/dist/es/shared/parseAst.js:396:41)
>     at error (file:///app/node_modules/rollup/dist/es/shared/parseAst.js:392:42)
>     at Module.error (file:///app/node_modules/rollup/dist/es/shared/node-entry.js:13858:16)
>     at Module.traceVariable (file:///app/node_modules/rollup/dist/es/shared/node-entry.js:14306:29)
>     at ModuleScope.findVariable (file:///app/node_modules/rollup/dist/es/shared/node-entry.js:11984:39)
>     at ChildScope.findVariable (file:///app/node_modules/rollup/dist/es/shared/node-entry.js:7432:38)
>     at ClassBodyScope.findVariable (file:///app/node_modules/rollup/dist/es/shared/node-entry.js:7432:38)
>     at ChildScope.findVariable (file:///app/node_modules/rollup/dist/es/shared/node-entry.js:7432:38)
>     at ChildScope.findVariable (file:///app/node_modules/rollup/dist/es/shared/node-entry.js:7432:38)
>     at FunctionScope.findVariable (file:///app/node_modules/rollup/dist/es/shared/node-entry.js:7432:38)

Missing CSRF token in calling endpoints is separated in an another issue: #2791

### Expected-behavior

All DOM `window` or `document` references should be prefixed with `self.window && window...` or `self.document && document...` (or some alternative) so that SW could run without encountering them.

### Reproduction

[sw-app.zip](https://github.com/user-attachments/files/17504344/sw-app.zip)

### System Info

Hilla 24.5

krissvaa:
Fixes #2867


packages/ts/frontend/src/vite-env.d.ts
```diff
@@ -0,0 +1,10 @@
+// / <reference types="vite/client" />
```
Lodin, line 1:
```suggestion
/// <reference types="vite/client" />
```

packages/ts/frontend/src/FluxConnection.ts
```diff
@@ -71,6 +71,16 @@ type EndpointInfo = {
...
 
+let atmosphere: Atmosphere.Atmosphere | undefined;
+
+(async () => {
+  if (!import.meta.env?.VITE_SW_CONTEXT) {
+    atmosphere = await import('atmosphere.js').then((module) => module.default);
+  }
+})().catch((e) => {
+  console.error('Failed to load atmosphere.js', e);
+});
```
Lodin, line 82:
```suggestion
if (!import.meta.env?.VITE_SW_CONTEXT) {
  try {
    atmosphere = await import('atmosphere.js').then((module) => module.default);
  } catch (e: unknown) {
    console.error('Failed to load atmosphere.js', e);
  }
}
```

I'd suggest using top-level await here. Is it possible?

Lodin (changes requested):

packages/ts/frontend/src/FluxConnection.ts
Lodin: Also, we can probably check the presence of the `document` in `globalThis` here, instead of `VITE_SW_CONTEXT`.

krissvaa: top-level await is not allowed according to our project configuration.
> Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)

platosha: Let's avoid top-level await for now.

Lodin: But that creates a possible race condition, and in some cases the `FluxConnection` can possibly be created before the Atmosphere is loaded.

platosha (approved):

platosha:
Investigation revealed that production builds get broken without these workarounds for Vite/Rollup: https://github.com/vaadin/flow/pull/21276

platosha (dismissed):

platosha (approved):

PR #2915, opened 2024-11-20
feat(file-router): add skipLayouts feature

### LINKED ISSUE #2385: Add feature for rendering an FS router view outside the main layout
### Describe your motivation

Some special views, such as a login view, might need to be rendered without any wrapping layout. The way of doing this right now is to copy `generated/routes.tsx` and customize it but this is not very convenient.

Add a property to `ViewConfig` specifically for this case.

```
  skipLayouts: true,
  menu: { exclude: true }
}
```

* There has been an idea to add a property to `ViewConfig` that would allow defining an arbitrary layout (or no layout at all) for a view. This would be relatively complex to map to the React Router configuration structure which is why I instead suggest defining a shorthand specifically for having no layout at all.
* There has been an idea to add support for `@phantom/` directory that would have its own layout without contributing to the path. This would map almost directly to the React Router configuration structure but it would also mean that the filesystem structure wouldn't match the URL structure and you would have to move around all existing views if you want to make one view skip the layout.

The suggested `skipLayouts` property could remain as a shorthand even if we implement either of the more flexible alternatives later on.

### Additional context

_No response_

Lodin:
Resolves #2385.


packages/ts/file-router/src/runtime/RouterConfigurationBuilder.ts
```diff
@@ -23,34 +24,89 @@ interface RouteBase {
...
 }
 
-type RoutesModifier = (routes: readonly RouteObject[] | undefined) => readonly RouteObject[] | undefined;
-
 function isReactRouteModule(module?: Module): module is RouteModule<ComponentType> | undefined {
   return module ? 'default' in module && typeof module.default === 'function' : true;
 }
 
-type RouteTransformer<T> = (
+export type RouteGroup = readonly RouteObject[];
```
platosha, line 31: Let's name it `RouteList`.

platosha (approved):

PR #2926, opened 2024-11-26
chore: add web-test-runner for react-auth

Lodin:
Fixes test failures in #2901 by replacing Karma Test Runner (deprecated) with Web Test Runner.

Web Test Runner is applied only to the `react-auth` package. The plan is to incrementally transfer other browser-based packages over time.

### DIFF
--- a/karma.config.cjs
+++ b/karma.config.cjs
@@ -1,3 +1,4 @@
+/* eslint-disable import/unambiguous */
 const { parseArgs } = require('node:util');
 const { basename } = require('node:path');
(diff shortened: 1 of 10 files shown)


Lodin:
Please ignore SonarCloud results. The code duplication is expected because there is some files moved out of `karma.config.js` and placed in TypeScript files. Since Karma is unable to work with TS, I had to keep both versions. When we move to WTR completely, the duplication will be gone.

platosha (approved):

PR #2934, opened 2024-11-28
fix(file-router): allow extension to apply on a route wrapper

Lodin:
In case `layouts.json` contains a module extension (`flowLayout: true`) for a specific route that is not represented by either regular file or a `@layout.tsx`, we should be able to apply that extension on the route. This PR allows the route `module` to be null though still have the extension.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7635,6 +7635,16 @@
         "chai": ">= 2.1.2 < 6"
       }
     },
+    "node_modules/chai-deep-equal-ignore-undefined": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/chai-deep-equal-ignore-undefined/-/chai-deep-equal-ignore-undefined-1.1.1.tgz",
+      "integrity": "sha512-BE4nUR2Jbqmmv8A0EuAydFRB/lXgXWAfa9TvO3YzHeGHAU7ZRwPZyu074oDl/CZtNXM7jXINpQxKBOe7N0P4bg==",
+      "dev": true,
+      "license": "MIT",
+      "peerDependencies": {
+        "chai": ">= 4.0.0 < 5"
+      }
+    },
     "node_modules/chai-dom": {
       "version": "1.12.0",
       "resolved": "https://registry.npmjs.org/chai-dom/-/chai-dom-1.12.0.tgz",
@@ -19351,6 +19361,7 @@
         "@types/sinon": "^10.0.17",
         "@types/sinon-chai": "^3.2.12",
         "chai-as-promised": "^7.1.1",
+        "chai-deep-equal-ignore-undefined": "^1.1.1",
         "chai-fs": "^2.0.0",
         "chai-like": "^1.1.1",
         "deep-equal-in-any-order": "^2.0.6",
--- a/packages/ts/file-router/package.json
+++ b/packages/ts/file-router/package.json
@@ -26,7 +26,7 @@
     "build:copy": "cd src && copyfiles *.d.ts **/*.d.ts ..",
     "lint": "eslint src test",
     "lint:fix": "eslint src test --fix",
-    "test": "mocha test/**/applyLayouts.spec.{ts,tsx} --config ../../../.mocharc.cjs",
+    "test": "mocha test/**/*.spec.{ts,tsx} --config ../../../.mocharc.cjs",
     "test:coverage": "c8 --experimental-monocart -c ../../../.c8rc.json npm test",
     "typecheck": "tsc --noEmit"
   },
@@ -70,6 +70,7 @@
     "@types/sinon": "^10.0.17",
     "@types/sinon-chai": "^3.2.12",
     "chai-as-promised": "^7.1.1",
+    "chai-deep-equal-ignore-undefined": "^1.1.1",
     "chai-fs": "^2.0.0",
     "chai-like": "^1.1.1",
     "deep-equal-in-any-order": "^2.0.6",
--- a/packages/ts/file-router/package.json
+++ b/packages/ts/file-router/package.json
@@ -26,7 +26,7 @@
     "build:copy": "cd src && copyfiles *.d.ts **/*.d.ts ..",
     "lint": "eslint src test",
     "lint:fix": "eslint src test --fix",
-    "test": "mocha test/**/*.spec.{ts,tsx} --config ../../../.mocharc.cjs",
+    "test": "mocha test/**/RouterConfigurationBuilder.spec.{ts,tsx} --config ../../../.mocharc.cjs",
     "test:coverage": "c8 --experimental-monocart -c ../../../.c8rc.json npm test",
     "typecheck": "tsc --noEmit"
   },
--- a/packages/ts/file-router/package.json
+++ b/packages/ts/file-router/package.json
@@ -26,7 +26,7 @@
     "build:copy": "cd src && copyfiles *.d.ts **/*.d.ts ..",
     "lint": "eslint src test",
     "lint:fix": "eslint src test --fix",
-    "test": "mocha test/**/RouterConfigurationBuilder.spec.{ts,tsx} --config ../../../.mocharc.cjs",
+    "test": "mocha test/**/applyLayouts.spec.{ts,tsx} --config ../../../.mocharc.cjs",
     "test:coverage": "c8 --experimental-monocart -c ../../../.c8rc.json npm test",
     "typecheck": "tsc --noEmit"
   },
--- a/packages/ts/file-router/src/runtime/RouterConfigurationBuilder.ts
+++ b/packages/ts/file-router/src/runtime/RouterConfigurationBuilder.ts
@@ -25,7 +25,14 @@ interface RouteBase {
 }
 
 function isReactRouteModule(module?: Module): module is RouteModule<ComponentType> | undefined {
-  return module ? 'default' in module && typeof module.default === 'function' : true;
+  if (!module) {
+    return true;
+  }
+
+  return (
+    ('default' in module && typeof module.default === 'function') ||
+    ('config' in module && typeof module.config === 'object')
+  );
 }
 
 export type RouteList = readonly RouteObject[];
@@ -126,7 +133,9 @@ export class RouterConfigurationBuilder {
       if (added) {
         const { module, path, flowLayout } = added;
         if (!isReactRouteModule(module)) {
-          throw new Error(`The module for the "${path}" section doesn't have the React component exported by default`);
+          throw new Error(
+            `The module for the "${path}" section doesn't have the React component exported by default or a ViewConfig object exported as "config"`,
+          );
         }
 
         const element = module?.default ? createElement(module.default) : undefined;
--- a/packages/ts/file-router/src/runtime/RouterConfigurationBuilder.ts
+++ b/packages/ts/file-router/src/runtime/RouterConfigurationBuilder.ts
@@ -212,6 +212,8 @@ export class RouterConfigurationBuilder {
                 children: server.length + ambivalent.length > 0 ? [...server, ...ambivalent] : undefined,
               } as RouteObject);
             } else if (server.length > 0) {
+              // Even if the route doesn't have the flag, it goes to the server
+              // list if any of the children has the flag enabled.
               lists.server.push({
                 ...route,
                 children: server,
@@ -228,10 +230,10 @@ export class RouterConfigurationBuilder {
               } as RouteObject);
             }
 
+            // The route without the flag go to the `default` list. Then it will
+            // be moved to either server or client list based on the parent
+            // route.
             if (flag === undefined && lists.server.every(({ path }) => path !== route.path)) {
-              // The route without the flag go to the `default` list. Then it will
-              // be moved to either server or client list based on the parent
-              // route.
               lists.ambivalent.push({
                 ...route,
                 children: ambivalent.length > 0 ? ambivalent : undefined,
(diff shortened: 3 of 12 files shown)


caalador:
The test project `vaadin-platform-react-hybrid-test` doesn't with these changes render anything.

Client fails with:
```
 Uncaught Error: The module for the "flow" section doesn't have the React component exported by default
    at @vaadin_hilla-file-router_runtime__js.js?v=7b932dbe:218:17
    at @vaadin_hilla-file-router_runtime__js.js?v=7b932dbe:371:41
    at Array.map (<anonymous>)
    at @vaadin_hilla-file-router_runtime__js.js?v=7b932dbe:371:26
    at transformTree (@vaadin_hilla-file-router_runtime__js.js?v=7b932dbe:138:10)
    at @vaadin_hilla-file-router_runtime__js.js?v=7b932dbe:345:27
    at @vaadin_hilla-file-router_runtime__js.js?v=7b932dbe:384:72
    at Array.reduce (<anonymous>)
    at RouterConfigurationBuilder.build (@vaadin_hilla-file-router_runtime__js.js?v=7b932dbe:384:51)
    at routes.tsx:69:6
```
Also still only the parent routes are checked. This was originally fixed and working in 
https://github.com/vaadin/hilla/pull/2853/commits/e2d86cf4b985841e30f23439869da428711e8baa before the refactoring.

caalador:
So this now creates
```
    createRoute_1("flow", extendModule_1(null, { "flowLayout": true }), [
        createRoute_1("hello-hilla", Page_2)
    ]),
```
when it should actually create:
```
    createRoute_1("flow", null) [
        createRoute_1("hello-hilla", extendModule_1(Page_2, { "flowLayout": true })
    ]),
```
So we should handle leaves and not roots when setting the layout flag.
Also the generated routes in the browser should support splitting to 2 sameish trees with different leaves when there are different requirements on layout or not.

So for instance if we have `/flow/hello` and `/flow/nono` where `nono` has `flowLayout: false` we should get `[serverElement]/flow/hello` and `/flow/nono`.

Lodin:
Yes, that behavior is respected in the new implementation, as you can see from [these](https://github.com/vaadin/hilla/blob/main/packages/ts/file-router/test/runtime/RouterConfigurationBuilder.spec.tsx#L526) tests.

It splits the tree to three separate trees:

1. The `skipLayout` routes (without Flow layout).
2. The `flowLayout` routes (with Flow layout).
3. The regular routes.

The algorithm works the following way:

1. If the route has the mark, it along with children is moved to a specific list.
2. If the route is leaf without the mark, it goes to a regular list.
3. Then two previous steps are applied to all remaining children.

So, according to it, the `flowLayout: true` parent should move the whole branch to a `flowLayout` tree. And if it breaks, I believe there is something specific happen...

Lodin:
Yes, so do I 🙂
I actually created a patch that fixes the error, please take a look if it works (it works for me).
cc @caalador @ZheSun88

ZheSun88:
ah yeah.. seems UI gets loaded for me and the side-nav is there.. 
I will let @caalador to review this PR, as he knows more about this issue.

caalador:
With the latest changes the default project seemingly works, but if I add a `flowLayout: false` view into views/flow/ i still see
```
    createRoute_1("flow", extendModule_1(null, { "flowLayout": true }), [
        createRoute_1("hello-hilla", Page_2),
        createRoute_1("no-layout", Page_3)
    ]),
```
Which is wrong, also after that the hilla pages do not render anymore and clicking in from localhost:8080/flow you can see `chunk-6C7XMPYY.js?v=7b932dbe:214 Matched leaf route at location "/flow/" does not have an element or Component. This means it will render an <Outlet /> with a null value by default resulting in an "empty" page.` in the console.

Also looking at the file-routes.json there is only `"flowLayout": false,` and no route is marked as `"flowLayout": true,` so this will make many thing work wrong as the json is checked for data. the server for instance cancels server rendering for client routes that are not `flowlayout: true` 
```
// If navigation target is Hilla route, terminate Flow navigation logic
// here.
```

caalador:
Also in the current version if I make a view in a sub directory: `views/home/deep/hello.tsx` and have a DeepLayout.java that has the annotation `@Layout("/home/deep")`  this should get `flowLayout: true`, but looking at the
ts file
```
    createRoute_1("home", [
        createRoute_1("deep", [
            createRoute_1("hello", Page_7)
        ])
    ])
``` 
and the json
```
  {
    "route": "home",
    "params": {},
    "children": [
      {
        "route": "deep",
        "params": {},
        "children": [
          {
            "route": "hello",
            "menu": {
              "title": "Hello React in Flow Layout"
            },
            "title": "Hilla in Flow",
            "flowLayout": false,
            "params": {}
          }
        ]
      }
    ]
  }
```

so no flow layout available even if it should be.
The paths checked for layout is also
```
== ''
== 'flow'
== 'hilla'
== 'home'
```
So the full child path is never checked, making any multipart layout not work. 
In this instance the check is for parts [home, deep] and there should be the child route [home, deep, hello] but only [home] is checked for.

caalador:
As a simple test target the platform file-routes.json should be
```
[
  {
    "route": "",
    "menu": {
      "title": "root"
    },
    "flowLayout": false,
    "params": {},
    "title": "Public"
  },
  {
    "route": "flow",
    "params": {},
    "children": [
      {
        "route": "hello-hilla",
        "menu": {
          "title": "Hello React in Flow Layout"
        },
        "title": "Hilla in Flow",
        "loginRequired": true,
        "rolesAllowed": [
          "ROLE_USER",
          "ROLE_ADMIN"
        ],
        "flowLayout": true,
        "params": {}
      }
    ]
  },
  {
    "route": "hilla",
    "flowLayout": false,
    "params": {},
    "title": "Layout",
    "children": [
      {
        "route": "",
        "menu": {
          "exclude": true,
          "title": "ERROR!"
        },
        "flowLayout": false,
        "params": {},
        "title": "Hilla"
      },
      {
        "route": "admin",
        "menu": {
          "title": "Hello Admin"
        },
        "loginRequired": true,
        "rolesAllowed": [
          "ROLE_ADMIN"
        ],
        "flowLayout": false,
        "params": {},
        "title": "Hello React"
      },
      {
        "route": "hello-react",
        "menu": {
          "title": "Hello World React"
        },
        "flowLayout": false,
        "params": {},
        "title": "User"
      },
      {
        "route": "user",
        "menu": {
          "title": "Hello User"
        },
        "loginRequired": true,
        "rolesAllowed": [
          "ROLE_USER"
        ],
        "flowLayout": false,
        "params": {},
        "title": "Hello React"
      }
    ]
  },
  {
    "route": "login",
    "menu": {
      "exclude": true
    },
    "flowLayout": false,
    "params": {},
    "title": "Login"
  }
]
```
and the old file-routes.ts was 
```
const routes: readonly AgnosticRoute[] = [
    createRoute("", false, Page0),
    createRoute("flow", true, [
        createRoute("", true, Page1),
        createRoute("hello-hilla", true, Page2)
    ]),
    createRoute("hilla", false, Layout8, [
        createRoute("", false, Page4),
        createRoute("admin", false, Page5),
        createRoute("hello-react", false, Page6),
        createRoute("user", false, Page7)
    ]),
    createRoute("login", false, Page9)
];
```
So if the `extendModule_1` hits the correct spots with true we should have quite the correct resolution.

caalador:
Still getting the wrong values in the `file-routes.json` as no route is marked as `flowLayout: true` so going directly to the route an empty page is displayed. (starting from a flowlayout flow route the hilla routes are for some reason rendered)
Also the sub route `home/deep/hello` is in no case rendered under the flow layout even though it should.
`file-routes.ts` also still generates
```
    createRoute_1("", Page_1),
    createRoute_1("flow", extendModule_1(null, { "flowLayout": true }), [
        createRoute_1("hello-hilla", Page_2),
        createRoute_1("no-layout", Page_3)
    ]),
    createRoute_1("hilla", Layout_1, [
        createRoute_1("", Page_4),
        createRoute_1("components", Page_5),
        createRoute_1("hello-react", Page_6)
    ]),
    createRoute_1("home", [
        createRoute_1("deep", [
            createRoute_1("hello", Page_7)
        ])
    ])
```
when it should look like:
```
    createRoute_1("", Page_1),
    createRoute_1("flow",  [
        createRoute_1("hello-hilla", extendModule_1(Page_2, { "flowLayout": true })),
        createRoute_1("no-layout", Page_3)
    ]),
    createRoute_1("hilla", Layout_1, [
        createRoute_1("", Page_4),
        createRoute_1("components", Page_5),
        createRoute_1("hello-react", Page_6)
    ]),
    createRoute_1("home", [
        createRoute_1("deep", [
            createRoute_1("hello", extendModule_1(Page_7, { "flowLayout": true }))
        ])
    ])
```

caalador:
I would expect that having
```
async function applyLayouts(routeMeta, layoutsFile) {
... 
 return routeMeta.map(
      (route) => layoutExists(layoutPaths, stripLeadingSlash(route.path)) ? enableFlowLayout(route) : route
    );
}
```
handle the child paths also here would make it right. So layout exist should be checked for route.path (only if not a layout) and then for each child as `route.path + child.path` and so on all the way to the bottom and any with `layoutExists==true` should get `enableFlowLayout` run on them.

This would reflect correctly to the json where specifically layout false would override with any user set value.

Lodin:
Thanks. I was too focused on the algorithm implementation and completely forgot to address `applyLayouts` issues. Let me do it now.

ZheSun88:
by the way.. should we add some tests to hilla project to catch this failure, besides platform?

Lodin:
Sure, I'm adding tests wherever possible. I don't want this situation to happen again...

Lodin:
We had a call with @caalador and figured out that there should be a couple of additional changed in the `applyLayouts` function that would allow the Flow server to understand what routes require Flow Layout. 

Basically, instead of doing one-level application using `routes.map`, we have to use `transformTree` function that will go deep and apply the layout logic to any nested route as well.

Additionally, we need a test that covers this case. 
I'm working on it right now.

caalador:
Has a timing changed as on initial build I always get
```
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker - 
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -  ERROR(TypeScript)  Cannot find module 'Frontend/generated/file-routes' or its corresponding type declarations.
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -  FILE  C:/Users/mgran/Code/Vaadin/platform/vaadin-platform-react-hybrid-test/src/main/frontend/generated/routes.tsx:37:24
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker - 
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -     35 | import { RouterConfigurationBuilder } from '@vaadin/hilla-file-router/runtime.js';
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -     36 | import Flow from 'Frontend/generated/flow/Flow';
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -   > 37 | import fileRoutes from 'Frontend/generated/file-routes';
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -        |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -     38 |
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -     39 | export const { router, routes } = new RouterConfigurationBuilder()
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker -     40 |     .withFileRoutes(fileRoutes) // (1)
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker - 
[dev-server-output] INFO com.vaadin.base.devserver.DevServerOutputTracker - [TypeScript] Found 1 error. Watching for file changes.
```
Which would point at the file being generated later than before.
Also the matcher seems to have been removed and replaced, but now it gives the wrong results.
For layouts.json `[{"path":"/home/deep"},{"path":"/flow"}]` and the paths
```
    createRoute_1("home", [
        createRoute_1("deep", extendModule_1(null, { "flowLayout": true }), [
            createRoute_1("hello", extendModule_1(Page_7, { "flowLayout": true }))
        ]),
        createRoute_1("deepend", extendModule_1(null, { "flowLayout": true }), [
            createRoute_1("deep", extendModule_1(Page_8, { "flowLayout": true }))
        ])
    ])
```
the path `/home/deepend/deep` also gets flowLayout true event though the parts `flow` and `deep`  do not in fact match `flow` and `deepend`. So now it is again in a startsWith and not a path parts match matcher mode.
This leads to the server side exception `com.vaadin.flow.router.NotFoundException: No layout for client path 'home/deepend/deep'` as the layout is for `deep` and not `deepend`

caalador:
At least one issue still. So now the flowLayout definitions seem to be set correctly and they take into account the path parts, but the paths
```
    createRoute_1("", Page_1),
    createRoute_1("flow", extendModule_1(null, { "flowLayout": true }), [
        createRoute_1("hello-hilla", extendModule_1(Page_2, { "flowLayout": true })),
        createRoute_1("no-layout", extendModule_1(Page_3, { "flowLayout": true }))
    ]),
    createRoute_1("hilla", Layout_1, [
        createRoute_1("", Page_4),
        createRoute_1("components", Page_5),
        createRoute_1("hello-react", Page_6)
    ]),
    createRoute_1("home", [
        createRoute_1("deep", extendModule_1(null, { "flowLayout": true }), [
            createRoute_1("hello", extendModule_1(Page_7, { "flowLayout": true }))
        ]),
        createRoute_1("deepend", [
            createRoute_1("deep", Page_8)
        ])
    ])
``` 
only register the paths:
```
ambivalent: Array(1)
0: {path: 'hilla', element: {…}, children: Array(3), handle: {…}}
length: 1
client: Array(2)
0: {element: {…}, handle: {…}, index: true, children: undefined}
1: {path: 'flow', element: undefined, children: Array(1), handle: {…}}
length: 2
[[Prototype]]: Array(0)
server: Array(2)
0: {path: 'flow', element: undefined, children: Array(1), handle: {…}}
1: {path: 'home', element: undefined, children: Array(1), handle: {…}}
length : 2
```
So the path `/home/deepend/deep` is not not a t all added to the client paths.

caalador (approved):

PR #2938, opened 2024-11-28
feat: add endpoint detection

tepi:
Add check for if endpoint generator needs to run

Fixes https://github.com/vaadin/flow/issues/20289
Fixes https://github.com/vaadin/flow/issues/18800


tepi:
Needs to be tested along with https://github.com/vaadin/flow/pull/20567 in order to compile

packages/java/endpoint/src/main/java/com/vaadin/hilla/startup/EndpointUsageDetectorImpl.java
```diff
@@ -0,0 +1,23 @@
...
+ * generator task needs to be run.
+ */
+public class EndpointUsageDetectorImpl implements EndpointUsageDetector {
+
+    @Override
+    public boolean areEndpointsUsed(Options options) {
+        ClassFinder classFinder = options.getClassFinder();
+        return !classFinder.getAnnotatedClasses(Endpoint.class).isEmpty()
+                || !classFinder.getAnnotatedClasses(BrowserCallable.class)
+                        .isEmpty();
```
mcollovati, line 20: This check should probably exclude Hilla internal annotated classes, like `SignalsHandler`

mshabarov: Excellent point, indeed, without this Flow would always try to generate the endpoints unless one completely excludes Hilla.
Something similar is needed:
```java
        ClassFinder classFinder = options.getClassFinder();
        return Stream.concat(
            classFinder.getAnnotatedClasses(Endpoint.class).stream(),
                classFinder.getAnnotatedClasses(BrowserCallable.class).stream())
            .anyMatch(annotatedClass -> !annotatedClass.getPackage().getName()
                .startsWith("com.vaadin.hilla"));
```

packages/java/hilla/src/test/java/com/vaadin/hilla/InternalBrowserCallableTest.java
```diff
@@ -0,0 +1,54 @@
...
+        Reflections reflections = new Reflections(BASE_PACKAGE,
+                Scanners.TypesAnnotated);
+        Set<Class<?>> annotatedClasses = reflections
+                .getTypesAnnotatedWith(InternalBrowserCallable.class);
+
+        for (Class<?> clazz : annotatedClasses) {
+            if (!clazz.isAnnotationPresent(BrowserCallable.class)
+                    && !clazz.isAnnotationPresent(Endpoint.class)) {
+                Assertions.fail("Class " + clazz.getName()
+                        + " is annotated with @BrowserCallable or @Endpoint, but missing @InternalBrowserCallable.");
```
mcollovati, line 50: I think the assertion message is not correct here: in this case, all `annotatedClasses` will have `@InternalBrowserCallable` annotation, but they might miss `@BrowserCallable` or `@Endpoint`

mcollovati: Probably the messages in the two tests are inverted

mshabarov: Yes, they are inverted indeed, thanks for noticing it 👍

mshabarov:
This should be ready for review now, trying locally and it seems to pass on existing tests and new ones.
Requires review and a suitable time to merge flow and hilla parts simultaneously.

cromoteca (dismissed):

packages/java/endpoint/src/main/java/com/vaadin/hilla/startup/EndpointUsageDetectorImpl.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Override
+    public boolean areEndpointsUsed(Options options) {
+        ClassFinder classFinder = options.getClassFinder();
+
+        return Stream.concat(
+                classFinder.getAnnotatedClasses(Endpoint.class).stream(),
+                classFinder.getAnnotatedClasses(BrowserCallable.class).stream())
+                .anyMatch(annotatedClass -> !annotatedClass
+                        .isAnnotationPresent(InternalBrowserCallable.class));
```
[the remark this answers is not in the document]
cromoteca: For the record, this is where we define the list of endpoints using Spring `ApplicationContext`: https://github.com/vaadin/hilla/blob/efc7da0d5e2e6fa0befab645134f3b80d5c6478d/packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointCodeGenerator.java#L114-L125

platosha (changes requested):
What should happen when the user removes last Hilla browser callable? I think, ideally, we should remove previously generated files then. And it's maybe better for such cleanup implementation to be in Hilla.

I'm thinking of taking a different approach here: Hilla “generator” are really two tasks:

- `TaskGenerateOpenAPI` is the Java parser, and is purely in Java. It's basically safe to leave it always on, since there aren't Node dependencies.
- `TaskGenerateEndpoint` is the TypeScript generator, which actually requires Node and npm dependencies). This one is the one that needs caution.

Instead of trying to skip both of them in Flow, we could keep both of them enabled, but leave it up to `TaskGenerateEndpoint` implementation to skip invoking Node when not needed. 

This way, introducing a separate detection logic, Hilla would find the endpoints using its regular approach. It would also make it possible to implement a proper cleanup in the `TaskGenerateEndpoint` task, removing any previously generated files listed in the `generated-file-list.txt`.

tepi:
@platosha Just to confirm I understood correctly: In Flow, in `com.vaadin.flow.server.frontend.NodeTasks#addEndpointServicesTasks` we should always add both Hilla tasks, and Hilla will then decide what needs to be executed? If so, I can rework the Flow PR to do this, and someone from Hilla team can change this PR / create a new PR that will do the detection using existing Hilla approach.

platosha:
> @platosha Just to confirm I understood correctly: In Flow, in `com.vaadin.flow.server.frontend.NodeTasks#addEndpointServicesTasks` we should always add both Hilla tasks, and Hilla will then decide what needs to be executed? If so, I can rework the Flow PR to do this, and someone from Hilla team can change this PR / create a new PR that will do the detection using existing Hilla approach.

Correct. Sounds good, let's do so.

tepi:
Updated https://github.com/vaadin/flow/pull/20567 to execute both Hilla tasks whenever Hilla is available on the class path.

packages/java/endpoint/src/main/resources/META-INF/services/com.vaadin.flow.server.frontend.EndpointUsageDetector
```diff
@@ -0,0 +1 @@
+com.vaadin.hilla.startup.EndpointUsageDetectorImpl
```
tepi, line 1: This file needs to be removed

platosha:
@tepi Hi! Could you, please, try if this change is sufficient for Flow's needs?

tepi:
> @tepi Hi! Could you, please, try if this change is sufficient for Flow's needs?

Yes, I will test it today. Thank you!

tepi:
Looks good to me!

platosha (dismissed):

tepi:
@platosha Could you please check if the Sonar failure needs to be addressed?

platosha:
> @platosha Could you please check if the Sonar failure needs to be addressed?

@tepi seems fine to me, it's coming from the default index HTML template in a new IT test project.

PR #2939, opened 2024-11-29
fix: log operation validator messages

### LINKED ISSUE #2916: [Full-Stack Signals] Operation validation error message is not propagated to the client
### Describe the bug

When an operation validator is defined for a full-stack signal, no matter what the rejection error is, on the client side, the reason is: _"server rejected the operation"_. 

### Expected-behavior

To be able to show a the validation error message that is defined for the server-side Signal instance's validator.

### Reproduction

Define an operation validator for signal instance that is returned to the client:
```java
private final ListSignal<Message> chatSignal = new ListSignal<>(Message.class).withOperationValidator(
            operation -> {
                if (operation instanceof ValueOperation<Message> valueOp) {
                    if (valueOp.value().text().toLowerCase().contains("bad")) {
                        return ValidationResult.reject("Bad words are not allowed");
                    }
                }
                return ValidationResult.allow();
            });
```
or even with a readonly signal:
```java
private final ListSignal<Message> guestSignal = chatSignal.asReadonly();
```
When user tries to insert a message to any of the above signals instances, and the operation validation gets rejected, the only available reason for in the `onrejected` callback is always: _"server rejected the operation"_:
```ts
chatSignal.insertLast({text: 'Hi', author:'John'})
               .result.then(() => {}, (reason) => Notification.show(reason));
```

### System Info

Vaadin: 24.6.0.alpha4
Hilla: 24.6.0.alpha5

taefi:
## Description

This logs the validation error messages on server side, and enhances the client-side error log for the developers to find the relevant log in server side easier.

Fixes #2916 

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/Signal.java
```diff
@@ -219,4 +224,8 @@ public int hashCode() {
     public static void setMapper(ObjectMapper mapper) {
         StateEvent.setMapper(mapper);
     }
+
+    private Logger getLogger() {
```
Lodin, line 228: Does it have any difference with the original `static` property implementation? 🤔

Lodin (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/Signal.java
taefi: It makes testing easier, while the static thing was very hard to spy on.

cromoteca (approved):

PR #2960, opened 2024-12-06
chore: Bump to 24.7-SNAPSHOT

Artur-:


packages/java/gradle-plugin/src/functionalTest/kotlin/com/vaadin/hilla/gradle/plugin/AbstractGradleTest.kt
```diff
@@ -25,7 +25,7 @@ import org.junit.Before
  */
 abstract class AbstractGradleTest {
 
-    val hillaVersion = System.getenv("hilla.version").takeUnless { it.isNullOrEmpty() } ?: "24.6-SNAPSHOT"
+    val hillaVersion = System.getenv("hilla.version").takeUnless { it.isNullOrEmpty() } ?: "24.7-SNAPSHOT"
```
taefi, line 28: Is there a reason that flow shouldn't be updated to `24.7-SNAPSHOT`?

packages/java/tests/gradle/single-module/gradle.properties
```diff
@@ -1,4 +1,4 @@
 org.gradle.daemon=false
-hillaVersion=24.6-SNAPSHOT
+hillaVersion=24.7-SNAPSHOT
```
taefi, line 2: Is there a reason that flow shouldn't be updated to `24.7-SNAPSHOT`?

taefi (commented):
Apart from the Flow version, I'd suggest to also update the npm versions to 24.7.0-alpha0 (similar to what we did for 24.6) so that it doesn't just work by pulling in the old versions unintentionally.

ZheSun88:
can you make the necessary changes @taefi

packages/java/gradle-plugin/src/functionalTest/kotlin/com/vaadin/hilla/gradle/plugin/AbstractGradleTest.kt
Artur-: No, but it is not related to this PR. It has some issues as you can see in #2961

packages/java/tests/gradle/single-module/gradle.properties
Artur-: No, but it is not related to this PR. It has some issues as you can see in #2961

Artur-:
Feel free to update the npm versions, in this PR or a separate one

taefi:
> can you make the necessary changes @taefi

I can, but it has to wait until I'm done with another build failure.

Artur-:
This does not break anything so why don't we merge this first, before some other PR cause conflicts with this?

Artur-:
If it is only about running `npm version 24.7.0-alpha0` in each folder, I can do that..

Artur-:
Updated, let's see what happens..

taefi (approved):

PR #2968, opened 2024-12-09
chore: bump remaining 24.6 versions to 24.7

taefi:
Upgrading `react-components` to `24.7` alpha needs a release from components, and will be done separately.


cromoteca (commented):
I'd avoid making those changes in `package.json` files, as they should come from the automatic PRs.

taefi:
> I'd avoid making those changes in `package.json` files, as they should come from the automatic PRs.

I've reverted the unnecessary components version upgrades, and only upgraded some weirdly not-upgraded deps such as `@vaadin/bundles` (some of the are upgraded, some are not). The rest, I'm updating them intentionally, to make sure, some tests are not passing just by pointing to old version accidentally.

PR #2986, opened 2024-12-10
fix: accept null values as valid vor positive/negative validators

### LINKED ISSUE #2826: @PositiveOrZero annotation should not make ts fields not-nullable
### Describe the bug

The javadoc of @PositiveOrZero says that "null elements are considered valid.". However, the client side validation will not accept null fields. 

image

### Expected-behavior

Null values are being accepted on the client-side for fields that are annotated with @PositiveOrZero

### Reproduction

Annotate a DTO-Field on the java side with @PositiveOrZero, then bind the client side representation of that field to an vaadin-integer-field in a form. Leave that field empty and call validate() on the binder. The empty value will not be accepted.

### System Info

Windows 11, Java 22, Chrome

cromoteca:
Addresses behavior of `Positive`, `Negative`, `PositiveOrZero`, and `NegativeOrZero`, which should accept null values as valid.

As API translates null to `NaN`, it should be accepted here.

Fixes #2826.

### DIFF
--- a/packages/ts/lit-form/src/Validators.ts
+++ b/packages/ts/lit-form/src/Validators.ts
@@ -268,7 +268,8 @@ export class Negative<T> extends AbstractValidator<T> {
   }
 
   override validate(value: T): boolean {
-    return toFloat(String(value)) < 0;
+    const fv = toFloat(String(value));
+    return isNaN(fv) || fv < 0;
   }
 
   readonly name = 'Negative';
@@ -280,7 +281,7 @@ export class NegativeOrZero<T> extends AbstractValidator<T> {
   }
 
   override validate(value: T): boolean {
-    return toFloat(String(value)) <= 0;
+    return (toFloat(String(value)) || 0) <= 0;
   }
 
   readonly name = 'NegativeOrZero';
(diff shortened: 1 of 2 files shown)


taefi (approved):
LGTM!

packages/ts/lit-form/src/Validators.ts
```diff
@@ -268,7 +268,8 @@ export class Negative<T> extends AbstractValidator<T> {
   }
 
   override validate(value: T): boolean {
-    return toFloat(String(value)) < 0;
+    const fv = toFloat(String(value));
+    return isNaN(fv) || fv < 0;
```
Lodin, line 272: Overall I'd approve these changes but wouldn't it be better to use null check before conversion?

```suggestion
    if (value != null) {
      return toFloat(String(value)) < 0
    }
    
    return true;
```

Lodin (changes requested):
LGTM but I have a question here

packages/ts/lit-form/src/Validators.ts
cromoteca: I wanted to tackle `null`, but also `undefined` and `''`.

Lodin: Ah, ok, so I missed the `''` one

Lodin (approved):

PR #2987, opened 2024-12-10
feat; add error message parameter to `onError` callback in subscription

### LINKED ISSUE #2062: Include the error in the onError callback of Subscription
### Describe your motivation

When subscribing to a flux connection, if you leave out the `onError` callback and there is an error, Hilla will throw the error and you can see what it is in the console. However, if you provide a callback and there is an error, you will only know that there is an error but you have no idea what it is.

Either Hilla should log the error regardless of whether there is a callback or not, or the error should be passed as a parameter to the callback.

_No response_

### Additional context

_No response_

cromoteca:
Closes #2062.

### DIFF
--- a/packages/ts/frontend/src/Connect.ts
+++ b/packages/ts/frontend/src/Connect.ts
@@ -42,7 +42,7 @@ export interface Subscription<T> {
   onComplete(callback: () => void): Subscription<T>;
 
   /** Called when an exception occured in the subscription. */
-  onError(callback: () => void): Subscription<T>;
+  onError(callback: (message: string) => void): Subscription<T>;
 
   /** Called when a new value is available. */
   onNext(callback: (value: T) => void): Subscription<T>;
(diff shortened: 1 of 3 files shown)


taefi (commented):
Generally, this is a nice enhancement, but, isn't it automatically expose some unwanted error messages to the client as well? For instance, when an endpoint method returning a Signal is annotated with `@RolesAllowed("ADMIN")`, so far, the only thing on client was: "Error in flux ...", and I wonder if this change exposes the the "access denied" error message to the client, and cause a security issue.

Lodin (approved):

taefi (changes requested):
Just adding this to block merging before the above question is answered.

cromoteca:
> Generally, this is a nice enhancement, but, isn't it automatically expose some unwanted error messages to the client as well? For instance, when an endpoint method returning a Signal is annotated with `@RolesAllowed("ADMIN")`, so far, the only thing on client was: "Error in flux ...", and I wonder if this change exposes the the "access denied" error message to the client, and cause a security issue.

The messages we're talking about here are already on the client. Without a callback, an error is thrown with the contents of the message. With a callback, the message was lost.

If the message contents are an issue on their own, I think that a separate issue should be opened.

PR #2998, opened 2024-12-12
fix: throw exception if feature flag is not enabled

### LINKED ISSUE #2937: [DX] When Full-stack signals feature flag is not set, it should throw an exception during the startup
### Describe the bug

When Full-stack signals feature flag is not set, there is no error in the server-side console/terminal, and a blank page is loaded on the browser. If the user opens the browser's console, there is a clear error about the feature flag missing, and how to set it. However, for Java full-stack developers, it is confusing, and normally, it is not their first choice to open the browser's console as the first step.

### Expected-behavior

When the Full-stack signals feature flag is not set, and the full-stack signals are used, it is convenient to prevent the application from getting any further during the startup, with a clear message that indicates what is missing and where it should be set.

### Background:
The Spring Beans needed for Full-stack Signals are not created unless the feature flag is set. This is controlled via annotating the beans with `@ConditionalOnFeatureFlag("fullstackSignals")`. So, in case an exception should be thrown, all browser-callable methods' return types should be scanned to check whether any full-stack signals are being returned without setting the feature flag or not.

Controlling the value of this feature flag during the instantiation of the full-stack signals is not easily possible, as they are not container managed beans.

### Reproduction

Without setting the feature-flag for the Full-stack Signals, create a browser-callable service that returns any type of full-stack signals, and run the application. 

### System Info

Version: Vaadin 24.6 beta1.

taefi:
## Description

This also removes the client-side check for
the feature flag to let the UI being rendered
normally, and server-side get a chance to
evaluate the feature-flag while handling the
requests for subscribe or update.

Fixes #2937

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/config/SignalsConfiguration.java
```diff
@@ -62,11 +63,12 @@ public SecureSignalsRegistry signalsRegistry() {
      *
      * @return SignalsHandler endpoint instance
      */
-    @ConditionalOnFeatureFlag("fullstackSignals")
+    // @ConditionalOnFeatureFlag("fullstackSignals")
```
cromoteca, line 66: Should be removed rather than commented out

cromoteca (changes requested):

packages/java/endpoint/src/main/java/com/vaadin/hilla/signals/config/SignalsConfiguration.java
taefi: Done.

cromoteca (approved):

PR #3001, opened 2024-12-13
chore: fix java doc

### LINKED ISSUE #3000: Outdated Javadoc on EndpointInvoker constructor
The Javadoc for `EndpointInvoker` constructor mentions `endpointObjectMapper` as a `optional factory bean to override the default ...`, but the argument is actually an `ObjectMapper`, not a factory.
Most likely a left-over from recent refactoring.

https://github.com/vaadin/hilla/blob/25417fcbc03b9b263db8b60c753b7fae6a6cc30f/packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointInvoker.java#L85-L90

taefi:
Fixes #3000

### DIFF
--- a/packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointInvoker.java
+++ b/packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointInvoker.java
@@ -84,7 +84,7 @@ public class EndpointInvoker {
      *            The Spring application context
      * @param endpointObjectMapper
      *            The object mapper to be used for serialization and
-     *            deserialization of state events. Use the
+     *            deserialization of request and response bodies. Use the
      *            {@link EndpointController#ENDPOINT_MAPPER_FACTORY_BEAN_QUALIFIER}
      *            qualifier to override the mapper.
      * @param explicitNullableTypeChecker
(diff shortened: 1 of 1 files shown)


packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointInvoker.java
```diff
@@ -83,9 +83,8 @@ public class EndpointInvoker {
      * @param applicationContext
      *            The Spring application context
      * @param endpointObjectMapper
-     *            optional factory bean to override the default
-     *            {@link JacksonObjectMapperFactory} that is used for
-     *            serializing and deserializing request and response bodies Use
+     *            The object mapper to be used for serialization and
+     *            deserialization of request and response bodies. Use the
      *            {@link EndpointController#ENDPOINT_MAPPER_FACTORY_BEAN_QUALIFIER}
      *            qualifier to override the mapper.
```
mcollovati, line 89: I would clarify a little bit the override hint, since the qualifier is not supposed to be put on mapper bean, but on a factory. Maybe something like


> To override the mapper, use the {@link EndpointController#ENDPOINT_MAPPER_FACTORY_BEAN_QUALIFIER} qualifier on a JacksonObjectMapperFactory bean definition.

mcollovati (approved):

PR #3035, opened 2024-12-17
chore: Bump sinon & chai

Artur-:


cromoteca (dismissed):
When updating versions of dependencies, it's probably better not to commit any `package-lock.json`: those changes will happen thanks to the automated PRs and conflicts will be less frequent.

Artur-:
I took some smaller steps here #3041, maybe we should merge that first as it seems to work

Artur-:
There seems to be one issue here: `filters correctly after changing model` freezes the browser when it does 
```
result.rerender(<AutoGrid service={_companyService} model={CompanyModel} />);
```

Artur-:
... but apparently it works on CI

cromoteca (approved):

PR #3047, opened 2024-12-18
fix: Make JpaFilterConverter work without an EntityManager

### LINKED ISSUE #2569: BLOCKER: Cannot upgrade to Vaadin 24.4 because of com.vaadin.hilla.crud.JpaFilterConverter
### Description of the bug

It is not so uncommon to have multiple datasources defined in spring boot application and by extension multiple entity managers which will cause the `com.vaadin.hilla.crud.JpaFilterConverter` to fail to autowire the entity manager since it cannot pick an entity manager factory it should use (usually specified by e.g. @PersistenceContext annotation's unitName attribute). You can partially get around this issue if you mark one of the factories as `@Primary` however that is not always possible, especially if the entity manager factory marked as primary is not the one that should be used by JpaFilterConverter. It should be possible to customize JpaFilterConverter bean where one can choose the correct entity manager to use. This is a blocker for upgrage to vaadin 24.4.

### Expected behavior

I need to be able specify which entity manager is used by `com.vaadin.hilla.crud.JpaFilterConverter`

1. Define multiple datasource beans in spring boot application (along with LocalContainerEntityManagerFactoryBean and PlatformTransactionManager for each datasource).
2. Application using vaadin 24.4.3 will fail to start because it will not be able to autowire he EntityManager in `com.vaadin.hilla.crud.JpaFilterConverter`

### Versions

- Vaadin / Flow version: 24.4.3
- Java version: 21
- OS version: MacOS 14.5
- IDE (if applicable): Intellij Idea Ultimate latest version

Artur-:
Fixes #2569


taefi (approved):
LGTM!

PR #3048, opened 2024-12-18
chore: Deprecate Hilla CRUD classes and recommend Flow ones

Artur-:


Artur-:
Depends on #3073  and relates to expose-flow-crud-methods

PR #3072, opened 2024-12-19
chore: Change the code so Eclipse JDT accepts it is ok

Artur-:
Currently, the project cannot be compiled in VS Code because it complains that 
```
Type mismatch: cannot convert from List<Class<capture#1-of ?>> to List<Class<?>>
```

This is not a real error but a problem in the Eclipse compiler.  his change makes the code compile also with VS Code / Eclipse


taefi (changes requested):
What is this trying to solve? Any description, or referred issue?

Artur-:
Updated description

taefi (approved):

PR #3073, opened 2024-12-20
chore: Expose methods from Flow crud classes to browser callables

Artur-:


Artur-:
This is related to https://github.com/vaadin/flow/pull/20743

packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/EndpointExposedPlugin.java
```diff
@@ -95,12 +95,20 @@ public NodeDependencies scan(@NonNull NodeDependencies nodeDependencies) {
...
                 .map(AnnotationInfoModel::getName)
-                .anyMatch(endpointExposedAnnotationName::equals);
+                .anyMatch(endpointExposedAnnotationName::equals)
+                || alwaysExpose(classInfo);
         var classInfoNode = exposed ? EndpointExposedNode.of(classInfo)
                 : EndpointNonExposedNode.of(classInfo);
         return classInfoNode;
     }
 
+    private boolean alwaysExpose(ClassInfoModel classInfo) {
```
cromoteca, line 105: Could this be treated in a way similar to https://github.com/vaadin/hilla/blob/f4a02726caa01ed7e318ff2d9e2621eaa6347550/packages/java/parser-jvm-core/src/main/java/com/vaadin/hilla/parser/core/Parser.java#L47-L48, i.e. define a list of internal exposed classes there?
I can do the change if you agree.

cromoteca (changes requested):

packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/EndpointExposedPlugin.java
Artur-: Sure, if that is a better place to define it

cromoteca: They seem to have something in common to me, but I'm not sure about the place, actually. Let's keep things as they are in your PR.

cromoteca (dismissed):

cromoteca (approved):

PR #3087, opened 2024-12-27
fix: restore endpoint registration after hotswap

cromoteca:
After #2616, endpoints were not registered correctly after a hotswap event.

This PR completes the code update to keep that working.


taefi (changes requested):
As discussed internally, this PR needs to take care of duplicate endpoint generation and registration that can happen e.g. in case of renaming an endpoint class name.

packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointController.java
```diff
@@ -124,27 +124,16 @@ public void registerEndpoints(URL openApiResource) {
         endpointBeans
                 .putAll(context.getBeansWithAnnotation(BrowserCallable.class));
 
-        if (endpointRegistry.isEmpty() && !endpointBeans.isEmpty()) {
-            LOGGER.debug("No endpoints found in openapi.json:"
-                    + " registering all endpoints found using the Spring context");
+        var currentEndpointNames = endpointBeans.values().stream()
+                .map(endpointRegistry::registerEndpoint)
+                .collect(Collectors.toSet());
```
taefi, line 129: As the Vaadin endpoint names are kept as all-lowercase, an extra mapping is needed, so that the `retainAll` doesn't remove all the endpoints:
```suggestion
        var currentEndpointNames = endpointBeans.values().stream()
                .map(endpointRegistry::registerEndpoint)
                .map(String::toLowerCase)
                .collect(Collectors.toSet());
```

taefi (changes requested):

taefi (approved):

PR #3094, opened 2024-12-30
chore: align dependencies version with Flow

mcollovati:
Hilla pins a couple of dependencies older than the ones defined by Flow. The clash is shown when Vaadin maven plugin fails for some reason and it dumps the project-plugin incompatibilities.

References vaadin/flow#20740


cromoteca:
@mcollovati is it necessary to pin those or we could count on a BOM?

mcollovati:
> @mcollovati is it necessary to pin those or we could count on a BOM?

They are not in Flow BOM. In Flow they are defined in parent pom dependency management and included in `flow-server` and some other module.

cromoteca (approved):

mcollovati:
This one should probably be picked to 24.6

PR #3097, opened 2024-12-30
feat: allow classes with one of `JsonValue` and `JsonCreator` to be generated

### LINKED ISSUE #3095: Hilla should tolerate classes using only one of `@JsonValue` and `@JsonCreator`
### Describe your motivation

There are use cases for one of those annotations being used without the other one. See for example [How to force Hilla not to consider java class with @JsonCreator as single-value?](https://vaadin.com/forum/t/how-to-force-hilla-not-to-consider-java-class-with-jsoncreator-as-single-value/168023) 

Such classes are not interpreted as domain objects and are treated as usual.

_No response_

### Additional context

_No response_

cromoteca:
This PR simply skips treating those classes as special. A debug log is emitted. Consequences are possible if that makes serialization and deserialization types different, as Hilla needs those to be the same.

Closes #3095.


Legioth:
A use case for `@JsonCreator` without `@JsonValue` was described in the forum discussion linked from the ticket. But is there any use case for `@JsonValue` without `@JsonCreator` or should that combination still be disallowed?

cromoteca:
While testing some possible use cases, I realized that `@JsonCreator` is actually optional, at least in some cases, as Jackson is smart enough to deserialize correctly a single value passed as constructor parameter for example. Here's the simplest case.

```java
public class Email {
    private final String email;

    public Email(String email) {
        this.email = email;
    }

    @JsonValue
    public String getEmail() {
        return email;
    }

    @Override
    public String toString() {
        return email;
    }
}
```

So we should probably just look at `@JsonValue`, as verifying that a corresponding creator (with annotation or not) exists is not easy.

cromoteca:
This version handles both serialization formats correctly (just comment out `@JsonValue` to verify):
```java
public class Email {
    private String email;

    public Email() {}

    public Email(String email) {
        this.email = email;
    }

    @JsonValue
    public String getEmail() {
        return email;
    }

    public void setEmail(String email) {
        this.email = email;
    }

    @Override
    public String toString() {
        return email;
    }
}
```

Legioth:
Sounds like it would make sense to only look for `@JsonValue` since that's the one with an actual semantic meaning but ignoring `@JsonCreator` which has no semantic meaning but is used as a hint in multiple different cases.

cromoteca:
Exactly. So I propose to:
- file a new bug
- remove `@JsonCreator` support completely from Hilla (does not break existing apps)
- cherry-pick to 24.6
- update docs

cromoteca:
I proposed to cherry7pick since the user who reported on the forum cannot upgrade from 24.5 to 24.6 due to this.
I can't see which case could lead to incompatibility: a class which only uses one is currently unsupported. And for those who use both, we get the type from the return type of the method marked with `@JsonValue`.

taefi (approved):
LGTM. I think we can just merge this in main, and do a smoke test in playground projects that were using the domain primitives, for example. If there were no discrepancies, we can back port it to 24.6 confidently.

cromoteca:
> LGTM. I think we can just merge this in main, and do a smoke test in playground projects that were using the domain primitives, for example. If there were no discrepancies, we can back port it to 24.6 confidently.

I agree with you, let's make sure that this approach works in our existing projects.

PR #3109, opened 2025-01-02
chore: Bump preact signals

Artur-:
Major version 3 drops support for auto tracking, not in use in Hilla anymore

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   },
   "devDependencies": {
     "@nx/js": "^20.2.2",
-    "@preact/signals-react-transform": "^0.5.0",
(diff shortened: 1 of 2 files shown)


cromoteca:
Are you sure that auto tracking is no longer needed? I remember that some discussion was still ongoing. @Legioth?

Artur-:
Yes, it does not work properly and it does not support React 19

Artur-:
If it was needed, presumably this build would have failed

cromoteca (approved):
It should no longer be needed indeed. It's easy to revert anyway in case of problems.

Legioth:
It was still used by accident a while ago but I assume that has been fixed.

PR #3156, opened 2025-01-14
feat: add kotlin compiler args to gradle plugin

### LINKED ISSUE #2343: Form validation does not work in Kotlin.
### Describe the bug

When I use Kotlin class for endpoint, validation does not work.
image

### Expected-behavior

show errors in form

### Reproduction

# Repository 
https://github.com/inithink/hilla-crm-tutorial.git

# Kotlin
## KotlinClass.kt
```kotlin
import jakarta.validation.constraints.Email
import jakarta.validation.constraints.NotNull

data class KotlinClass(var email: @Email @NotNull String)

```

## KotlinClassForm.tsx
```typescript
import {TextField} from "@hilla/react-components/TextField";
import {Button} from "@hilla/react-components/Button";
import {useForm} from "@hilla/react-form";
import KotlinClassModel from "Frontend/generated/com/example/application/services/KotlinClassModel";

export default function KotlinClassForm() {
  const {field, model, submit, reset} = useForm(KotlinClassModel,);

  return (
    <div className="flex flex-col gap-s items-start">
      <TextField label="Email" {...field(model.email)} />
      <div className="flex gap-m">
        <Button onClick={submit} theme="primary">Save</Button>
        <Button onClick={reset}>Reset</Button>
      </div>
    </div>
  )
}
```

# Java
## JavaClass.java
```java
package com.example.application.services;

import jakarta.validation.constraints.Email;
import jakarta.validation.constraints.NotNull;

public class JavaClass {
  public JavaClass(String email) {
    this.email = email;
  }

  @NotNull
  @Email
  String email;

  public String getEmail() {
    return email;
  }

  public void setEmail(String email) {
    this.email = email;
  }
}
```

## JavaForm.tsx 
```typescript
import {TextField} from "@hilla/react-components/TextField";
import {EmailField} from "@hilla/react-components/EmailField";
import {Select, SelectItem} from "@hilla/react-components/Select";
import {Button} from "@hilla/react-components/Button";
import {useForm} from "@hilla/react-form";
import ContactRecordModel from "Frontend/generated/com/example/application/services/CRMService/ContactRecordModel";
import {CRMService, JavaService} from "Frontend/generated/endpoints";
import {useEffect, useState} from "react";
import ContactRecord from "Frontend/generated/com/example/application/services/CRMService/ContactRecord";
import {StringModel} from "@hilla/form";
import JavaClassModel from "Frontend/generated/com/example/application/services/JavaClassModel";
import JavaClass from "Frontend/generated/com/example/application/services/JavaClass";

export default function JavaClassForm() {
  const {field, model, submit, reset} = useForm(JavaClassModel);

  return (
    <div className="flex flex-col gap-s items-start">
      <TextField label="Email" {...field(model.email)} />
      <div className="flex gap-m">
        <Button onClick={submit} theme="primary">Save</Button>
        <Button onClick={reset}>Reset</Button>
      </div>
    </div>
  )
}
```

### System Info

macOS 14.4.1(23E224), M1 Pro
hilla version: 2.5.6
browser: chrome 123.0.6312.124

taefi:
Fixes #2343 

For validation results check here: https://github.com/vaadin/hilla/actions/runs/13456971893

## Description

Vaadin-Hilla apps can be written in Kotlin with either Gradle or Maven. This PR is only addressing the Gradle side (since it is 
the default in SpringBoot projects downloaded from spring initializr. The same can be done for the Maven but in a different PR, since the approach is totally different. 

The approach of this PR was to use reflection to dynamically add the compiler arges to the `compileKotlin` task, so that users don't have to do it manually in their projects. The reflection approach has the following benefits:

* It only needs the `kotlin-reflect` dependency to be available in the user projects, and since that dependency is needed in almost all the Kotlin SpringBoot projects because of the limitations of Kotlin closed classes e.g. in spring-data-jpa, no extra dependency is imposed to user projects.

This could've been done in another way, without reflection, as follows:
`HillaPlugin.kt`
``` kotlin
project.plugins.withId("org.jetbrains.kotlin.jvm") {
      project.tasks.withType(KotlinCompile::class.java) { kotlinCompile ->
          // Configure Kotlin compiler options
          kotlinCompile.compilerOptions { kotlin ->
              kotlin.freeCompilerArgs.addAll(
                  listOf(
                      "-Xjsr305=strict",
                      "-Xemit-jvm-type-annotations"
                  )
             )
        }
    }
}
```
But, this needs `KotlinCompile` class to be available in the buildScript classpath (the classpath for the plugins and the processing of the build file, not the user application code), user projects need to add a `buildScript` block in either their `build.gradle` or `settings.gradle` to add the following dependency for our plugin:

`settings.gradle`
```groovy
buildscript {
    repositories {
        mavenCentral()
    }
    dependencies {
        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.9.25")
    }
}
```
This approach has a benefit that is more robust when upgrading gradle version, and possible errors in case of deprecation/removal of the used API would be easily noticed, while the reflection approach would fail at runtime (which can be covered by tests instead). 

Part of #2343


packages/java/gradle-plugin/src/main/kotlin/com/vaadin/hilla/gradle/plugin/HillaPlugin.kt
```diff
@@ -56,6 +57,16 @@ public class HillaPlugin : Plugin<Project> {
             }
         }
 
+        // Configure Kotlin-specific tasks only if Kotlin JVM plugin is applied
+        project.plugins.withId("org.jetbrains.kotlin.jvm") {
+            project.tasks.named("compileKotlin").configure { task ->
+                val compilerOptions = task.javaClass.getMethod("getCompilerOptions").invoke(task)
+                val freeCompilerArgs = compilerOptions.javaClass.getMethod("getFreeCompilerArgs")
```
cromoteca, line 64: I wonder if one of those `invoke`s can return `null`

cromoteca (changes requested):
It's just a comment, but I'm requesting changes to make it visible.

packages/java/gradle-plugin/src/main/kotlin/com/vaadin/hilla/gradle/plugin/HillaPlugin.kt
taefi: Technically, when the Kotlin jvm plugin is applied, the compileKotlin task, and its options cannot be null. But, added null check guards for possible edge cases.

cromoteca (approved):

PR #3157, opened 2025-01-14
fix(binder): convert property paths in arrays to dot notation

### LINKED ISSUE #3099: Server side validators not working correctly with array form binding
### Describe the bug

The error messages are not displayed on the field for server side validators when using the array data binding. 

The parameterName is generated with bracket notation from the server side and it looks like this:

`"parameterName":"pets[0].description"`

The frontend code expects the parameterName to look like this:

`"parameterName":"pets.0.description"`

### Expected-behavior

The server side validation message should be shown.

### Reproduction

I modified the pet clinic project to demonstrate the issue available here: https://github.com/aspan/spring-petclinic-hilla/tree/develop

The Owner edit form has been modified to add a server side validator to the description of a pet in the list of pets requiring between 3 and 100 characters.

I added a small servlet filter to replace the server side response body with the correct parameterName in `org.springframework.samples.petclinic.FixValidationResponseWithBrackets`

The project currently works because of the servlet filter but as soon as it is removed the validation error message is not shown any more.

### System Info

Hilla 24.6.0

cromoteca:
As the square bracket array syntax comes directly from Java valdiation, I preferred not to touch at the generated exception messages.

Instead, the binder now converts the notation on the fly when parsing the error message, so that it will work in the form.

Fixes #3099

### DIFF
--- a/packages/ts/lit-form/test/Validation.test.ts
+++ b/packages/ts/lit-form/test/Validation.test.ts
@@ -738,6 +738,32 @@ describe('@vaadin/hilla-lit-form', () => {
         }
       });
 
+      it('should display server validation error for product description', async () => {
+        binder.for(binder.model.customer.fullName).value = 'foobar';
+        binder.for(binder.model.notes).value = 'whatever';
+        await fireEvent(orderView.add, 'click');
+        const productModel = [...binder.model.products][0].model;
+        binder.for(productModel.description).value = 'foobar';
+        binder.for(productModel.price).value = 10;
+        const requestUpdateSpy = sinon.spy(orderView, 'requestUpdate');
+        try {
+          await binder.submitTo(() => {
+            requestUpdateSpy.resetHistory();
+            throw new EndpointValidationError('Validation error in endpoint "MyEndpoint" method "saveMyBean"', [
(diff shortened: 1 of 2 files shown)


platosha (approved):

packages/ts/lit-form/test/Validation.test.ts
```diff
@@ -754,6 +754,32 @@ describe('@vaadin/hilla-lit-form', () => {
         }
       });
 
+      it('should display server validation error for product description', async () => {
```
Lodin, line 757: Where in the test the changes made are tested? 🤔 Maybe, it would be better to add a check for transformed message?

Lodin (changes requested):

packages/ts/lit-form/test/Validation.test.ts
cromoteca: Line 769 contains `'products[0].description'` in square bracket notation and the test verifies that the associated error is propagated to the binder. And I verified that the test fails without the code change.

Lodin: Ah, ok, I see. Thanks for clarification

Lodin (approved):

PR #3159, opened 2025-01-15
(PR on top of #3048): add replacements in the `react-crud` package

cromoteca:


Artur-:
Looks good, if we rename the `.d.ts` files, will something stop working? It should be okay to still use the old interfaces, so should we make copies instead?

cromoteca:
If I recall, those are for internal compilation. User applications have all CRUD-related types generated by the generator.

PR #3167, opened 2025-01-20
feat: remove `EndpointsValidator` class

cromoteca:
This is an alternative to #2666.

`EndpointsValidator` checks if classes annotated with `@Endpoint` or `@BrowserCallable` are in the classpath and, in that case, throws an exception if Spring is not available.

In 24.7 we ask Spring for the list of endpoints, so by definition if there's no Spring, there are no endpoints. 

See also the discussion in #2938.


taefi:
Prior to the unified platform, the existence of `EndpointsValidator` was unnecessarily a strong check, since most of the Hilla projects would've been initiated either from start.vaadin.com, Spring Initializr, or the skeleton starters we provide. So the chances of users being confused about using endpoints in a non-spring project was very very low.

Now that we have a unified platform, users can start a Vaadin project without Spring, and later get interested in Hilla, and they can start following the docs at a wrong place that just instructs how to create endpoints. In that case, a warning in the console could prevent confusions, but, failing the application startup still feels unnecessary.

cromoteca:
If you create an endpoint in your application, you have `@BrowserCallable` in your classpath, which means that you have Hilla. Is there a possibility to create a non-Spring application that contains Hilla dependencies? I don't think that we provide such a thing, but maybe there's another path that users could follow and could need to be warned.

taefi:
The only scenario I can think of is a Vaadin Flow app that was based on some other containers, and when they've upgraded to 24.4+, they excluded Hilla and possibly any trace of SpringBoot from the dependencies. When adding `@BrowserCallable` on some class, it only needs the removal of the Hilla exclusion, but if they add exclusions for SpringBoot from Hilla, then confusion can happen. Feels like an edge case, and we can consider removing that validator altogether.

taefi (approved):

PR #3168, opened 2025-01-20
feat(generator): throw on overloaded methods

cromoteca:


cromoteca:
Closing as it does not seem to be the preferred solution

PR #3183, opened 2025-01-26
chore: redesign dependency system

Lodin:
This PR redesigns the dependency system in Hilla in order to avoid mismatched dependencies. According to Nx recommendation, all dev dependencies now are listed only in the root `package.json` to simplify update & managing.


packages/ts/react-crud/src/crud.d.ts
```diff
@@ -3,14 +3,15 @@ import type Pageable from './types/com/vaadin/hilla/mappedtypes/Pageable';
...
+
   delete(id: any): Promise<void>;
 }
 
 export interface ListService<T> {
   list(request: Pageable, filter: FilterUnion | undefined): Promise<T[]>;
 }
 
-export interface CountService<T> {
+export interface CountService {
```
Lodin, line 14: This could be a significant change; originally I changed it because Eslint shown an error here (`T` is not used anywhere). I would like to have someone's opinion here.

packages/java/tests/spring/react-signals/frontend/views/SharedNumberSignal.tsx
```diff
@@ -11,12 +11,16 @@ export default function SharedNumberSignal() {
 
   return (
     <div>
-      <span id="sharedValue">{sharedValue}</span>
+      <span id="sharedValue">
+        <>{sharedValue}</>
```
Lodin, line 15: This change is interesting. I got the idea of it right from the `@preact/signals-react` package [README](https://github.com/preactjs/signals/tree/main/packages/react#rendering-optimizations). With the `Fragment` used here the use of a signal without a `value` is permitted

karma.config.cjs
```diff
@@ -156,6 +156,11 @@ module.exports = (config) => {
         build: {
           target: 'esnext',
         },
+        server: {
+          warmup: {
+            clientFiles: ['src/**/*', 'test/**/*'],
```
Lodin, line 161: Adding the warmup fixes the issue with `useRef` hook that happens because of migration from `vite v5` to `Vite v6`

.npmrc
```diff
@@ -1 +1,3 @@
 git-tag-version=false
+prefer-dedupe=true
+save-exact=true
```
Lodin, line 3: With the new dependency system, we use the pinned dependencies, so we can easily remove and re-create `package-lock.json` without pulling dependency inconsistence. To update dependencies, we can use [`npm-check-updates`](https://www.npmjs.com/package/npm-check-updates) package

Lodin (commented):
Here are the moments I'd like to highlight in this giant PR. Other changes are mostly about Eslint complaints or changes to support the new dependency system.

Lodin:
Please ignore the Codecov failure; it has nothing to do with this PR

packages/ts/react-crud/src/crud.d.ts
cromoteca: Could we change it to `T = unknown` or `T = never`?

Lodin: Well... That won't fix the Eslint issue because `T` is still not used anywhere in the body of the interface.

packages/java/tests/gradle/single-module-tests/package.json
```diff
@@ -18,17 +18,18 @@
...
     "classnames": "^2.3.2",
     "construct-style-sheets-polyfill": "3.1.0",
-    "lit": "3.1.2",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
-    "react-router-dom": "6.22.2"
+    "lit": "3.2.1",
+    "react": "18.3.1",
+    "react-dom": "18.3.1",
+    "react-router-dom": "6.22.2",
```
taefi, line 24: Just a note: this version of the `react-router-dom` is outdated (same in the other Gradle IT project), with `"6.28.2"` being the latest. I have no good ideas why isn't it broken in the builds considering the `react` and `react-dom` are upgraded to the `18.3.1`. 
This version conflict was the reason of snapshot builds failure for 24.5 and 24.6. The curious part is that it was only happening in the `react-signals` ITs, and not for example in `react-grid-test` ITs (with an outdated version of `react-router-dom`).

taefi (changes requested):
Added a comment (requesting changes to prevent merging).

taefi (approved):
LGTM!

PR #3189, opened 2025-01-30
feat(connect): add support for File type in requests

### LINKED ISSUE #3132: Client support for file upload with Hilla endpoints
Extracted from #3130

When given at least one parameter of type `File`, the endpoint client should use `multipart/form-data` request type and include all parameters with their name, including all files.

Unit-testing is enough for the client part.

Example usage:

```tsx
export default function UploadForm({}) {
  const [uploadState, setUploadState] = useState("");

  const uploadFile = useCallback(async (event: SubmitEvent) => {
    event.preventDefault();
    const file = new FormData(event.target).get("file");
    await AppEndpoint.saveFile(file);
    setUploadState("ok!");
  });

  return <form onSubmit={uploadFile}>
    <input type="file" name="file" />
    <button type="submit">Upload</button>
    <output>{uploadState}</output>
  </form>;
}
```

cromoteca:
Switches to multipart requests if a `File` object is found anywhere in the body object.

Closes #3132


packages/ts/frontend/src/Connect.ts
```diff
@@ -24,6 +24,8 @@ $wnd.Vaadin.registrations.push({
   is: 'endpoint',
 });
 
+export const bodyPartName = 'hilla_body_part';
```
Lodin, line 27: I know, we are not good in following this rule but let's try to keep exported const in the uppercase.
```suggestion
export const BODY_PART_NAME = 'hilla_body_part';
```

packages/ts/frontend/src/Connect.ts
```diff
@@ -304,13 +340,31 @@ export class ConnectClient {
...
+        JSON.stringify(paramsWithoutFiles, (_, value) => (value === undefined ? null : value)),
+      );
+
+      for (const [path, file] of files) {
+        body.append(path, file);
+      }
+    } else {
+      headers['Content-Type'] = 'application/json';
+      body =
+        params !== undefined ? JSON.stringify(params, (_, value) => (value === undefined ? null : value)) : undefined;
```
Lodin, line 363: We can make it a bit more readable.

```suggestion
        if (params) {
          body = JSON.stringify(params, (_, value) => (value === undefined ? null : value))
        }
```

Lodin (changes requested):

packages/ts/frontend/test/Connect.test.ts
Lodin, line 1: Also, to make SonarCloud happy, you can wrap repeating test in the function and use it

Lodin (approved):

PR #3197, opened 2025-02-03
chore: update frontend dependency to fix snapshot failure

ZheSun88:


taefi (changes requested):
I suggest we get rid of `react-router-dom` dep in the ITs, and replace all the imports to be from `react-router`. I'll experiment this in a separate PR.

taefi:
> I suggest we get rid of `react-router-dom` dep in the ITs, and replace all the imports to be from `react-router`. I'll experiment this in a separate PR.

Unfortunately, this needs upgrading the `react-rotuer` to `7.+` (as done in `main`), so not a good idea for 24.6 patch releases.

Lodin:
@ZheSun88, I believe, the issue is fixed by #3199

PR #3204, opened 2025-02-04
test: add fallback for Playwright-provided browser

Lodin:
If you want to get rid of an additional Chrome binary in your `node_modules`, you can add the following environment variables:
```
PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
CHROME_BIN=/opt/google/chrome/chrome # or any other path to your Chrome binary
```
**Note**: if you have `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD` set without `CHROME_BIN`, it will throw an error.
**Note**: for this change to take effect you have to run `npm ci`.

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -165,6 +165,7 @@ jobs:
 
     steps:
       - name: Install Chrome
+        id: setup-chrome
         uses: browser-actions/setup-chrome@v1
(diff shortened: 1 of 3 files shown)


Lodin:
This PR changes the workflow, so the correct Validation is [here](https://github.com/vaadin/hilla/actions/runs/13158821464)

PR #3214, opened 2025-02-07
refactor: use OpenAPI file path for generator instead of stdin & modernize generator-cli

Lodin:
This pull request includes significant updates to both the Java and TypeScript codebases, focusing on improving the `GeneratorProcessor` and `GeneratorIO` functionalities. The most important changes include the use of direct read of an OpenAPI file instead of copying it to stdin, refactoring the TypeScript generator CLI to use URLs, and updating the test suite to reflect these changes.

### Java Changes:
* **GeneratorProcessor Enhancements:**
  - Used a OpenAPI file name path as an CLI argument instead of copying it to stdin. (`packages/java/engine-core/src/main/java/com/vaadin/hilla/engine/GeneratorProcessor.java`)

### TypeScript Changes:
* **GeneratorIO Refactoring:**
  - Updated `GeneratorIO` to use `URL` instead of string paths for file operations. (`packages/ts/generator-cli/src/GeneratorIO.ts`) [[1]](diffhunk://#diff-d8f1b95a6ecda0b903d1a0aa0e2bec36e6321619870014904ab09982c4e94372L1-R31) [[2]](diffhunk://#diff-d8f1b95a6ecda0b903d1a0aa0e2bec36e6321619870014904ab09982c4e94372L135-R126)

* **CLI Improvements:**
  - Refactored the CLI to use `parseArgs` for argument parsing and removed `meow`. (`packages/ts/generator-cli/src/index.ts`)
  - Removed the `processInput` utility function and updated the CLI to accept the input file path to read it directly. (`packages/ts/generator-cli/src/index.ts`)

* **Testing Enhancements:**
  - Updated tests to reflect changes in `GeneratorIO` and added new tests for the CLI. (`packages/ts/generator-cli/test/GeneratorIO.spec.ts`, `packages/ts/generator-cli/test/cli.spec.ts`) [[1]](diffhunk://#diff-d039c1bd06202d532f5bed32fe687dbb0c1a4b4ccfd0aba16a676eaa1bbb2616L1-R91) [[2]](diffhunk://#diff-f8e38da44093fe7be6621e12808d7b55db5a1a4f0bb0ec72619fe387aabc4526R1-R24)
  - Added a new test plugin for better test coverage. (`packages/ts/generator-cli/test/TestPlugin.ts`)


packages/java/engine-core/src/main/java/com/vaadin/hilla/engine/GeneratorProcessor.java
```diff
@@ -7,6 +7,7 @@
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Function;
```
cromoteca, line 10: Nit: seems unused

packages/ts/generator-cli/test/fixtures/version.snap
```diff
@@ -0,0 +1 @@
+24.7.0-alpha11
```
cromoteca, line 1: Is this something we will have to update manually?

cromoteca: If so, would it be better to remove this and check that "should print version" just outputs something that seems a version number? Like, it contains a number and a dot.

cromoteca (changes requested):
LGTM, just a little doubt about `packages/ts/generator-cli/test/fixtures/version.snap`

packages/ts/generator-cli/test/fixtures/version.snap
Lodin: Ah, sorry, I already faced this issue and just didn't remove the file. Currently, it reads `package.json` and expects the correct version there (yeah, a bit circly, but I guess it's ok for testing a code).

packages/ts/generator-cli/src/index.ts
```diff
@@ -1,59 +1,83 @@
...
-  `
-Usage:
-  tsgen
-    (will read JSON from stdin)
-  tsgen <OpenAPI JSON string>
-  tsgen <OpenAPI file path>
+  values: { inputFile, help, outputDir, plugin: plugins, verbose, version },
+} = parseArgs({
+  options: {
+    inputFile: {
```
platosha, line 14: Nit: should this be a positional argument now that it is required? I find it a bit strange to have a non-optional ”option”...

platosha (changes requested):

packages/ts/generator-cli/src/index.ts
Lodin: Hmm, well, sounds reasonable. However, instead of making it positional, what if I make it optional, and calling a bare generator would result in calling `--help`?

platosha: Would still be strange that non-option is an option. Since input file is mandatory, it does not need a specific name, I'd expect just `cmd file.json`. And also it should be not listed among other optional arguments (`--verbose` and such) in the `--help` contents, for instance.

We could only support single positional argument — always index 0 — appears to work just as expected with `parseAgrs()` even when there are options specified before that.

platosha (approved):

PR #3220, opened 2025-02-10
feat(parser): add support for Spring nullability annotations

### LINKED ISSUE #3205: `org.springframework.lang.Nullable` not work when generate endpoint return dto
### Describe the bug

In hilla, we can use `org.springframework.lang.NonNullApi` for nullability.

Here is the [doc](https://vaadin.com/docs/latest/hilla/reference/type-nullability#nonnullapi).

```java
@NonNullApi
package com.example.application;
```

```java
public class MyBean {
    public List<String> list;
    public Map<String, Integer> map;
    @Nullable
    public String nullable;
}
```

```java
export default interface MyBean {
  list: Array<string>;
  map: Record<string, number>;
  nullable?: string;
}
```

If `jakarta.annotation.Nullable` is used, it works.

But when `org.springframework.lang.Nullable` is used, It do not work, like below
```java
export default interface MyBean {
  list: Array<string>;
  map: Record<string, number>;
  nullable: string;
}
```

### Expected-behavior

When `org.springframework.lang.Nullable` is used, it should be 
```java
export default interface MyBean {
  list: Array<string>;
  map: Record<string, number>;
  nullable?: string;
}
```

### Reproduction

Just use `org.springframework.lang.Nullable` in dto, it will be reproducted.

### System Info

Linux mint 22.1
Hilla 24.5.9
Edge 132.0.2957.140

cromoteca:
Adds support for Spring nullability annotations.

Also removes from the list those that cannot be detected as not available at runtime anyway.

Closes #3205

### DIFF
--- a/packages/java/parser-jvm-plugin-nonnull/src/main/java/com/vaadin/hilla/parser/plugins/nonnull/NonnullPluginConfig.java
+++ b/packages/java/parser-jvm-plugin-nonnull/src/main/java/com/vaadin/hilla/parser/plugins/nonnull/NonnullPluginConfig.java
@@ -56,25 +56,18 @@ public class NonnullPluginConfig
                 // only matter when they are used in conjunction with
                 // package-level annotations
                 new AnnotationMatcher("jakarta.annotation.Nullable", true, 20),
-                new AnnotationMatcher("org.jetbrains.annotations.Nullable",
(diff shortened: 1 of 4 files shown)


taefi (approved):
LGTM.

PR #3223, opened 2025-02-12
chore(crud): expose Hilla classes

cromoteca:
As moving those classes to Flow has been reverted, we need to expose Hilla classes here.

### DIFF
--- a/packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/EndpointExposedPlugin.java
+++ b/packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/EndpointExposedPlugin.java
@@ -103,10 +103,8 @@ public final class EndpointExposedPlugin
     }
 
     private boolean alwaysExpose(ClassInfoModel classInfo) {
-        return classInfo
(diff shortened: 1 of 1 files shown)


packages/java/parser-jvm-plugin-backbone/src/main/java/com/vaadin/hilla/parser/plugins/backbone/EndpointExposedPlugin.java
```diff
@@ -103,10 +103,8 @@ public NodeDependencies scan(@NonNull NodeDependencies nodeDependencies) {
     }
 
     private boolean alwaysExpose(ClassInfoModel classInfo) {
-        return classInfo
-                .is("com.vaadin.flow.spring.data.jpa.CrudRepositoryService")
-                || classInfo.is(
-                        "com.vaadin.flow.spring.data.jpa.ListRepositoryService");
+        return classInfo.is("com.vaadin.hilla.crud.CrudRepositoryService")
```
Artur-, line 106: Why? Aren't the Hilla classes normal `@EndpointExposed`?

cromoteca: Correct. I went too fast there.

PR #3224, opened 2025-02-12
feat(crud): support configured max page size

### LINKED ISSUE #3203: AutoCrud: add support for limiting the pageSize without deriving/overriding the Hilla services
### Describe your motivation

The idea comes from [a forum question](https://vaadin.com/forum/t/limiting-the-page-size-on-the-backend/168259), asking for the possibility of setting `maxPageSize` limit for handling cases such as forged requests for either no page size or a very big one.

Support for this could be added by giving control to the users for setting `maxPageSize` either through already existing properties (being compatible with spring-data-jpa standards makes it easier to integrate Hilla in already existing projects): 
```
spring:
  data:
    web:
      pageable:
        max-page-size: 10
    rest:
      max-page-size: 10
``` 
or by enabling the users to define a default page bean which can be taken into account by the framework, if present. 

It would be desirable to support the already existing / standard ways to support this in Vaadin, unless some specific need prevents us from using them.

Creating a couple of CRUD super classes in the user workspace that set the limits by overriding the defaults in `CrudRepositoryService` and etc. when defining a BrowserCallable service.

### Additional context

_No response_

cromoteca:
When using `ListRepositoryService` and its subclasses, one can pass an arbitrary page size, potentially querying many lines. This is also true in Hilla, where the client side decides.

This PR uses the existing `spring.data.rest.pageable.max-page-size` property, if defined, and silently caps the maximum page size.

Closes #3203.


packages/java/endpoint/src/test/java/com/vaadin/hilla/crud/CrudRepositoryServiceJpaTest.java
```diff
@@ -93,4 +98,20 @@ public void deleteAll() {
...
+                        .getEnvironment());
+        // force init again, as it already happened
+        testCrudRepositoryService.init();
+
+        List<TestObject> result = testCrudRepositoryService
+                .list(Pageable.ofSize(5), null);
+
+        Assert.assertEquals(2, result.size());
+        Assert.assertEquals("John", result.get(0).getName());
+        Assert.assertEquals("Jeff", result.get(1).getName());
```
taefi, line 115: Isn't there a chance that the repository return the object in a different order and make this test flaky? I suggest removing these two asserts, as asserting the size is enough for the purpose of this testing this functionality.

packages/java/endpoint/src/main/java/com/vaadin/hilla/crud/ListRepositoryService.java
```diff
@@ -52,10 +57,13 @@ public ListRepositoryService(R repository) {
     }
 
     @PostConstruct
-    private void init() {
+    void init() {
         if (repository == null) {
             repository = resolveRepository();
         }
+        var env = applicationContext.getEnvironment();
+        maxPageSize = env.getProperty("spring.data.rest.pageable.max-page-size",
```
taefi, line 65: I this property should be like this:
```suggestion
        maxPageSize = env.getProperty("spring.data.web.pageable.max-page-size",
```
As the one from `spring.data.rest` is:
```
spring.data.rest.max-page-size
```
and users normally expect the latter to work only while using `spring.data.rest` to automatically expose rest endpoints from their repositories.

taefi (changes requested):

packages/java/endpoint/src/test/java/com/vaadin/hilla/crud/CrudRepositoryServiceJpaTest.java
cromoteca: I don't think so, but I agree that they can be removed.

taefi (approved):

taefi (approved):

PR #3249, opened 2025-02-18
feat(crud): allow to hide new item button

### LINKED ISSUE #3227: Disable New Button in AutoCrud Component
## Description
Currently, there is no built-in way to disable the New button in the AutoCrud component. While the Delete button can be disabled via formProps, there's no equivalent option for the New button. This feature would be useful for creating edit-only views where users should only modify existing records.

## Current Workarounds
- Using CSS to hide the toolbar
- Customizing the CrudService in the backend to remove support for inserting or deleting.

These workarounds are not ideal because:
1. CSS hiding doesn't prevent API calls
2. No clear documentation on implementing this safely

Add a prop to disable the New button, similar to how the Delete button can be disabled.

Add some wording to the docs describing how to safely disable creating new items and deleting existing items.

_No response_

### Additional context

_No response_

cromoteca:
This PR adds adds a `newButton` property to show/hide the new item button.

As an empty toolbar is rendered at half height, it is hidden completely when no button.

Closes #3227


packages/ts/react-crud/src/autocrud.tsx
```diff
@@ -76,6 +76,24 @@ export type AutoCrudProps<TModel extends AbstractModel = AbstractModel> = Compon
...
+    newButtonHidden?: boolean;
+    /**
+     * Whether to show the toolbar with the new button in the grid. This is
+     * `false` by default, so that the toolbar is shown.
+     *
+     * NOTE: While this also hides the new item button, it does not prevent from
+     * sending a new item to the service. Make sure that your backend Java service
+     * is properly secured to prevent unauthorized new items.
+     */
+    toolbarHidden?: boolean;
```
taefi, line 96: Just a consideration:

We called the property to hide the delete button as `deleteButtonVisible`:
https://github.com/vaadin/hilla/blob/main/packages/ts/react-crud/src/autoform.tsx#L194-L209

And for the things that are enabled by default and the user can disable them, we used the naming as `noHeaderFilters`:
https://github.com/vaadin/hilla/blob/main/packages/ts/react-crud/src/autogrid.tsx#L94-L97

I suggest we keep some sort of harmony for naming the properties, with either of the previously existing ones. I personally am in favor of `noToolbar` and `noNewButton`, as the the presence of the property is enough (`noToolbar` means `noToolbar={true}`),

taefi (changes requested):

packages/ts/react-crud/src/autocrud.tsx
taefi: Btw, `noNewButton` is almost the same as `newButtonHidden`, they're both having the default as `false`, and their presence means not showing the UI control they are targeting. The outlier is `deleteButtonVisible` (which should be deprecated in my opinion).

cromoteca: Renamed to `newButtonVisible` at the moment, waiting for the discussion about names to finalize.

cromoteca:
Let's discuss the parameter name in https://github.com/vaadin/hilla/issues/3258

packages/ts/react-crud/src/autocrud.tsx
cromoteca: Renamed to `newButton` following internal choice.

taefi (approved):

PR #3257, opened 2025-02-19
fix(parser): check for misuse of `MultipartFile`

### LINKED ISSUE #3244: `MultipartFile` can be used as or in return type
### Describe the bug

The newly introduced support for `MultipartFile` allows to use it everywhere, including as type of an entity property. But such entities can easily be misused as return types in endpoints.
Although unlikely, `MultipartFile` can also be used as return type directly.
Of course, it is impossible to return a `MultipartFile`, so the generator should throw when `MultipartFile` is used anywhere else than as method parameter.

### Expected-behavior

_No response_

### Reproduction

Try returning a record that has a `MultipartFile` property.

### System Info

24.7.0.alpha9

cromoteca:
Fixes #3244


Lodin (changes requested):
Overall LGTM but there is something I'd like to improve

Lodin (approved):

PR #3259, opened 2025-02-19
fix(file-router): correctly exclude children paths of "excluded" & paths with variable segments

### LINKED ISSUE #3219: File-based Router: routes with required param show up in the menu created using `createMenuItems` utility
### Describe the bug

According to the docs, a file or directory with single curly braces in its name should result in having a required route parameter, which should automatically be excluded from the menu items when the menu is created using `createMenuItems` utility function.

### Expected-behavior

_No response_

### Reproduction

* Download a Vaadin-Hilla project from start.vaadin.com.
* Add `customers` under `views`.
* Add `{id}` directory under `customers`.
* Add an `@index.tsx` view under `{id}` directory:
```tsx
export default function DetailsView() {
  return <h3>Details view</h3>;
}
```
* Add an `edit.tsx` view under `{id}` directory:
```tsx
export default function EditView() {
  return <h3>Edit view</h3>;
}
```
* You should end up in a directory structure like this:
```
views
└─ customers
   └─ {id}
      ├─ @index.tsx
      └─ edit.tsx
...
```
* Run the project.
* Check that those views are in the menu (like this):

Image

Image

### System Info

Vaadin `24.6.4` and `24.7.0.alpha8`

Lodin:
Fixes #3219.


cromoteca (approved):
Verified on a real app

PR #3267, opened 2025-02-20
dummy: for checking GHA changes in #3156

taefi:
This is a Dummy PR just to trigger the `validation.yml` of #3156 to see whether the added `New Gradle tests` is successful or not.


taefi:
Not needed anymore.

PR #3273, opened 2025-02-21
feat: Add a dependency list to useGridDataProvider and useComboBoxDataProvider

Artur-:
This is to support using state in the fetch callback


taefi (approved):
LGTM.

PR #3291, opened 2025-02-27
refactor!(crud): apply parameter naming conventions

### LINKED ISSUE #3258: Uniform configuration parameter names in Crud
### Describe your motivation

Crud currently has `deleteButtonVisible` and `noHeaderFilters`. We're going to add a property to hide the "New" button. This introduced a debate to choose the best name for that property and to uniform existing property names.

We should stick to a pattern. Candidates are:

| Pattern       | Example                 | Pros ✅ | Cons ❌ |
|--------------|-------------------------|--------|--------|
| **`showX`**  | `showNewButton={true}`  | ✅ **Matches UI library conventions** (Material-UI, AntD).<br>✅ **Simple and easy to understand**.<br>✅ **Consistent with other enable/disable patterns**.<br>✅ **Less likely to be confused with other properties**. | ❌ **Redundant for default-visible elements**.<br>❌ Can lead to awkward names (`showShowButton`). |
| **`XVisible`**  | `newButtonVisible={true}` | ✅ **Clearly expresses visibility**.<br>✅ **Reads well in boolean logic** (`if (newButtonVisible)`).<br>✅ Avoids verb confusion (`showShowButton`). | ❌ **More verbose** than `showX`.<br>❌ **Less common in UI libraries**.<br>❌ Could be confused with visibility states (`visible`, `hidden`). |
| **`noX`**  | `noNewButton={true}` | ✅ **Short and explicit negation**.<br>✅ Defaults to `false`, so no prop means "enabled". | ❌ **Unusual and less readable**.<br>❌ **Negative logic (`!noX`) can be confusing**.<br>❌ Doesn't scale well for more complex options. |
| **`X` (Boolean or ReactNode)** | `newButton={true}` / `newButton={<button>New</button>}` | ✅ **Combines toggling and custom overrides**.<br>✅ **Compact and flexible API**.<br>✅ **Removes the need for separate visibility & customization props**. | ❌ **Might be confusing (`newButton` could mean "creating a button" vs. "the button for creating")**.<br>❌ **Not standard in most UI libraries**. |

_No response_

### Additional context

_No response_

cromoteca:
Let's try to stick to agreed naming conventions as much as possible: for a feature `X`, especially if boolean, let's avoid prefixes and suffixes like `no` or `Visible`. So, for example, `noHeaderFilters` becomes `headerFilters`, and `deleteButtonVisible` becomes `deleteButton`.

The recently introduced `newButton` toggler already respects this convention.

Closes #3258.


cromoteca:
With the new convention there's actually nothing to rename.

PR #3295, opened 2025-02-27
feat(endpoint): Add constructors for filter classes

Artur-:
These were added to the Flow versions but not copied back to Hilla. They still are useful when you work with the filter classes yourself


Artur-:
This is to allow using a Hilla service from Flow as

```java
comboBox.setItemsPageable((pageable, filterString) -> productService.list(pageable, new PropertyStringFilter("name", Matcher.CONTAINS, filterString)));
```

instead of

```java
comboBox.setItemsPageable((pageable, filterString) -> {
  PropertyStringFilter filter = new PropertyStringFilter();
  filter.setPropertyId("name");
  filter.setMatcher(PropertyStringFilter.Matcher.CONTAINS);
  filter.setFilterValue(filterString);
  return productService.list(pageable, filter);
});
```

platosha (changes requested):
I appreciate this small and nice feature, thanks! However, it needs some tests.

platosha:
> I appreciate this small and nice feature, thanks! However, it needs some tests.

An unit test would be enough, I think.

Artur-:
What should it test?

platosha:
> What should it test?

Constructors invoked within the use case outlined above: https://github.com/vaadin/hilla/pull/3295#issuecomment-2689914113

Artur-:
Updated tests to use the constructors

platosha (approved):

PR #3314, opened 2025-03-03
fix(parser): add a fallback if AOT fails

### LINKED ISSUE #3297: The new spring aot based hilla generator in Vaadin 24.7.0.beta1 not working at all
### Describe the bug

The problem is that my application can't be spun up properly with spring AOT. First it couldn't determine the datasource type so I tried to add the configuration property spring.datasource.driver-class-name to the configuration but it was ignored. I added the h2 driver to the dependencies and then it got further but recorded a lot of errors in the logs during application startup since it could not be configured correctly and ultimately hung after this line in the log.

```
2025-02-27T20:28:49.712+01:00  INFO 92583 --- [Application] [           main] org.reflections.Reflections              : Reflections took 24 ms to scan 48 urls, producing 421 keys and 2302 values
2025-02-27T20:28:49.734+01:00 ERROR 92583 --- [Application] [           main] c.v.h.springnative.HillaHintsRegistrar   : Resource /hilla-openapi.json is not available

```

To elaborate a bit. I think this is too much of a breaking change to keep in a minor release with the state that it is in. I have made apps run with the AOT compiler before and had to spend days making it work. This time I gave up after 4 hours scratching my head and nothing I did made it work. We are in the process of migrating an old Vaadin 8 application into a Vaadin 24 flow hilla hybrid application and we are trying to stay with the latest possible release all the time and this would hinder our progress significantly. 

### Expected-behavior

The endpoints should be generated. 

### Reproduction

Not sure how to make a simple reproduction since it is a large application and not sure where the app fails to be run with the spring aot processor.

### System Info

Hilla 24.7.0.beta1

cromoteca:
While AOT has proved to be reliable in most cases, it can fail on some projects and we don't want users to have to struggle with it just to compile their apps for production.

This PR adds a fallback in case of AOT failure. All browser-callable services found in the classpath will be reported, which could lead to generation of some TypeScript files not needed by the final application, but the build system will discard unused code anyway.

Also, to reduce noise in the logs, the output of the AOT process is redirected to a file and only debug messages are used.

Fixes #3297

### DIFF
--- a/packages/java/engine-core/src/main/java/com/vaadin/hilla/engine/AotBrowserCallableFinder.java
+++ b/packages/java/engine-core/src/main/java/com/vaadin/hilla/engine/AotBrowserCallableFinder.java
@@ -64,10 +64,9 @@ class AotBrowserCallableFinder {
                         }
                     }).filter(Objects::nonNull).findFirst().orElse(null);
             if (mainClass == null) {
-                LOGGER.warn(
+                LOGGER.debug(
                         "This project has not been recognized as a Spring Boot"
-                                + " application because a main class could not be found."
-                                + " Hilla services will not be available.");
+                                + " application because a main class could not be found.");
             }
             return mainClass;
         } catch (NoClassDefFoundError e) {
@@ -102,6 +101,8 @@ class AotBrowserCallableFinder {
         var argsFile = engineConfiguration.getBuildDir()
                 .resolve("hilla-aot-args.txt");
         Files.write(argsFile, settings);
+        var report = engineConfiguration.getBuildDir()
+                .resolve("hilla-aot-report.txt");
 
         var javaExecutable = ProcessHandle.current().info().command()
                 .orElse(Path.of(System.getProperty("java.home"), "bin", "java")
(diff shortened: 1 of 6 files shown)


taefi (approved):
LGTM!

packages/java/engine-core/src/main/java/com/vaadin/hilla/engine/EngineConfiguration.java
```diff
@@ -129,8 +135,15 @@ public BrowserCallableFinder getBrowserCallableFinder() {
         return () -> {
             try {
                 return AotBrowserCallableFinder.findEndpointClasses(this);
-            } catch (IOException | InterruptedException e) {
-                throw new ExecutionFailedException(e);
+            } catch (Exception e) {
+                if (classFinder != null) {
+                    LOGGER.info(
+                            "AOT-based detection of browser-callable classes failed. Falling back to classpath scan.");
```
taefi, line 141: Maybe we can improve the log message here so that it either:
1. instruct the user to run the build with DEBUG log level, so they get the info were to find the report (if they're interested).
2. to reveal the repot file path here directly, since the report is already there.

packages/java/engine-core/src/main/java/com/vaadin/hilla/engine/AotBrowserCallableFinder.java
```diff
@@ -102,19 +101,24 @@ private static Path generateAotArtifacts(
         var argsFile = engineConfiguration.getBuildDir()
                 .resolve("hilla-aot-args.txt");
         Files.write(argsFile, settings);
+        var report = engineConfiguration.getBuildDir()
+                .resolve("hilla-aot-report.txt");
```
taefi, line 105: Nit: declare this variable as close as possible to where it is needed e.g. line 116.

taefi (changes requested):

packages/java/engine-core/src/main/java/com/vaadin/hilla/engine/EngineConfiguration.java
cromoteca: Added a note to suggest using debug level. That will allow to discover the report file path.

packages/java/engine-core/src/main/java/com/vaadin/hilla/engine/AotBrowserCallableFinder.java
cromoteca: Made another change that makes that order irrelevant

taefi (approved):

PR #3318, opened 2025-03-03
feat(generator-plugin-transfertypes): add generics & importing from the correct module

### LINKED ISSUE #3208: generator-transfer-type plugin is missing support for generics
### Describe the bug

The generator-transfer-type plugin is missing the support for generic types, so for the cases such as not generating the TS interfaces for ListSignal, ValueSignal, NumberSignal, ... it's not simply possible to filter them.

### Expected-behavior

_No response_

### Reproduction

NA

### System Info

N/A

Lodin:
Fixes #3208 

This pull request includes several changes across multiple files to enhance functionality and improve the codebase. The most important changes involve adding new annotations, refactoring methods, and updating dependencies. Below is a summary of the key changes:

### Enhancements and New Features:
* Added new annotation `@FromModule` to specify module information for transfer types in `packages/java/runtime-plugin-transfertypes/src/main/java/com/vaadin/hilla/transfertypes/annotations/FromModule.java`.
* Introduced new `Signal` classes and corresponding test cases to handle different signal types in `packages/java/parser-jvm-plugin-transfertypes/src/test/java/com/vaadin/hilla/signals/`. [[1]](diffhunk://#diff-048db61818e082a162b065cb22c8d5b7a732b8e647983c2d4849e511ef550b99R1-R3) [[2]](diffhunk://#diff-028dc601e2ec66734f3a8721440f91b49d95ea54bfe896d142f64b9588ff2d7dR1-R3) [[3]](diffhunk://#diff-914a6f30b0ab5fb1acd45374d2eea4d9fed43b09746d3a2205b5dc0077127757R1-R3) [[4]](diffhunk://#diff-85c11f63358aa06cfadd96fed9d6efe61868d31e7b6175de60a9beec8f30cb67R1-R3)

### Code Refactoring:
* Refactored `scan` method in `Plugin` interface to provide a default implementation in `packages/java/parser-jvm-core/src/main/java/com/vaadin/hilla/parser/core/Plugin.java`.
* Updated `TransferTypesPlugin` to handle new signal classes and added logic to replace signal classes with local signal records in `packages/java/parser-jvm-plugin-transfertypes/src/main/java/com/vaadin/hilla/parser/plugins/transfertypes/TransferTypesPlugin.java`. [[1]](diffhunk://#diff-a5846aa41bb2c30afc74f3523cc063e7706b43015426adc0fe22ef4b43ef2a0bR5-R41) [[2]](diffhunk://#diff-a5846aa41bb2c30afc74f3523cc063e7706b43015426adc0fe22ef4b43ef2a0bR58-R137)

### Dependency and Import Management:
* Added necessary imports for new signal classes and annotations in `TransferTypesPlugin` and other related files.
* Updated `module-info.java` to export the new `annotations` package in `packages/java/runtime-plugin-transfertypes/src/main/java/module-info.java`.

### Testing and Validation:
* Added comprehensive test cases for new signal functionality in `packages/java/parser-jvm-plugin-transfertypes/src/test/java/com/vaadin/hilla/parser/plugins/transfertypes/signals/SignalTest.java`.

These changes collectively enhance the functionality and maintainability of the codebase by introducing new features, refactoring existing methods, and ensuring proper dependency management.


packages/java/parser-jvm-plugin-transfertypes/src/test/java/com/vaadin/hilla/parser/plugins/transfertypes/signals/SignalEndpoint.java
```diff
@@ -0,0 +1,25 @@
...
+import com.vaadin.hilla.signals.ListSignal;
+import com.vaadin.hilla.signals.NumberSignal;
+import com.vaadin.hilla.signals.Signal;
+import com.vaadin.hilla.signals.ValueSignal;
+
+@Endpoint
+public class SignalEndpoint {
+    public Signal<String> getStringSignal() {
+        return null;
+    }
```
taefi, line 12: I think we should actually prevent this case from happening, though, it is possible in java.

Lodin: What do you mean? It's just a test file, isn't it?

packages/java/parser-jvm-plugin-transfertypes/src/main/java/com/vaadin/hilla/parser/plugins/transfertypes/TransferTypesPlugin.java
```diff
@@ -50,18 +55,59 @@ public final class TransferTypesPlugin
...
+                        .findFirst().ifPresent((annotationModel) -> {
+                            var annotation = (FromModule) annotationModel.get();
+                            var namedSpecifier = annotation.namedSpecifier();
+                            var defaultSpecifier = annotation
+                                    .defaultSpecifier();
+
+                            if (namedSpecifier.isBlank()
+                                    && defaultSpecifier.isBlank()) {
+                                throw new IllegalArgumentException(String
+                                        .format("@FromImport annotation for class %s must have at least one named specifier or a default specifier",
```
taefi, line 92: I assume this is typo:
```suggestion
                                        .format("@FromModule annotation for class %s must have at least one named specifier or a default specifier",
```

packages/java/parser-jvm-plugin-transfertypes/src/test/java/com/vaadin/hilla/signals/Signal.java
```diff
@@ -0,0 +1,3 @@
+package com.vaadin.hilla.signals;
+
+public record Signal<T>() {}
```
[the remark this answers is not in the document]
taefi: Not actually sure why should there be two sets of placeholder classes are needed for the signals, one with annotations on top of them, and one without any annotations. Why wouldn't putting the newly introduced annotation on top of the actual implementation of the `ListSignal`, `ValueSignal`, `NumberSignal`, and others work?

taefi (changes requested):

packages/java/parser-jvm-plugin-transfertypes/src/test/java/com/vaadin/hilla/parser/plugins/transfertypes/signals/SignalEndpoint.java
Lodin: Hmm, that's a valid point but I believe it is out of scope for this PR since it is more about generics in general (pun intended), and signals are only an example for them.

packages/java/parser-jvm-plugin-transfertypes/src/test/java/com/vaadin/hilla/signals/Signal.java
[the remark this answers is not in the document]
Lodin: That's just test mocks to avoid import of the `endpoint` module. I thought it's gonna be easier for the implementation to not have dependency only for tests.

Regarding the second part of your question, I'm purposely avoid using the real implementation for intermediate OpenAPI-only types to avoid any possible pollution of the generation output. The reason is that we have no tools to stop parser/generator traverse, so if the parsed class contains methods that mention non-related dependencies, they will go directly to the list of entities (recursively!), which is definitely not desired. That's why it's easier to simply replace the real class with a placeholder that doesn't have dependencies at all. It is perfectly addresses our need to replace it later in the generator using the correct library type.

packages/ts/generator-plugin-signals/src/SignalProcessor.ts
```diff
@@ -47,293 +49,195 @@ export default class SignalProcessor {
...
+      if (node.kind === SyntaxKind.ArrayType) {
+        modelName = 'ArrayModel';
+      } else if (node.kind === SyntaxKind.StringKeyword) {
+        modelName = 'StringModel';
+      } else if (node.kind === SyntaxKind.NumberKeyword) {
+        modelName = 'NumberModel';
+      } else if (node.kind === SyntaxKind.BooleanKeyword) {
+        modelName = 'BooleanModel';
+      } else {
+        return undefined;
```
taefi, line 235: This could've remained the same as before where it was handled using a static mapping between the `SyntaxKind` and primitive type's model names.

taefi (approved):
Added one comment, but generally, LGTM!

packages/ts/generator-plugin-signals/src/SignalProcessor.ts
taefi: Sounds like a nano optimization 😅 But, `switch` looks better at least!

platosha (changes requested):

platosha (approved):

PR #3325, opened 2025-03-06
fix(endpoint): update file routes security per every request in dev mode

### LINKED ISSUE #2884: Server-side access control is not updated when a new view is added
### Describe the bug

In an application with access control (from start.vaadin.com), the server-side access control rules are not applied immediately when a new view is added. This leads to being redirected to `/login` if you try to navigate to a recently added public view if you are not logged in. The view can be accessed without being logged in after restarting the server.

### Expected-behavior

Expected that the server-side access control settings are updated immediately when a new view is added.

### Reproduction

1. Create an application on start.vaadin.com with access control for a `/foo` view and no access control for a`/bar` view.
2. Launch in dev mode
3. Copy `views/bar.tsx` into `views/baz.tsx`
4. Navigate to the new view from the new menu entry
5. Reload the page and observe that you're redirected to the login page.
6. Restart the server
7. Repeat steps 4 and 5 and observe that the view is still rendered after the reload

### System Info

Vaadin 24.5.2

platosha:
Fixes #2884

### DIFF
--- a/packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUtil.java
+++ b/packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUtil.java
@@ -1,13 +1,11 @@
 package com.vaadin.hilla.route;
 
 import com.vaadin.flow.internal.hilla.FileRouterRequestUtil;
-import com.vaadin.flow.server.VaadinContext;
-import com.vaadin.flow.server.VaadinRequest;
-import com.vaadin.flow.server.VaadinServletContext;
+import com.vaadin.flow.server.*;
 import com.vaadin.flow.server.menu.AvailableViewInfo;
 import com.vaadin.flow.internal.menu.MenuRegistry;
-import com.vaadin.flow.server.startup.ApplicationConfiguration;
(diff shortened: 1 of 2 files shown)


packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUtil.java
```diff
@@ -63,8 +60,12 @@ public void setRoutes(
      */
     @Override
     public boolean isRouteAllowed(HttpServletRequest request) {
-        if (registeredRoutes == null) {
-            collectClientRoutes(request);
+        var config = ApplicationConfiguration
+                .get(new VaadinServletContext(request.getServletContext()));
+        boolean isLiveReloadMode = config.getMode()
+                .equals(Mode.DEVELOPMENT_FRONTEND_LIVERELOAD);
```
cromoteca, line 66: This looks quite heavy and it would be great to determine execution mode once and for all, but let's avoid premature optimization.

cromoteca (approved):
I have a performance doubt, but otherwise LGTM

PR #3326, opened 2025-03-06
fix(crud): use `noNewButton` name and fix form visibility

### LINKED ISSUE #3324: New item form in AutoCrud remains enabled even if the New button is hidden
### Describe the bug

If you click the New button to enable the New item form in AutoCrud and then do something that causes `newButton` to be set to false, then the New item form remains active.

### Expected-behavior

Expected that I can no longer enter a new item if the feature is disabled even tough I could do that previously

### Reproduction

1. Click the New button
2. Do something that causes `newButton` to be set to false
3. Observe that the New item form is still enabled

### System Info

Vaadin 24.7.0.beta1

cromoteca:
Changes the recently introduced `newButton` property to `noNewButton` and hides the empty form when the new property is `true`.

Fixes #3324

### DIFF
--- a/packages/ts/react-crud/src/autocrud.tsx
+++ b/packages/ts/react-crud/src/autocrud.tsx
@@ -218,7 +218,7 @@ export function AutoCrud<TModel extends AbstractModel>({
     </div>
   );
 
-  // If the "new" button is visible, the form is always shown.
+  // If the "New" button is visible, the form is always shown.
   // Otherwise, the form is only shown when an item is being edited.
   const formEnabled = noNewButton !== true || (item && item !== emptyItem);
   return formEnabled ? editableGrid : mainSection;
--- a/packages/ts/react-crud/src/autocrud.tsx
+++ b/packages/ts/react-crud/src/autocrud.tsx
@@ -218,6 +218,8 @@ export function AutoCrud<TModel extends AbstractModel>({
     </div>
   );
 
+  // If the "new" button is visible, the form is always shown.
+  // Otherwise, the form is only shown when an item is being edited.
   const formEnabled = noNewButton !== true || (item && item !== emptyItem);
   return formEnabled ? editableGrid : mainSection;
 }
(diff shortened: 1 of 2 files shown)


packages/ts/react-crud/src/autocrud.tsx
```diff
@@ -77,14 +77,14 @@ export type AutoCrudProps<TModel extends AbstractModel = AbstractModel> = Compon
...
     /**
-     * Determines whether to display the "New" button in the toolbar. By default,
-     * this is set to `true`, meaning the button will be shown.
+     * Determines whether to hide the "New" button in the toolbar. By default,
+     * this is set to `false`, meaning the button will be shown.
      *
      * NOTE: This setting only hides the button; it does not prevent new items
      * from being sent to the service. Ensure your backend Java service is
      * properly secured to prevent unauthorized creation of new items.
      */
```
Lodin, line 86:
```suggestion
     * Determines whether to hide the "New" button in the toolbar.
     *
     * NOTE: This setting only hides the button; it does not prevent new items
     * from being sent to the service. Ensure your backend Java service is
     * properly secured to prevent unauthorized creation of new items.
     *
     * @defaultValue `false` meaning the button will be shown.
     */
```

packages/ts/react-crud/src/autocrud.tsx
```diff
@@ -162,7 +162,7 @@ export function AutoCrud<TModel extends AbstractModel>({
       ></AutoGrid>
       {/* As the toolbar only contains the "New" button at the moment, and as an empty toolbar
           renders as a half-height bar, let's hide it completely when the button is hidden */}
-      {newButton !== false && (
+      {noNewButton !== true && (
```
Lodin, line 165:
```suggestion
      {!noNewButton && (
```

packages/ts/react-crud/src/autocrud.tsx
```diff
@@ -197,7 +197,7 @@ export function AutoCrud<TModel extends AbstractModel>({
     />
   );
 
-  return (
+  const editableGrid = (
```
Lodin, line 200: I'd suggest using `if` block here to avoid unnecessary creation of a JSX object in case form is not enabled. Something like:

```tsx
if (noNewButton !== true || (item && item !== emptyItem)) {
  // return created editable grid 
}

// return created mainSection
```

Lodin (changes requested):

Lodin (approved):

PR #3341, opened 2025-03-11
fix(react-crud): make height consistent regardless of button toolbar presence

### LINKED ISSUE #3322: AutoCrud height changes when the new button is disabled
### Describe the bug

The default height of AutoCrud is different depending on whether the new button is enabled. The height difference seems to be exactly the same as the height of the toolbar.

Image

### Expected-behavior

Expected that the height stays the same when toggling.

### Reproduction

This is most visible when toggling the state but can also be shown side-by-side like this:
```
<section className="flex items-start">
  <AutoCrud service={HelloWorldService} newButton={false} model={PersonModel} />
  <AutoCrud service={HelloWorldService} model={PersonModel} />
</section>
```

Note that without `items-start`, the flexbox with stretch the smaller one to the same size as the natural size of the larger one.

### System Info

Hilla 24.7.0.beta1

platosha:
Fixes #3322

### DIFF
--- a/packages/ts/react-crud/src/autocrud.obj.css
+++ b/packages/ts/react-crud/src/autocrud.obj.css
@@ -2,13 +2,15 @@
   display: flex;
   overflow: hidden;
   border: solid 1px var(--lumo-contrast-20pct);
+  height: 462px; /* vaadin-grid intrinsic 400px + 62px for the button toolbar */
 }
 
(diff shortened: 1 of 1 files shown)


Lodin (approved):

cromoteca (approved):
Tested on a real project

PR #3365, opened 2025-03-20
feat(file-router): implement lazy route generation

### LINKED ISSUE #3264: Make the FS router support lazy loading
### Describe your motivation

In a big application, views should be split up into separate JS bundles so that the main bundle doesn't keep growing as more views are added.

Add a `lazy: boolean` field to `ViewConfig`. The default value is `false` for a view mapped to `/` or `/login` and `true` for all other views. (Yes, this is yet another property where the default is `true` but that's fine-ish since `lazy` is much more descriptive than any other name that I've considered).

For a view that is lazy (either explicitly configured or based on the default), we generate `file-routes.ts` to do `const ViewName = lazy(() => import("../views/view-name.js"))` instead of the current `import * as ViewName from "../views/view-name.js"`.

We might also want to make it possible to group multiple related views into a single import instead of creating a separate import for each view. We can add that later by changing the type of `lazy` to `boolean | string` and changing the logic for generating `file-routes.ts` so that all views using the same `lazy` string value are imported from the same URL (by generating a separate helper file that imports and re-exports all desired views).

### Additional context

_No response_

Lodin:
Resolves #3264.


Lodin:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=vaadin_fusion&pullRequest=3365) **Quality Gate failed**
> Failed conditions  [10.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=vaadin_fusion&pullRequest=3365&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=vaadin_fusion&pullRequest=3365)

Ignore this failure because it complains about code duplication in fixtures which is acceptable.

platosha (approved):

PR #3399, opened 2025-04-11
build: remove Vite copies workaround

platosha:
Depends on vaadin/flow#21283


platosha:
Converted to draft, as this depends on the change in Flow.

Lodin (approved):

PR #3401, opened 2025-04-11
feat(endpoint): add custom HTTP exception

### LINKED ISSUE #3376: Allow customizing the HTTP status code in Hilla endpoints
### Describe your motivation

I would like to return an 403 status code from within the Hilla endpoint. In my case it is not viable to use access check annotations because of dynamic logic inside the endpoint.

Throwing an `EndpointAccessDeniedException`from within a Hilla endpoint confusingly ends with a 400 Bad Request status code.

Either allow throwing `EndpointAccessDeniedException` (and similar) or rely on the standard spring `ResponseStatusException` for controlling http status codes.

_No response_

### Additional context

_No response_

cromoteca:
Introduce a new `EndpointHttpException` class to handle HTTP status codes in endpoint invocations, allowing for more granular error handling.

Other exceptions become subclasses of this one, which simplifies the handling of those exceptions around the code.

Both `EndpointHttpException` and its subclasses can be thrown by user code in endpoints, and they are correctly rethrown by Hilla.

Closes #3376


packages/java/endpoint/src/main/java/com/vaadin/hilla/EndpointController.java
```diff
@@ -334,17 +332,13 @@ private ResponseEntity<String> doServeEndpoint(String endpointName,
...
-            return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(
-                    endpointInvoker.createResponseErrorObject(e.getMessage()));
-        } catch (EndpointBadRequestException e) {
-            return ResponseEntity.badRequest().body(
-                    endpointInvoker.createResponseErrorObject(e.getMessage()));
-        } catch (EndpointInternalException e) {
-            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(
-                    endpointInvoker.createResponseErrorObject(e.getMessage()));
+        } catch (EndpointHttpException e) {
+            var resp = ResponseEntity.status(e.getHttpStatusCode());
```
platosha, line 336: Should we also disallow throwing `2xx` status codes to prevent any abuse of this mechanic? It is not supposed to involve success handling on the client...

platosha (changes requested):

platosha (approved):

PR #3416, opened 2025-04-16
build: use Flow versions for React 18

platosha:
Change the way React 18 is installed in its GHA job to utilize Flow versions and alignment scripts. This way we make sure that the 19 → 18 step gets done for entire workspace simultaneusly and guarantee that the same React 18 is used in testing.


platosha:
Run with the workflow changes: https://github.com/vaadin/hilla/actions/runs/14494111985

Lodin (approved):

PR #3422, opened 2025-04-17
fix: display integer and decimal zero values correctly

### LINKED ISSUE #3396: Bug: Zero value not displayed
### Describe the bug

When a numeric value is 0, it is currently rendered as an empty string instead of "0".
This happens because the formatter checks for truthy values using value && Number.isFinite(value), which skips 0.

---
`local.ts` in `react-crud` packages makes that.
```typescript
  formatInteger(value?: number): string {
    return value && Number.isFinite(value) ? this.#integer.format(value) : '';
  }

  formatDecimal(value?: number): string {
    return value && Number.isFinite(value) ? this.#decimal.format(value) : '';
  }
```

Maybe it should be changed below.
```typescript
  formatInteger(value?: number): string {
    return typeof value === 'number' && Number.isFinite(value) ? this.#integer.format(value) : '';
  }

  formatDecimal(value?: number): string {
    return typeof value === 'number' && Number.isFinite(value) ? this.#decimal.format(value) : '';
  }
```

### Expected-behavior

Value 0 should be displayed as "0".

### Reproduction

Render a value of 0 in the AutoGrid.

### System Info

Windows 11
Intellij

rbrki07:
Fixes #3396


packages/ts/react-crud/test/autogrid.spec.tsx
```diff
@@ -1713,6 +1713,16 @@ describe('@vaadin/hilla-react-crud', () => {
         expect(grid.getBodyCellContent(3, columnIndex)).to.have.rendered.text('-12.00');
       });
 
+      it('renders integer zero value as 0', async () => {
+        const columnIndex = await grid.findColumnIndexByHeaderText('Integer');
+        expect(grid.getBodyCellContent(4, columnIndex)).to.have.rendered.text('0');
```
cromoteca, line 1718: Maybe adding this line to `renders integers as right aligned numbers` is enough, instead of adding a new test. Same for the other test.

cromoteca (changes requested):

packages/ts/react-crud/src/locale.ts
```diff
@@ -84,11 +84,11 @@ export class LocaleFormatter {
   }
 
   formatInteger(value?: number): string {
-    return value && Number.isFinite(value) ? this.#integer.format(value) : '';
+    return value !== undefined && Number.isFinite(value) ? this.#integer.format(value) : '';
```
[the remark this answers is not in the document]
rbrki07: It has to be `return Number.isFinite(value) ? this.#integer.format(value!) : '';`, because `this.#integer.format` expects `value` not to be undefined and `Number.isFinite(value)` alone won't let TypeScript treat `value` as undefined.

packages/ts/react-crud/test/autogrid.spec.tsx
rbrki07: Personally, I would prefer to have separate tests with a stronger context, because the correct zero value formatting has nothing to do with right alignment.

packages/ts/react-crud/src/locale.ts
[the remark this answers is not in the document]
cromoteca: What I mean is that `value !== undefined` is not necessary as `Number.isFinite(undefined) === false`, so `format` won't be called for it anyway.

packages/ts/react-crud/test/autogrid.spec.tsx
cromoteca: OK for me. The suggestion came from SonarQube reporting code duplication.

packages/ts/react-crud/src/locale.ts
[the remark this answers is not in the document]
rbrki07: Yes, that's fine. But `!` is still required, because otherwise TypeScript will complain. Can we keep it as it is now?

packages/ts/react-crud/test/autogrid.spec.tsx
rbrki07: Let's leave it as it is now, ok?

cromoteca (approved):
LGTM. Thanks for contributing this fix!

rbrki07:
Is there anything I can do to fix the current build error: https://github.com/vaadin/hilla/actions/runs/14514413827/job/40721491842?pr=3422?

cromoteca:
@rbrki07 do you need this fix in 24.7 or another version?

platosha:
> Is there anything I can do to fix the current build error: https://github.com/vaadin/hilla/actions/runs/14514413827/job/40721491842?pr=3422?

It's not related to changes here.

Please follow https://github.com/vaadin/hilla/pull/3424, it should address the build failure. Once it's merged, just updating this PR to the main branch should be enough.

cromoteca:
> Is there anything I can do to fix the current build error: https://github.com/vaadin/hilla/actions/runs/14514413827/job/40721491842?pr=3422?

It's not related to your PR. We're working on it, sorry for the inconvenience.

PR #3425, opened 2025-04-17
fix: AutoCrud in flexbox uses full width

### LINKED ISSUE #3323: AutoCrud in a flexbox does not use the full width on small screens
### Describe the bug

The grid section of AutoCrud shrinks rather than grows when it becomes so narrow that the form is hidden and instead opens as a modal popover.

Slightly wider browser window:
Image

Slightly narrower browser window:
Image

### Expected-behavior

Expected that the grid fills the full available width in a narrow browser.

### Reproduction

```
<section className="flex">
  <AutoCrud service={HelloWorldService} model={PersonModel} />
</section>
```

### System Info

Hilla 24.7.0.beta1
Chrome 133

rbrki07:
Fixes #3323

### DIFF
--- a/packages/ts/react-crud/src/autocrud.obj.css
+++ b/packages/ts/react-crud/src/autocrud.obj.css
@@ -1,4 +1,5 @@
 .auto-crud {
+  flex: 1;
   display: flex;
   overflow: hidden;
   border: solid 1px var(--lumo-contrast-20pct);


packages/ts/react-crud/src/autocrud.obj.css
```diff
@@ -1,4 +1,5 @@
 .auto-crud {
+  flex: 1;
```
platosha: Let's use `flex: auto`, it is more robust due to keeping flex-basis.

```suggestion
  flex: auto;
```

platosha (approved):

PR #3429, opened 2025-04-18
Fix #3428 Fixed subtype hierarchy search and property names

### LINKED ISSUE #3428: Property name on JsonTypeInfo not used and type hierarchies don't generate the type field
### Describe the bug

For polymoprhic types annotated with @JsonSubTypes, there are a couple issues. First, when you specify a different property name for the discriminator on the root type, it does not utilize that property name on the generated typescript. Second, it does not apply a type guard to the root type if it appears in the subtypes list. Third, if one of the subtypes is deeper in the hierachy, it does not apply a type guard.

### Expected-behavior

When it generates typescript interfaces representing the polymorphic type on a return or parameter field of a browser callable, the type discriminator should use the value defined in @JsonTypeInfo(property = "..."). Currently, it is always using @type, which is the default for Jackson. 

Also, when you use the base class as one of the sub types, it doesn't place the type guard property field.

Lastly, when there is an inheritance hierarchy, the sub type interfaces do not have the type guard on them.

### Reproduction

With this example browser callable: 

```
@Endpoint
public class TestService {
    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "valueType")
    @JsonSubTypes({
        @JsonSubTypes.Type(value = MyTest.class, name = "base"),
        @JsonSubTypes.Type(value = DerivedTest.class, name = "derived"),
        @JsonSubTypes.Type(value = DeeperTest.class, name = "deeper")
    })
    public static class MyTest {
        public Long foo;
    }
    
    public static class DerivedTest extends MyTest {
        public Long bar;
    }

    public static class DeeperTest extends DerivedTest {
        public Long meh;
    }

    public MyTest example() {
        return new MyTest();
    }
}
```

This generates:

```
interface MyTest {
    foo?: number
}
export default MyTest
```
which should've included the property `valueType : "base"`

```
import type MyTest_1 from "./MyTest.js";
interface DerivedTest extends MyTest {
    bar?: number;
    "@type": "derived"
}
export default DerivedTest
```
where `@type` should've been `valueType`

```
import type DerivedTest_1 from "./DerivedTest.js"
interface DeeperTest extends DerivedTest_1 {
    meh?: number;
}
export default DeeperTest
```
which should've included property `valueType: "deeper"`

### System Info

Windows 10, Vaadin 24.7.2, Chrome 135.0.7049.85

FatalCatharsis:
Fix #3428 find JsonSubTypes on supertype hierarchy and use specified property name

## Description

Recurses up through all superclasses and finds the first instance of JsonSubTypes to use, instead of looking only at the direct ancestor. Also uses the property on the related JsonTypeInfo, instead of assuming it's the default `@type` property.

Just reported this today and figured I'd spend an hour or so since the fix was really obvious. A more aggressive change I kind of wanted to make was replacing the polymorphic type on the endpoint method with the created union type, but I couldn't grok the plugin logic enough to work out if that was possible, let me know if that's desirable and possible.

note: the provided eclipse settings file made a couple whitespace changes, so I'm thinking the file was made before you had the eclipse style xml. Either which way, I promise I applied it :)

Fixes #3428

## Type of change

- [x] Bugfix
- [ ] Feature

## Checklist


cromoteca (changes requested):
Hi @FatalCatharsis,
Thank you for contributing this. It is really appreciated.
Before merging, I'm asking you to make some minor changes to better match existing code.
Also, please run `mvn formatter:format` before committing, as otherwise the PR check will fail.

lucyydotp:
Hi folks, we've recently started using Hilla and use sealed class hierarchies quite a lot, having them reflected properly in TS code would be really nice. Is there anything I can do to help get this over the line?

jessejia1991 (changes requested):
Comments from testing tools, sorry for that

PR #3431, opened 2025-04-22
feat(router): add user-defined `detail` type to `ViewConfig`

### LINKED ISSUE #2564: Allow defining custom FS route configuration
### Describe your motivation

Custom menu building logic would benefit from a way of defining custom metadata in the route configuration. As an example, you might want to manually define a hierarchy or provide an additional row of text that should be shown separately from the view title.

The initial FS router design suggested that unknown properties from `export const config` would just be passed through as-is so that they could be used in various parts of the system (not only for generating the menu). This does currently not seem to work as it was defined in the design.

After further consideration, I think it might be better to designate a specific property for this purpose since that makes the type definitions less unruly and removes concerns for backwards compatibility if we introduce new properties in `ViewConfig`. There would thus be a `detail?: unknown` property in `ViewConfig` and a corresponding property in `MenuItem`. `createMenuItems` could have a type parameter for asserting the `detail` type.

I could then do something like this in my view:
```ts
  title: 'Foo',
  detail: { description: 'Bar'} as MyViewDetails
 };
```

And render the menu like this:
```ts
{createMenuItems<MyViewDetails>().map(({ to, title, icon, detail}) => (
  <SideNavItem path={to} key={to}>
    {icon ? <Icon src={icon} slot="prefix"></Icon> : <></>}
    {title}<span>{detail.description}</span>
  </SideNavItem>
))}
```

The `icon` string can be abused to pass arbitrary data as long as you properly extract values out of the value when rendering the menu item.

### Additional context

_No response_

cromoteca:
Introduce generics to the ViewConfig and related types to allow for more flexible detail objects in menu items and view configurations. 

Depends on https://github.com/vaadin/flow/pull/21311
Closes #2564

### DIFF
--- a/packages/ts/file-router/src/runtime/createMenuItems.ts
+++ b/packages/ts/file-router/src/runtime/createMenuItems.ts
@@ -23,7 +23,7 @@ function hasVariablePathSegment(path: string): boolean {
  *
  * @returns A list of menu items.
  */
-export function createMenuItems(): readonly MenuItem[] {
+export function createMenuItems<T = unknown>(): ReadonlyArray<MenuItem<T>> {
   // @ts-expect-error: esbuild injection
   // eslint-disable-next-line @typescript-eslint/no-unsafe-call
   __REGISTER__('createMenuItems', (window as VaadinWindow).Vaadin);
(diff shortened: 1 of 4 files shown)


taefi (approved):
LGTM!

Lodin (approved):

PR #3490, opened 2025-05-16
fix(frontend): make doFormLayout async and waiting

### LINKED ISSUE #3487: PiT 24.8: Regression: in authenticated apps logout does not work - form is never submitted to the server
### Describe the bug

In any Hilla react with authentication downloaded from start.vaadin.com, logout does not work when firing a click on the logout button.

In a further investigation I've seen that:
- The issue is in `24.8.0.alpha8` and `24.7.4` but not in `24.8.0.alpha7` and `24.7.3`
- It does not work when firing a synthetic click `document.querySelectorAll('vaadin-button')[0].click()` but work when clicking manually
- It happens in Chrome but not in FF
 

### Expected-behavior

There is no regression

### Reproduction

1. Create a project with start wizard and run it:

```
curl 'https://start.vaadin.com/dl?preset=default&preset=partial-react&preset=partial-auth&testdata=true' --output my-app.zip
unzip my-app.zip
cd my-app
mvn -Pproduction
```
2. Open it in browser and log into de app e.g. user/user
3. Open debug console and run `document.querySelector('vaadin-button').click()`
    You can see that page reloads, but login screen is not shown, you can try several times with the same result
4. Click manually on the logout button and you can see that application logs out correctly

watch this video with the issue

Image

5. Downgrade vaadin version:

```
npx upgrade-vaadin 24.7.3
mvn -Pproduction
```

6. Repeat the process of firing click from browser debug console, and you can see that it works as shown in this video:

Image

You can repeat the process with different versions like `24.8.0.alpha7` and `24.8.0.alpha6` and check that regression was introduced in `24.8.0.alpha7` and `24.7.4`

### System Info

Hilla 24.7.4 & 24.8.0.alpha7
Browser: Chrome

cromoteca:
Transform the `doFormLogout` function to be asynchronous, adding a timeout mechanism to allow existing code to await.

Update related test.

The secondary effect is that users need to continue awaiting if they don't want to ignore the promise.

Fixes #3487

### DIFF
--- a/packages/ts/frontend/src/Authentication.ts
+++ b/packages/ts/frontend/src/Authentication.ts
@@ -79,6 +79,9 @@ async function doFormLogout(url: URL | string, parameters: Record<string, string
   // Append form to page and submit it to perform logout and redirect
   document.body.appendChild(form);
 
+  // No code should run after a form submission, as it will navigate away.
+  // The promise will reject after a long timeout to avoid executing code after
+  // (old user code has a `reload` call that could happen before the form submission).
   return new Promise((_, reject) => {
     setTimeout(() => {
       reject(new Error('Form submission did not navigate away after 10 seconds.'));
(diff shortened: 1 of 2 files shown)


taefi (approved):
LGTM

PR #3528, opened 2025-05-27
build: make Gradle Windows build faster

platosha:


platosha:
See branch workflow run: https://github.com/vaadin/hilla/actions/runs/15276106415

PR #3547, opened 2025-05-30
fix(frontend): correct Atmosphere and related environment-dependant imports

### LINKED ISSUE #3546: Reactive endpoints are broken in beta1
### Describe the bug

When using 24.8 beta1 Hilla TS packages, reactive endpoints don't trigger a server call.

### Expected-behavior

_No response_

### Reproduction

Upgrade an application from alpha8 to snapshot, will use beta1 in package.json

### System Info

Vaadin 24.8-SNAPSHOT

platosha:
Fixes #3546

### DIFF
--- a/packages/ts/frontend/src/CsrfInfoSource.ts
+++ b/packages/ts/frontend/src/CsrfInfoSource.ts
@@ -1,8 +1,6 @@
 /// <reference lib="webworker" />
 // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-import { VAADIN_BROWSER_ENVIRONMENT } from './utils.js';
-
-const cookieManagerPromise = VAADIN_BROWSER_ENVIRONMENT ? import('./CookieManager.js') : undefined;
+import CookieManager from './CookieManager.js';
 
 /** @internal */
 export type NameValueEntry = readonly [name: string, value: string];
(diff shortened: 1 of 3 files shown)


packages/ts/frontend/src/CsrfInfoSource.ts
```diff
@@ -1,6 +1,8 @@
 /// <reference lib="webworker" />
 // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-const cookieManagerPromise = globalThis.document ? import('./CookieManager.js') : undefined;
+import { VAADIN_BROWSER_ENVIRONMENT } from './utils.js';
+
+const cookieManagerPromise = VAADIN_BROWSER_ENVIRONMENT ? import('./CookieManager.js') : undefined;
```
taefi, line 5: It seems that this conditional import isn't needed here(?), since the only usage in `extractCsrfInfoFromMeta` is always given a non-null document as an argument.

taefi (changes requested):

packages/ts/frontend/src/CsrfInfoSource.ts
cromoteca: Probably the function is never called from service workers, but still the import needs to be done conditionally to avoid an error.

cromoteca (approved):

packages/ts/frontend/src/CsrfInfoSource.ts
platosha: We've tried a different technique to make Vite drop the Cookie Manager code from the service worker, and turns out declaring `"sideEffects"` is even better.

PR #3559, opened 2025-06-03
fix: implement `translateDynamic`

### LINKED ISSUE #3553: [i18n] using `key` in `ViewConfig` causes an error
### Describe the bug

I want to define `title` in `ViewConfig` as `I18nKey` so that I can use it with `translate`.
Doing that, though, causes an error.

Using
```typescript
const script = new Script(`function key(x) { return x[0]; } (${code})`);
```
in `createViewConfigJson.ts` fixes the issue and allows the keys to be collected, but those coming from `ViewConfig` should go in the top-level chunk.

### Expected-behavior

`key`-tagged literals are accepted.

### Reproduction

```typescript
  title: key`home.title`,
};
```

### System Info

Vaadin 24.8.0.beta2

cromoteca:
Allows to use `key` in `ViewConfig`.

Allows to request single keys to the server.

Implements a `translateDynamic` function that accepts a string as key and returns a signal. To be used in `ViewConfig`-related code. Can also be used when a dynamic key is needed, although at a performance cost.

Requires https://github.com/vaadin/flow/pull/21614

Fixes #3553


platosha (dismissed):

packages/ts/react-i18n/src/index.ts
```diff
@@ -150,43 +154,96 @@ export class I18n {
...
+    }
+    // Wait to possibly collect other synchronously requested keys
+    this.#batchedKeysPromise = Promise.resolve(this.#batchedKeys);
+    await this.#batchedKeysPromise;
+    this.#batchedKeysPromise = undefined;
+    const batchedKeys = [...this.#batchedKeys];
+    return this.updateLanguage({ keys: batchedKeys }).then(() => {
+      console.warn(
+        `A server call was made to translate keys those were not loaded:\n${batchedKeys.map((k) => `  - ${k}`).join('\n')}`,
+      );
```
cromoteca, line 189:
```suggestion
console.warn(['A server call was made to translate keys those were not loaded', ...batchedKeys].join('\n - '));
```

platosha (approved):

packages/ts/react-i18n/src/index.ts
```diff
@@ -150,43 +154,94 @@ export class I18n {
...
+    for (const k of nonBatchedKeys) {
+      this.#batchedKeys.add(k);
+    }
+    // Wait to possibly collect other synchronously requested keys
+    this.#batchedKeysPromise = Promise.resolve(this.#batchedKeys);
+    await this.#batchedKeysPromise;
+    this.#batchedKeysPromise = undefined;
+    const batchedKeys = [...this.#batchedKeys];
+    return this.updateLanguage({ keys: batchedKeys }).then(() => {
+      console.warn(['A server call was made to translate keys those were not loaded:', ...batchedKeys].join('\n - '));
```
platosha, line 187: Let's make tests pass:

```suggestion
      console.warn(['A server call was made to translate keys those were not loaded:', ...batchedKeys].join('\n  - '));
```

platosha (approved):

PR #3568, opened 2025-06-04
chore: bump ts package versions to 25

taefi:


cromoteca:
25 alpha is now released

Issue #2571, opened 2024-06-24
Flow view with default route it is not loaded in hybrid application


MarcinVaadin:
### Describe the bug

Flow view with default route it is not loaded in hybrid application.

[Slack thread](https://vaadin.slack.com/archives/C6X43FE8M/p1719218290742189)

From @caalador:
> I can see hilla generating into file-routes.ts the line createRoute("", Layout0) which I expect overrides the flow "" route as it's a match on the client router

### Expected-behavior

View should be loaded and rendered

### Reproduction

Create new Start project, add new Custom View with single button, download it as hybrid application.

### System Info

Hilla: 24.4.1
Flow: 24.4.2
Vaadin: 24.4.3
Copilot: 24.4.3
Copilot IDE Plugin: 1.0.9-intellij
Java: Oracle Corporation 22.0.1
Java Hotswap: false
Frontend Hotswap: Enabled, using Vite
OS: aarch64 Mac OS X 14.5
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/124.0.0.0 Safari/537.36

Artur-:
Fixed in #2572, released in 24.4.4

PR #3576, opened 2025-06-06
fix(router): always add children, even if empty

### LINKED ISSUE #3563: Flow view is not render when there is no React view in Full Stack Application
### Description of the bug

I downloaded a full-stack React Flow project from the start. Then I created some flow views, then decided to remove all hilla views. Suddenly, the Flow View stopped rendering. Then I reverted the changes and realized that at least you have one React view should be present to render the Flow View.

### Expected behavior

The application should render a flow view regardless a React view is present or not. 

Run the application in this commit and a blank page will be opened

https://github.com/vaadin/hackathon-24-8/commit/d020072f4c8304f1210b42a4a2ee1f6c424a527a

Then check out https://github.com/vaadin/hackathon-24-8/commit/7fa00a253e021f58b70714115a6c3278b593ec11 commit and will see that the page is rendered properly. 

### Versions

Hilla: 24.8.0.beta2
Flow: 24.8.0.beta2
Vaadin: 24.8.0.beta2
Spring Boot: 3.5.0
Spring: 6.2.7
Spring Security: 
Spring Data JPA: 
Copilot: 24.8.0.beta2
Frontend Hotswap: Enabled ⋅ Vite
OS: aarch64 Mac OS X 14.7.1
Java: JetBrains s.r.o. 21.0.5
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36
Java Hotswap: Java Hotswap is enabled
IDE Plugin: 1.0-SNAPSHOT IntelliJ

cromoteca:
Always adding children in generated file-router.ts seems not to have side effects and fixes the issue.

Fixes #3563.

### DIFF
--- a/packages/ts/file-router/src/vite-plugin/createRoutesFromMeta.ts
+++ b/packages/ts/file-router/src/vite-plugin/createRoutesFromMeta.ts
@@ -178,7 +178,7 @@ export default routes;`.source.statements,
         : module.configId;
     }
 
-    const _children = factory.createArrayLiteralExpression(children ?? [], true);
+    const _children = children ? factory.createArrayLiteralExpression(children, true) : '';
 
     // ```ts
     // createRoute("grandparent", {...grandparentConfig, flowLayout: true}, [
--- a/packages/ts/file-router/src/vite-plugin/createRoutesFromMeta.ts
+++ b/packages/ts/file-router/src/vite-plugin/createRoutesFromMeta.ts
@@ -178,7 +178,7 @@ export default routes;`.source.statements,
         : module.configId;
     }
 
-    const _children = children && children.length > 0 ? factory.createArrayLiteralExpression(children, true) : '';
+    const _children = factory.createArrayLiteralExpression(children ?? [], true);
 
     // ```ts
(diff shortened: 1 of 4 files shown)


platosha (changes requested):
Seems excessive. It basically turns every route into layout route.

As an effect, Flow fallback routes would be added to every leaf. For instance, with a request for `/reactView/flowView` would match `<ReactView><FlowView /></ReactView>`, although this could not possibly render correctly, because leaf views do not render the necessary `<Outlet/>`.

Only the layouts are expected to have children.

platosha:
> Seems excessive. It basically turns every route into layout route.
> 
> As an effect, Flow fallback routes would be added to every leaf. For instance, with a request for `/reactView/flowView` would match `<ReactView><FlowView /></ReactView>`, although this could not possibly render correctly, because leaf views do not render the necessary `<Outlet/>`.
> 
> Only the layouts are expected to have children.

Addressed myself.

platosha:
BTW this is another occrence of #2571

PR #3610, opened 2025-06-13
chore(i18n): rename `k` param to `key` for public API

cromoteca:
The `k` name is not good for public API and is worth an internal rename of the function with the same name.


platosha (changes requested):
Would it make sense to also rename `i18n.translate(k)` and `i18n.translateDynamic(k)` the same way?

PR #3769, opened 2025-07-21
chore: do not rely on transitive JCIP annotations

platosha:


platosha:
Ah, fixed by #3765 already.

PR #3806, opened 2025-07-28
fix(test): simplify item selection logic

cromoteca:
Change how an item in a select is found using TestBench to fix error in test.


platosha (approved):

cromoteca:
LGTM. @Lodin can you have a look?

Issue #3697, opened 2025-07-01
[24.8] CSRF token handling is broken in Stateless Security


taefi:
### Describe the bug

Based on the reports in the [forum post](https://vaadin.com/forum/t/csrf-handling-changes-from-24-7-to-24-8/178058/1) about abnormal behavior of CSRF token handing in applications with stateless security, it seems that the changes for sharing the CSRF token with Service Worker introduced a bug that results in getting 401 for subsequent requests after a successful login. 
Disabling the CSRF (which makes the application work) is not an option in production since it makes the application vulnerable.

### Expected-behavior

As by default the JWT token is set in cookies, the application should work with CSRF enabled.

### Reproduction

Needs reproduction.

### System Info

V24.8

PR #3807, opened 2025-07-28
test(security): verify setup with sessions disabled and stateless security

platosha:
Connected to #3697

Depends on vaadin/flow#21939


packages/java/tests/spring/security/src/main/java/com/vaadin/flow/spring/fusionsecurity/SecurityConfig.java
```diff
@@ -68,6 +69,14 @@ public SecurityFilterChain securityFilterChain(HttpSecurity http,
                     .requestMatchers("/error/**").permitAll();
         });
 
+        if (stateless) {
```
cromoteca, line 72: Does this configuration block need to be in any starter?

cromoteca (approved):

PR #3913, opened 2025-08-21
fix(engine-core): always consume Process output

### LINKED ISSUE #3696: Hilla Frontend Code generation gets stuck on windows after adding another Endpoint/BrowserCallable
### Describe the bug

On Windows, on a previously working projekt. When i add another Endpoint it seems to hit some threshold where the build (running locally in dev mode or on our windows-jenkins-server in production-build) gets stuck. 

- Java 21
- node: 20.11.1
- vaadin-version: tested with 24.7.1 as well as 24.8.1

For a while i noticed that on startup the generation code will produce output like this
```
(node:33616) Warning: Closing file descriptor 3 on garbage collection
(Use `node --trace-warnings ...` to show where the warning was created)
(node:33616) [DEP0137] DeprecationWarning: Closing a FileHandle object on garbage collection is deprecated. Please close FileHandle objects explicitly using FileHandle.prototype.close(). In the future, an error will be thrown if a file descriptor is closed during garbage collection.
(node:33616) Warning: Closing file descriptor 4 on garbage collection
(node:33616) Warning: Closing file descriptor 5 on garbage collection
(node:33616) Warning: Closing file descriptor 6 on garbage collection
(node:33616) Warning: Closing file descriptor 24 on garbage collection
(node:33616) Warning: Closing file descriptor 7 on garbage collection
(node:33616) Warning: Closing file descriptor 8 on garbage collection
(node:33616) Warning: Closing file descriptor 9 on garbage collection
(node:33616) Warning: Closing file descriptor 23 on garbage collection
(node:33616) Warning: Closing file descriptor 10 on garbage collection
(node:33616) Warning: Closing file descriptor 11 on garbage collection
```
We ignored this, because it is only a warning and hoped that with this will go away with a future version. But it now seems to be the reason for the stuck frontend-build. 

### Expected-behavior

The build not getting stuck. 
1) The node-execution should ideally not produce these warnings. 
2) the process-invocation via java should properly handle the input/output streams so it won't get stuck when the process produces output

### Reproduction

it sounds a lot like this problem i found here
https://vaadin.com/forum/t/vaadin-24-4-12-building-on-windows-takes-over-an-hour-mac-m2-les-than-a-minute/167101

it gets stuck here in the CommandRunner (https://github.com/vaadin/hilla/blob/main/packages/java/engine-core/src/main/java/com/vaadin/hilla/engine/commandrunner/CommandRunner.java#L187)

Image

reading https://stackoverflow.com/questions/2783190/java-hangs-when-trying-to-close-a-processbuilder-outputstream it seems to me that the process-invocation is not properly implemented if the process produces a significant amount of output since it is not using threads as far as i see. (i found com.vaadin.flow.server.frontend.FrontendUtils#consumeProcessStreams in the codebase)

can you please suggest a workaround in the meantime?

i attach a rather quick-n-dirty fix i attempted which seems to work [CommandRunner.java.txt](https://github.com/user-attachments/files/20998701/CommandRunner.java.txt)

### System Info

- Windows 11
- Java 21
- node: 20.11.1
- vaadin-version: tested with 24.7.1 as well as 24.8.1

platosha:
Fixes #3696


cromoteca (dismissed):
Tests seems to run forever, meaning that something keeps waiting.

PR #3980, opened 2025-09-04
refactor: address style issues from component overlay refactoring

sissbruecker:
## Description

Addresses some small style issues in `AutoGrid` and `AutoCrud` that resulted from the web components overlay refactoring.

## Type of change

- Refactoring


packages/java/tests/spring/react-grid-test/frontend/themes/react-grid-test/styles.css
```diff
@@ -1 +1,2 @@
-/* Import your application global css files here or add the styles directly to this file */
\ No newline at end of file
+/* Import your application global css files here or add the styles directly to this file */
+@import url('@vaadin/vaadin-lumo-styles/lumo.css');
```
sissbruecker, line 2: The component's custom CSS uses Lumo variables so for now they only work properly with the Lumo theme.

packages/ts/react-crud/src/autocrud.obj.css
```diff
@@ -1,6 +1,7 @@
 .auto-crud {
   display: flex;
   overflow: hidden;
+  position: relative;
```
sissbruecker, line 4: Not sure where this issue came from, but without this the box shadow of the form area expands outside of the `AutoCrud`:

Bildschirmfoto 2025-09-04 um 10 31 24

packages/ts/react-crud/src/autocrud.obj.css
```diff
@@ -87,10 +88,7 @@
 }
 
 .auto-crud-dialog {
-  top: 0;
-  bottom: 0;
-  left: 0;
-  right: 0;
+  --vaadin-overlay-viewport-inset: 0;
```
sissbruecker, line 91: The actual overlay element is now in the shadow root of the dialog and can not be targeted with global CSS, there isn't a part name for it either. To customize the offset between the viewport and the overlay this property needs to be used now. Otherwise it looked like this:

Bildschirmfoto 2025-09-04 um 10 34 42

packages/ts/react-crud/src/autogrid.obj.css
```diff
@@ -53,7 +53,7 @@
   flex: 1 1 0;
   width: 120px;
 }
-.auto-grid-enum-filter vaadin-select-item {
+.auto-grid-enum-filter vaadin-select-value-button {
```
sissbruecker, line 56: As the select component now keeps the overlay as a child, this also targeted items in the overlay, but it should only target the selected value. Targeting the value button works as an alternative.

PR #3998, opened 2025-09-09
fix(autocrud): stabilize layout to remove flickering

### LINKED ISSUE #3996: AutoCrud with disabled new button flickers with grid refresh when selecting items
### Describe the bug

When the new item button is shown, the form stays open to allow to add a new item. This doesn’t happen when creating a new item is not allowed, so the form opens and closes and that triggers flickering.

See also: https://vaadin.com/forum/t/autocrud-component-grid-refresh-issue-and-custom-toolbar-buttons/178472/2

### Expected-behavior

No flickering.

### Reproduction

1. Add a view with `<AutoCrud>`
2. Disable new button
3. Try using the view

### System Info

Vaadin 25.0 SNAPSHOT

cromoteca:
Fixes #3996


packages/ts/react-crud/src/autocrud.tsx
```diff
@@ -214,10 +212,12 @@ export function AutoCrud<TModel extends AbstractModel>({
       ) : (
         <SplitLayout theme="small">
           {mainSection}
-          <div className="auto-crud-form">
-            <div className="auto-crud-form-header">{formHeader}</div>
-            {autoForm}
-          </div>
+          {(!noNewButton || (item && item !== emptyItem)) && (
```
Lodin, line 215:
```suggestion
          {(!noNewButton || (item !== emptyItem)) && (
```
I assume we don't need to check the item to be `undefined` as `undefined` never equals to `emptyItem`.

UPD: Just noticed that there is a lot of the same `item && item !== emptyItem` comparisons. Yeah, they are also redundant but we can address it in another PR if you want.

Lodin (changes requested):
LGTM

packages/ts/react-crud/src/autocrud.tsx
cromoteca: I didn't apply your suggestion directly as it was necessary to also remove parentheses.

cromoteca: The logic is actually different: we want an item AND not empty. By simplifying, we're accepting everything different from empty, including null/undefined.

Lodin (approved):

PR #4076, opened 2025-09-25
chore: Don't mix Java package.json files with root package.json

Artur-:
e.g. ITs print warnings because Flow cannot find @vaadin/common-frontend as that is not in the project node_modules

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,12 +7,7 @@
       "name": "@vaadin/hilla",
       "license": "Apache-2.0",
       "workspaces": [
-        "packages/ts/*",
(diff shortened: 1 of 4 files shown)


platosha:
Don't do it.

Warnings aren't meaningful here: Java can't find npm packages in its presumed location (because they're installed in the repository root instead) and thus prints a warning. However, Vite and TypeScript resolve them just fine.

Everything is installed in the repo root for a reason: any misalignment often results in two instances of the same code not compiling against each other. Obscure errors from TypeScript as a result of react-router being installed twice would be a very confusing input to work on...

PR #4081, opened 2025-09-26
fix: improve JSON comparison to provide easier to read errors

Artur-:
Replace string-based JSON comparison with JSONAssert library to avoid test failures caused by formatting changes like whitespace, indentation, or property ordering. Enhanced error messages now provide both precise field-level diffs and complete JSON context for easier debugging.

Changes:
- Add JSONAssert dependency to parser-jvm-test-utils
- Replace string comparison with structural JSON comparison
- Configure module system to allow JSONAssert automatic modules
- Add enhanced error messages showing full JSON context
- Configure compiler and javadoc plugins for module compatibility

### DIFF
--- a/packages/java/parser-jvm-test-utils/pom.xml
+++ b/packages/java/parser-jvm-test-utils/pom.xml
@@ -19,6 +19,33 @@
     <formatter.basedir>${project.parent.basedir}</formatter.basedir>
   </properties>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <compilerArgs>
+            <!-- Allow reading all unnamed modules for JSONAssert -->
+            <arg>--add-reads</arg>
+            <arg>com.vaadin.hilla.parser.testutils=ALL-UNNAMED</arg>
+          </compilerArgs>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <configuration>
+          <additionalJOptions>
+            <!-- Allow reading all unnamed modules for JSONAssert -->
+            <additionalJOption>--add-reads</additionalJOption>
+            <additionalJOption>com.vaadin.hilla.parser.testutils=ALL-UNNAMED</additionalJOption>
+          </additionalJOptions>
+        </configuration>
+      </plugin>
+    </plugins>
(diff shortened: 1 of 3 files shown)


packages/java/parser-jvm-test-utils/src/main/java/com/vaadin/hilla/parser/testutils/JsonAssertions.java
```diff
@@ -11,7 +13,35 @@ public final class JsonAssertions {
...
+            // field
+            String originalMessage = e.getMessage();
+            String enhancedMessage = enhanceErrorMessage(originalMessage,
+                    expected, actual);
+            throw new AssertionError(enhancedMessage, e);
+        }
+    }
+
+    private static String enhanceErrorMessage(String originalMessage,
+            Object expected, Object actual) throws JsonProcessingException {
```
mcollovati, line 35: This method could directly take expected and actual as strings instead of doing an additional conversion that might be different from the one used in test.

packages/java/parser-jvm-test-utils/src/main/java/com/vaadin/hilla/parser/testutils/JsonAssertions.java
```diff
@@ -11,7 +13,35 @@ public final class JsonAssertions {
 
     public static void assertEquals(Object expected, Object actual)
             throws JsonProcessingException {
-        Assertions.assertEquals(printer.pretty().writeAsString(expected),
-                printer.pretty().writeAsString(actual));
+        try {
+            String expectedJson = printer.writeAsString(expected);
+            String actualJson = printer.writeAsString(actual);
```
mcollovati, line 18: I would move this outside the try catch used for the assertion. It should simplify debugging if something fails with the `printer` and it allows enhancing the JSON assertion message by directly providing the compared JSON strings instead of having to serialize the objects again,

packages/java/parser-jvm-test-utils/pom.xml
```diff
@@ -19,6 +19,33 @@
     <formatter.basedir>${project.parent.basedir}</formatter.basedir>
   </properties>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <compilerArgs>
```
mcollovati, line 28: Maybe `<testCompilerArgument>` is better if it is only for testing purposes.

mcollovati (changes requested):

mcollovati (approved):

PR #4300, opened 2025-11-03
test: fix FluxIT login race condition causing CI timeouts

Artur-:
The helloAdmin test was timing out in CI because the login() method had a race condition between form submission and page navigation. The test would click the login form submit button (triggering a POST to /login) and immediately call open() to navigate to /flux, causing the navigation to race with the form POST redirect.

Fixed by adding proper waits after form submission:
- waitForDocumentReady() to ensure form POST completes
- waitUntil() to ensure Spring Security redirect completes before navigating

This matches the pattern used in the working SecurityIT test.


platosha (dismissed):

Artur-:
Could you re-stamp this @platosha ?

PR #4373, opened 2025-11-19
Prepare RouteUtil for changes in FileRouterRequestUtil

heruan:
This prepares `RouteUtil` for changes in https://github.com/vaadin/flow/pull/22753

- [x] Implement `FileRouterRequestUtil::isAnonymousRoute`
- [x] Adds a flag to reuse the common logic needed for the new `isAnonymousRoute` and the deprecated `isRouteAllowed`, ensuring that the former does not invoke `request.getUserPrincipal()`
- [x] Provide a way for Spring Security to check for required roles
- [x] Tests

This prepares for https://github.com/vaadin/flow/pull/22753 implementing new `FileRouterRequestUtil` methods that will later need the `@Override` annotation.


packages/java/endpoint/src/main/java/com/vaadin/hilla/route/RouteUtilConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.context.annotation.Bean;
+
+import com.vaadin.flow.internal.hilla.FileRouterRequestUtil;
+
+@AutoConfiguration
+class RouteUtilConfiguration {
+
+    @Bean
+    @ConditionalOnMissingBean(FileRouterRequestUtil.class)
```
mcollovati: `FileRouterRequestUtil` is an internal API, introduced to create a bridge between Flow and Hilla to configure Spring Security. Could we remove the `@ConditionalOnMissingBean`?
I guess we don't want users to override it implementing custom behaviors.

PR #4705, opened 2026-02-13
fix: handle null loadingCount in grid test waitForLoading method

mshabarov:
The waitForLoading method in AbstractGridTest was causing NullPointerException when globalThis.Vaadin.connectionState.loadingCount was undefined or null. This occurred when the Vaadin object or connectionState wasn't initialized yet during test execution.

Changes:
- Add optional chaining (?.) in JavaScript to safely access nested properties
- Use nullish coalescing (?? 0) to default to 0 when value is undefined/null
- Check if result is a Number instance before calling intValue() in Java

Fixes all 12 failing tests in ReadOnlyGridWithHeaderFilterIT, GridUseDataProviderHookIT, ReadOnlyGridCustomFilterIT, ReadOnlyGridOrFilterIT, and ReadOnlyGridSinglePropertyFilterIT that were failing with the same NullPointerException at AbstractGridTest.java:124.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/java/tests/spring/react-grid-test/src/test/java/com/vaadin/hilla/test/reactgrid/AbstractGridTest.java
+++ b/packages/java/tests/spring/react-grid-test/src/test/java/com/vaadin/hilla/test/reactgrid/AbstractGridTest.java
@@ -118,7 +118,11 @@ public abstract class AbstractGridTest extends ChromeBrowserTest {
         // endpoints finish loading
         this.waitUntil((driver) -> Boolean.TRUE
                 .equals(((JavascriptExecutor) driver).executeAsyncScript(
-                        "const resolve = arguments[arguments.length - 1];\n"
-                                + "globalThis.setTimeout(() => { resolve(globalThis.Vaadin.connectionState.state === 'connected'); }, 100)")));
+                        """
+                                const resolve = arguments[arguments.length - 1];
+                                globalThis.setTimeout(() => {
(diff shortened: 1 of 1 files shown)


mcollovati (dismissed):

packages/java/tests/spring/react-grid-test/src/test/java/com/vaadin/hilla/test/reactgrid/AbstractGridTest.java
```diff
@@ -117,11 +117,12 @@ private void waitForLoading() {
...
         // endpoints finish loading
         this.waitUntil((driver) -> {
-            int loadingCount = ((Number) ((JavascriptExecutor) driver)
-                    .executeAsyncScript(
-                            "const resolve = arguments[arguments.length - 1];\n"
-                                    + "globalThis.setTimeout(() => { resolve(globalThis.Vaadin.connectionState.loadingCount); }, 100)"))
-                    .intValue();
+            Object result = ((JavascriptExecutor) driver).executeAsyncScript(
+                    "const resolve = arguments[arguments.length - 1];\n"
+                            + "globalThis.setTimeout(() => { resolve(globalThis.Vaadin?.connectionState?.loadingCount ?? 0); }, 100)");
```
platosha: Would be better to assert `connectionState?.state !== 'loading'`, as `.loadingCount` is private.

platosha (changes requested):

Issue #3968, opened 2025-09-01
[Full-Stack Signals] implement latency compensation


cromoteca:
### Describe your motivation

Signal commands should be applied immediately on the client, then reverted if failed on the server.

Depends on https://github.com/vaadin/hilla/issues/3967.

platosha:
Barebones structure of the implementation in terms of signals could look like:

```tsx
const confirmedValue = signal(initialValue);

function commandInterpreter(prevValue, command: Command) {
  // implement commands here
}

const unconfirmedCommands = signal<readonly Command[]>([]);
const unconfirmedCommandHanlders = new Map<string, {
  readonly resolve: (value) => void,
  readonly reject: (reason) => void,
}>();

function addUserCommand(command: Command) {
  // Add the command to the unconfirmed list and its callbacks to the map
  const { promise, resolve, reject } = Promise.withResolvers();
  unconfirmedCommandHanlders.set(command.id, { resolve, reject });
  unconfirmedCommands.value = [...pendingCommands, command];
  // Notify the server
  signalEndpoint(userValue.value, command);
  // Return the promise to the user
  return promise;
}

signalFlux.subscribe((confirmedCommand: Command) => {
  batch(() => {  
    // Interpret command regardless of where it's coming from and update the `confirmedValue`
    const [ value, commandResult ] = commandInterpreter(confirmedValue.value, confirmedCommand);
    confirmedValue.value = value;
    if (unconfirmedCommandHandlers.has(confirmedCommand.id)) {
      // The command is coming from this client
      unconfirmedCommands.value = pendingCommands
        .filter((command) => command.id !== confirmedCommand.id);
      const { resolve, reject } = unconfirmedCommandHandlers.get(confirmedCommand.id);
      unconfirmedCommandHandlers.delete(confirmedCommand.id);
      if (commandResult.ok) {
        resolve(commandResult.value);
      } else {
        reject(commandResult.reason);
      }
    } else {
      // Someone else's command, nothing else to do
    }
  });
});

const userValue = computed(() => {
  const [ value, ] = unconfirmedCommands.reduce(commandInterpreter, confirmedValue.value);
  return value;
});
```

PR #4890, opened 2026-03-16
feat: implement shared signals command rejection, latency compensation, and MapSignal

Artur-:
- Server (InternalSignal): add accepted/reason fields to command JSON, notify only the originating client on rejection, handle server-originated commands without NPE
- Client (FullStackSignal): route rejected commands via $handleRejection
- ValueSignal: track confirmed value and pending sets, revert on rejection
- NumberSignal: optimistic incrementBy with pending tracking, revert on rejection
- ListSignal: optimistic insert/remove, add clear(), revert on rejection
- Add MapSignal (TS + Java transfer type + generator plugin mapping)
- Add ClearCommand, PutCommand, RemoveByKeyCommand to commands.ts
- Re-enable all 4 integration tests (BasicSignalIT, NumberSignalIT, SignalBasicSecurityIT, ParamsIT)
- Fix test services: use peek() instead of get() for non-reactive context


platosha (changes requested):
This extends the current implementation of optimistic updates, which has known design issues with its core approach.

A better design is sketched here: https://github.com/vaadin/hilla/issues/3968#issuecomment-3244293275, and some implementation work was started in this branch: https://github.com/vaadin/hilla/compare/main...refactor/node-signals 

The main differences are:
- There is a common unconfirmed commands queue in the signal for all commands instead of ad-hoc command-specific queues here and there
- There is no `setValueQuietly` at all, the method is removed entirely. Instead, the client-side signal state consists of the confirmed state and the unconfirmed commands queue. The value of the signal is always computed by applying the current unconfirmed commands over the current confirmed state.
- Rejection removes the unconfirmed command from the unconfirmed queue. No manual reverting of the state is needed — modifying the unconfirmed queue triggers re-computation of the value explained above, the value is expected to update accordingly.

The branch also contains some good foundation cleanup, would be great to incorporate that. Maybe we can even merge it here and continue from then.

Artur-:
Removed unnecessary changes

PR #4978, opened 2026-03-31
test: use stylesheet instead of theme folder in all test projects

Artur-:


ZheSun88:
`Configurator.java` — All @StyleSheet paths now use context:// prefix so they resolve relative to the web context root instead of the Vaadin servlet mapping. This is required when vaadin.urlMapping is set (per the              
  https://vaadin.com/docs/latest/upgrading#themes-and-styling):                   

`SecurityConfig.java` — Added /css/** and /lumo/** to the public permit list, since @StyleSheet resources (unlike @Theme resources) aren't auto-permitted by Vaadin's security infrastructure:
  .requestMatchers("/public/**", "/css/**", "/lumo/**").permitAll()

platosha (approved):

PR #5150, opened 2026-04-17
fix(TypeParameterModel): prevent StackOverflowError on F-bounded generic equals

Mati0kez:
## Root Cause

When comparing two  instances whose bounds contain
F-bounded type variables (e.g., `T extends Foo<T>`), the existing
`equals()` implementation caused infinite recursion:

1. `TypeParameterModel.equals()` calls `getBounds().equals()`
2. If a bound is a `TypeVariableModel`, `SignatureModel.equals()`
   delegates to `TypeVariableModel.equals()`
3. `TypeVariableModel.equals()` calls `resolve().equals()`
4. `resolve()` returns a `TypeParameterModel` whose `equals()`
   calls `getBounds().equals()`
5. If bounds contain the same `TypeVariableModel` (self-referential
   bound), the cycle repeats → `StackOverflowError`

## Fix

In `TypeParameterModel.boundEqual()`, when comparing two
`TypeVariableModel` bounds, compare their names directly instead of
calling `equals()`. This breaks the cycle while preserving correct
equality semantics (two type variables with the same name are equal,
unless annotations differ — but annotation comparison is handled by the
parent `boundsEqual` loop).

## Changes

- `TypeParameterModel.java`: Replace `getBounds().equals()` with a
  bounds-element-wise comparison that short-circuits the cycle for
  `TypeVariableModel` pairs
- `TypeParameterModelTests.java`: Add `FBoundedTypeParameter` to
  `Sample` class and a regression test for self-equality

## Testing

Added `should_NotStackOverflowOnFBoundedEquals` regression test.
Could not run the full test suite (no Maven/Gradle/Java toolchain in
this environment); please run `mvn test -Dtest=TypeParameterModelTests`
to verify.


platosha (approved):
Looks good. Thanks for the contribution!

platosha:
@Mati0kez Hi, we require the Contributor License Agreement signature from all contributors. In this case, since the commit author is @BillionClaw, they need to sign the CLA before we could accept this change. Here is the link again: https://cla-assistant.io/vaadin/hilla?pullRequest=5150

platosha:
@Mati0kez sorry, we couldn't move forward with this without the CLA, even though the fix seems good. Feel free to reach out and resubmit it when you're ready to sign.

PR #5172, opened 2026-04-21
chore: fix snapshot failures

ZheSun88:
fix test failure in snapshot build 
https://bender.vaadin.com/buildConfiguration/HillaProject_HillaSnapshotRelease/863030?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildTestsSection=true

image

### DIFF
--- a/packages/java/tests/spring/react-grid-test/src/main/frontend/views/ComboBoxUseComboBoxDataProviderHook.tsx
+++ b/packages/java/tests/spring/react-grid-test/src/main/frontend/views/ComboBoxUseComboBoxDataProviderHook.tsx
@@ -1,5 +1,6 @@
 import { useComboBoxDataProvider } from '@vaadin/hilla-react-crud';
 import { useSignal } from '@vaadin/hilla-react-signals';
+import { useSignals } from '@preact/signals-react/runtime';
 import { Button, ComboBox } from '@vaadin/react-components';
 import { useState } from 'react';
 import { PersonCustomService } from 'Frontend/generated/endpoints';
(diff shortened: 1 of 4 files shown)


ZheSun88:
```
  Summary                                                                                                                                                                                                                          
                                                                                                                                                                                                                                   
  Root cause: Flow 25.2-SNAPSHOT's bundled vite.generated.ts template was changed (between the April 17 and April 20 snapshot builds) to run @preact/signals-react-transform as a enforce: 'post' plugin — after reactPlugin's OXC 
  converts JSX into _jsx() calls. The signals-react-transform only looks for JSXElement/JSXFragment AST nodes by default and doesn't mark components as "contains JSX" when it sees _jsx(...) calls, so it skips wrapping them with
   useSignals(). Without that wrapper, signal reads no longer trigger re-renders.                                                                                                                                                  
                                                                          
  This is why:
  - ComboBoxUseComboBoxDataProviderHookIT.filteringUsingSignalWorks still fails after the dispatchEvent fix — filterSignal.value = "c" updates the signal, but the component never re-renders to re-evaluate
  useComboBoxDataProvider(..., [filterSignal.value]).
  - All 8 BasicI18NIT.setLang* tests fail — setValue fires the events fine, i18n.setLanguage() updates the signals, but <span>{i18n.language.value}</span> doesn't re-render.
  - GridUseGridDataProviderHookIT.filteringUsingSignalWorks passes — that view happens to include a setState call alongside the signal update as a manual re-render trigger.

  Fix: Added an explicit useSignals() call from @preact/signals-react/runtime to both affected views. That sets up a tracking effect for signals read during render, which doesn't depend on the Babel transform recognizing the
  components.

  - packages/java/tests/spring/react-grid-test/src/main/frontend/views/ComboBoxUseComboBoxDataProviderHook.tsx
  - packages/java/tests/spring/react-i18n/src/main/frontend/views/basic-i18n.tsx

  This is a workaround until Flow's template is fixed (either by passing detectTransformedJSX: true to the transform, or by moving it back to enforce: 'pre' so it runs on raw TSX).

```

ZheSun88:
we are not going to take this approach.

PR #5291, opened 2026-05-05
chore: migrate away from tsc-template dependency

ZheSun88:
PR has been made to https://github.com/ausginer/tsc-template/pull/1
if the upstream PR cannot be merged in time, let us take this 

Migration done. Summary:                                                          
                                                                                                                                                   
  - New file: packages/ts/generator-utils/src/tsc-template.ts—single file inlining of all 5 upstream files, with Apache 2.0 attribution preserved
   in the header. Exposes ast (default) and createTransformer.                                                                                     
  - Exports: added ./tsc-template.js to generator-utils package.json exports map. The existing "files" glob already covers it.                     
  - Consumers updated: file-router/src/vite-plugin/createRoutesFromMeta.ts and generator-plugin-signals/src/SignalProcessor.ts now import from     
  @vaadin/hilla-generator-utils/tsc-template.js.                                                                                                   
  - Dependency dropped from both consumer package.json files; npm install removed it from the lockfile and node_modules.                           
  - Verified: typecheck + tests pass for generator-utils, file-router (51 tests), generator-plugin-signals (5 tests). The original npm warn peer   
  typescript@"5" warning is gone since the package no longer exists in the tree.

PR #5292, opened 2026-05-05
chore: align javaparser version among projects in main

ZheSun88:


ZheSun88:
separate the unrelated change from #5291 to make the validation build run

PR #5580, opened 2026-06-11
build: add branch parameter to prepare script

manolo:
## Summary

The prepare script now accepts a `--platform-branch` parameter to fetch versions.json from a specific platform branch instead of always using main. This enables testing against PR branches before they are merged.

## Context

When testing platform changes in a PR, we need to pull versions.json from the PR branch rather than from main. Previously, this required manually editing the hardcoded branch constant in config.ts.

## Usage

Default behavior (uses main):
```bash
npm run build:prepare
```

Test against a specific branch:
```bash
npm run build:prepare -- --platform-branch=25.1
npm run build:prepare -- --platform-branch=feature/my-pr-branch
```

## Changes

- Parse `--platform-branch` CLI argument in config.ts
- Default to main branch when no parameter is provided
- Backward compatible with existing scripts and CI workflows


ZheSun88:
a few findings: 
1. in the release build and snapshot build, we dont call `npm run build:prepare` task directly, but use `npm run build`. 

i tried to use `npm run build -- --branch=25.1`, from the log, 
```
Fetching versions from platform branch: main
Generating release files
```

to make our life easier, would it be possible to make it work with `npm run build`?

2. a drawback: this will introduce a loop here
- we make a release via using `--branch=x.y`
- in the release build, we update all package.json with the version from branch x.y and commit back to the main branch here.
- this commit will trigger a hilla snapshot build, which uses the versions.json from default branch, and revert the previous updates in all package.json 

should not be a big problem, but might cause confusions. 
 
3. this PR cannot be cherry-picked directly as the default branch here is `main`, it needs to be upgraded to the matching branches when cherry-picking.

scripts/prepare/config.ts
```diff
@@ -49,6 +48,12 @@ export const remote = {
   versions: new URL(`platform/${branch}/versions.json`, repoUrl),
 };
 
+export function setBranch(newBranch: string): void {
+  branch = newBranch;
+  remote.src = new URL(`platform/${branch}/scripts/generator/src/`, repoUrl);
+  remote.versions = new URL(`platform/${branch}/versions.json`, repoUrl);
+}
```
platosha: This is subject of timing issues if some work for some reason happens before `setBranch`. To avoid that, I'd apply arguments right in this module, so that `remote.src` is already modified when value is exported.

Let me follow up with commits on this branch with adjustments about this.

platosha (changes requested):

scripts/prepare/config.ts
platosha: My suggestions are in. Also I renamed the argument to `--platfrom-branch=`, as generic `branch` is a bit confusing to me.

platosha:
@manolo mind if we name the parameter as `--platform-branch=`?

manolo:
I think everything is done in this PR, @platosha can we merge it?

PR #5734, opened 2026-07-01
build: purge cached Vaadin Maven snapshots in every validation build

platosha:
Addresses the version skew affecting Gradle CI validation build.


mcollovati:
Gradle build seems to still fail

platosha:
Seems to work now: https://github.com/vaadin/hilla/actions/runs/28526451873

.github/workflows/validation.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - name: Build TypeScript
         run: npm run build
       - name: Build Java
-        run: mvn install -B -ntp -DskipTests -Dmaven.test.skip=true -Dmaven.javadoc.skip=true
+        run: mvn install -B -U -ntp -DskipTests -Dmaven.test.skip=true -Dmaven.javadoc.skip=true
```
mcollovati: nit: If Vaadin snapshots are removed from the repository, we can probably avoid `-U`

mcollovati (dismissed):

.github/workflows/validation.yml
platosha:
```suggestion
        run: mvn install -B -ntp -DskipTests -Dmaven.test.skip=true -Dmaven.javadoc.skip=true
```

mcollovati (approved):

PR #5748, opened 2026-07-02
chore: update CI branches

platosha:


.github/workflows/validation.yml
```diff
@@ -6,7 +6,7 @@ defaults:
 
 on:
   push:
-    branches: [main, '25.1', '25.0', '24.9', '2.5', '1.5']
+    branches: [main, '25.2', '25.1', '24.10', '24.9', '2.5', '1.5']
```
tltv, line 9: Is 25.0 out of maintenance or is it just missed from the list here?

tltv: I see it's marked EOL last week, so should be ok but let's confirm it.

tltv (approved):

PR #5910, opened 2026-07-27
fix: do not trigger the production frontend build from hillaGenerate

caalador:
Flow 25.3 registers the production frontend bundle as an extra output directory of the main source set (vaadin/flow#25024), which makes the `classes` lifecycle task depend on `vaadinBuildFrontend`. Since hillaGenerate depended on `classes`, running it started a full production frontend build first: the bundle got built before the endpoints were generated, and the frontend files Flow generates for the bundle - node_modules included - were cleaned up afterwards, so the generator could no longer resolve @vaadin/hilla-generator-cli and the task failed with

    GeneratorException: Unable to resolve npm package "@vaadin/hilla-generator-cli"

Depend on the source set compile tasks and processResources instead of the `classes` lifecycle task, and add a functional test asserting that hillaGenerate keeps the compile tasks but does not pull in vaadinBuildFrontend.

### DIFF
--- a/packages/java/gradle-plugin/src/functionalTest/kotlin/com/vaadin/hilla/gradle/plugin/SingleModuleTest.kt
+++ b/packages/java/gradle-plugin/src/functionalTest/kotlin/com/vaadin/hilla/gradle/plugin/SingleModuleTest.kt
@@ -63,6 +63,28 @@ class SingleModuleTest : AbstractGradleTest() {
     }
 
     @Test
+    fun `hillaGenerate does not trigger the production frontend build`() {
+        createProject(productionMode = true, disableAllTasksToSimulateDryRun = true)
+
+        addHelloReactEndpoint()
+        addMainClass()
+
+        val buildResult: BuildResult = testProject.build("hillaGenerate", checkTasksSuccessful = false)
+
+        expect(true, "hillaGenerate should depend on the compiled classes") {
+            buildResult.task(":compileJava") != null
+        }
+        // The production frontend bundle is registered as an extra output
+        // directory of the source set, which makes the `classes` lifecycle task
+        // depend on vaadinBuildFrontend. hillaGenerate must not be affected by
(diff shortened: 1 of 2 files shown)


ZheSun88:
```
Root cause: Flow 25.2 commit 59f2393dc — "chore: Update frontend dependencies (25.2)" (#25050, today 06:56) — raised Flow's default frontend deps:                                                                               
                                                                                                                                                                                                                                   
  react / react-dom      19.2.7 → 19.2.8                                                                                                                                                                                           
  @vitejs/plugin-react     6.0.3 → 6.0.4                                                                                                                                                                                           
  @preact/signals-react-transform  0.8.1 → 0.8.2                                                                                                                                                                                   
  @vaadin/common-frontend  0.0.23 → 0.0.24                                                                                                                                                                                         
                                                                                                                                                                                                                                   
  Hilla 25.2 still pins react/react-dom at exactly 19.2.7. The test apps are npm workspace members, so when Flow writes react: 19.2.8 into an app's package.json, npm keeps 19.2.7 hoisted at the root (used by                    
  packages/ts/react-*) and nests 19.2.8 under the app. I reproduced that locally by setting one app to 19.2.8:                                                                                                                     
                                                                                                                                                                                                                                   
  19.2.7  node_modules/react                                                                                                                                                                                                       
  19.2.8  packages/java/tests/gradle/kotlin-gradle-test/node_modules/react                                                                                                                                                         
                                                                                                                                                                                                                                   
  Two React instances in one bundle → the jsx-runtime-*.js chunk gets null internals → Uncaught TypeError: Cannot read properties of null (reading 'useRef') (216 occurrences in the react-grid app), so nothing renders: blank    
  pages in the Gradle IT screenshots, BasicI18NIT setup timeouts, AccessModifierIT reading empty text.  
```

now after regenerate the package.json .. and apply the same fix here. the build should pass https://github.com/vaadin/hilla/pull/5892

ZheSun88:
we can close this PR now. :-)

PR #6029, opened 2026-08-14
chore: upgrade kotlin.version to resolve vulnerability

ZheSun88:
https://github.com/advisories/GHSA-r937-wjx7-w2jp


ZheSun88:
let us wait for a stable version instead of using prerelease versions

PR #6102, opened 2026-08-27
chore: update web components to 25.3.0-alpha14

manolo:
Platform main is already on alpha14, and react-crud was still pinning alpha13. With both pinned we end with two versions of the components installed.

Bumped the whole set so nothing gets duplicated, `npm ci` is happy.


manolo:
Closing, I had the direction wrong. Platform main is on alpha13 and Hilla follows it, so this would be reverted tonight. We bump platform first.

PR #6216, opened 2026-09-09
refactor: implement NameRegistry for proper identifier uniqueness

platosha:
- **New `NameRegistry` class** - A new utility that tracks identifier names in a single generated file, ensuring uniqueness by appending counters when collisions occur (e.g., `name_1`, `name_2`)
- **Removed old function** - The previous `createFullyUniqueIdentifier.ts` was deleted as it's no longer needed
- **Test snapshots updated** - Multiple test files had their identifier suffixes (`_1`) removed since the new registry properly handles uniqueness


Artur-:
Should this rather be done on top of #6151 ?

platosha:
> Should this rather be done on top of #6151 ?

This one is relatively less dependant, no strong opinion. However, generally, I think it makes sense to simplify the output before going to Java, as this reduces scope for the Java generator implementation.

PR #6224, opened 2026-09-09
test(generator-plugin-model): smoke test that output compiles

platosha:
* Add runtime tests for ModelPlugin that check generated model files can be evaluated
* Test handles circular dependencies between mutually referencing models (e.g., FormEntity ↔ FormArrayTypes)
* Update .gitignore to exclude test-generated directories from version control


platosha:
@claude update the description to reflect all changes in the PR after the last review comment

